PR #358, opened 2026-04-15
refactor(config): centralize environment and auth configuration

pvliesdonk:
- Move OIDC, Bearer, and Remote auth logic from mcp_server.py to config.py
- Refactor embedding providers to be configuration-driven instead of reading os.environ
- Update CollectionConfig to include server identity and auth settings
- Update cli.py and tests to use the new centralized configuration


pvliesdonk:
@claude review

PR #367, opened 2026-04-16
refactor(collection): extract text normalization and edit helpers

pvliesdonk:
Addressing review feedback: added utils/__init__.py, used str.translate, added edge case tests.


pvliesdonk:
Closing: superseded by new issues with a redesigned approach. The refactoring will use explicit dependency injection instead of back-references, and config centralization will be a standalone effort. See the new issues for details.

PR #393, opened 2026-04-19
ci: move manifest bump into semantic-release build_command

### LINKED ISSUE #392: release: move manifest bump into semantic-release build_command to eliminate tag-move race
## Context

The release workflow currently has a structural race between manifest-bumping and tag creation:

1. python-semantic-release commits pyproject.toml + CHANGELOG, tags that commit as \`v\$VERSION\`, pushes, and publishes the GitHub release.
2. \`release.yml\` then bumps three manifest files (\`server.json\`, \`.claude-plugin/plugin/.claude-plugin/plugin.json\`, \`.claude-plugin/plugin/.mcp.json\`), commits them as \"chore: update manifests to v\$VERSION [skip ci]\", and **force-updates the tag** to point at the bump commit (\`release.yml:102-103\`).

The tag move is deliberate — the Claude plugin marketplace checks out \`.claude-plugin/plugin/\` at \`v\$VERSION\` and needs the bumped manifests there. But the tag move also breaks any workflow triggered by the \`release\` event: by the time those workflows run, the tag SHA no longer matches what the event payload said.

**Manifestations:**

- **v1.23.0:** release-triggered \`docs.yml\` build failed with \`The ref 'refs/tags/v1.23.0' does not point to the expected commit 'a337271...'. The ref may have been updated after the workflow was triggered.\` Worked around by #391, which forces release-triggered docs builds to checkout \`main\` instead of the triggering tag.
- **Any future release-triggered workflow** will hit the same race unless it uses the same defensive-checkout pattern.
- **Git history has a useless \"chore: update manifests [skip ci]\" commit on every release.**

## Proposed structural fix

Move the manifest bump into python-semantic-release's \`build_command\`. In semantic-release v10, \`build_command\` runs after version determination but before the release commit, and any files it modifies are included in that commit.

### Concrete steps

1. **New script:** \`scripts/bump_manifests.sh\` (or .py) — does the same three jq/edit operations currently in \`release.yml:66-89\`, reading the version from \`\$NEW_VERSION\` (semantic-release provides it).

2. **\`pyproject.toml\` config:**

   \`\`\`toml
   [tool.semantic_release]
   version_toml = [\"pyproject.toml:project.version\"]
   build_command = \"bash scripts/bump_manifests.sh\"
   # ... existing config
   \`\`\`

3. **\`release.yml\` deletions:**
   - Remove lines 61-103 (the \"Update versioned manifests to released version\" step, including the \`git tag -f\` + force push).
   - Keep \`Build package\` (uv build), SBOM generation, release-asset uploads, and the downstream jobs unchanged.

4. **Keep #391's defensive checkout in \`docs.yml\`** as belt-and-suspenders. No harm in both.

### Result

- One commit per release: pyproject.toml + CHANGELOG + server.json + plugin.json + .mcp.json all atomic.
- One tag, no force-retag.
- No race for release-triggered workflows.
- No more \"chore: update manifests [skip ci]\" noise in git history.
- Marketplace still gets the manifests at the tag (because they're in the same commit as the tag now).

## Test plan

The release flow can't be tested without cutting a release. Options:

- **RC release** (\`prerelease: true\` via workflow_dispatch) — validates the refactor without publishing to PyPI or the marketplace. The RC flow skips \`publish-claude-plugin-pr\`, so the marketplace impact is moot; the relevant thing to verify is that the release commit contains all three bumped manifests and no \"chore: update manifests\" commit follows.
- **Dry-run locally:** \`uvx python-semantic-release --noop\` can show the planned release actions, including whether the build_command would run and what files it would modify.

## Out of scope

- Keeping #391's defensive checkout in \`docs.yml\` — it stays as defense in depth.
- Broader changes to the manifest layout (e.g., generating them at install time) — a separate, much larger design question.

## References

- Race diagnosed during v1.23.0 release: [docs.yml run](https://github.com/pvliesdonk/markdown-vault-mcp/actions/runs/24637391168) (the release-triggered one that failed).
- Defensive workaround merged: #391.
- python-semantic-release build_command docs: https://python-semantic-release.readthedocs.io/en/latest/configuration.html#build-command

pvliesdonk:
## Summary

- Moves the three-manifest version bump (`server.json`, `.claude-plugin/plugin/.claude-plugin/plugin.json`, `.claude-plugin/plugin/.mcp.json`) from a post-release workflow step into python-semantic-release's `build_command`, with the files declared as `assets` so PSR stages + commits them into the release commit it then tags.
- Deletes the 44-line "Update versioned manifests to released version" step in `release.yml`, including the `git tag -f` + force push that was the source of the race.
- Result: one release commit, one tag, no force-retag, no "chore: update manifests [skip ci]" follow-up, and release-triggered workflows (e.g. `docs.yml`) no longer see a tag whose SHA changed mid-run.

Closes #392. Keeps #391's defensive `main`-checkout in `docs.yml` as belt-and-suspenders.

## Why `assets` is needed

PSR's `build_command` runs between `apply_version_to_source_files` and `git_add` (see [`version.py:662-711`](https://github.com/python-semantic-release/python-semantic-release/blob/master/src/semantic_release/cli/commands/version.py)), but it does **not** auto-stage files that the build command modifies — `all_paths_to_add` is built from version-declaration paths + `assets` only. Listing the three manifests in `[tool.semantic_release] assets` is what lands them in the release commit.

## Behavior change

The previous workflow only bumped manifests on stable releases (`!inputs.prerelease`). With `build_command`, PSR runs the script for every release including RCs, so RC commits will now also carry bumped manifests. This is the behavior the issue explicitly endorses — the marketplace PR is still gated on `!inputs.prerelease`, so RC bumps never escape the repo.

## Test plan

- [x] `NEW_VERSION=9.9.9-smoke bash scripts/bump_manifests.sh` — verified all three files update to the expected shape; restored cleanly.
- [x] `uvx --from python-semantic-release==10.5.3 semantic-release --noop version` — config parses cleanly (only pre-existing deprecation warnings for `angular` parser and `changelog_file` location).
- [x] `uv run pytest -x -q` — 1397 passed.
- [x] `uv run mypy src/` — no issues.
- [x] `uv run ruff check --fix . && uv run ruff format . && uv run ruff format --check .` — clean.
- [ ] Cut an RC release (`workflow_dispatch` with `prerelease: true`) after merge and verify: (1) the release commit contains all three bumped manifests, (2) no separate "chore: update manifests" commit follows, (3) release-triggered workflows check out a stable tag SHA.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -58,50 +58,6 @@ jobs:
           prerelease: ${{ inputs.prerelease }}
           prerelease_token: rc
 
-      - name: Update versioned manifests to released version
-        if: steps.release.outputs.released == 'true' && !inputs.prerelease
-        env:
-          VERSION: ${{ steps.release.outputs.version }}
-        run: |
-          # server.json (existing logic)
-          jq --arg v "$VERSION" '
-            .version = $v |
-            .packages |= map(
-              if .registryType == "pypi" then .version = $v
-              elif .registryType == "oci" then .identifier = ("ghcr.io/pvliesdonk/markdown-vault-mcp:v" + $v)
-              else . end
-            )
-          ' server.json > server.json.tmp
-          mv server.json.tmp server.json
-
-          # Claude Code plugin.json (new)
-          jq --arg v "$VERSION" '.version = $v' \
-            .claude-plugin/plugin/.claude-plugin/plugin.json > plugin.json.tmp
-          mv plugin.json.tmp .claude-plugin/plugin/.claude-plugin/plugin.json
-
-          # Claude Code .mcp.json — keep --from spec in lockstep with plugin.json (new)
-          jq --arg v "$VERSION" '
-            ."markdown-vault-mcp".args = [
-              "--from", ("markdown-vault-mcp[all]==" + $v),
-              "markdown-vault-mcp", "serve"
-            ]
-          ' .claude-plugin/plugin/.mcp.json > mcp.json.tmp
-          mv mcp.json.tmp .claude-plugin/plugin/.mcp.json
-
-          git config user.name "github-actions"
-          git config user.email "actions@users.noreply.github.com"
-          git add server.json \
-                  .claude-plugin/plugin/.claude-plugin/plugin.json \
-                  .claude-plugin/plugin/.mcp.json
-          git diff --cached --quiet || git commit -m "chore: update manifests to v${VERSION} [skip ci]"
-          git push
-
-          # Re-point the release tag to the manifest-bump commit so that
-          # marketplace installs (which checkout .claude-plugin/plugin at
-          # ref: vVERSION) get the updated plugin.json and .mcp.json.
-          git tag -f "v${VERSION}"
-          git push --force origin "refs/tags/v${VERSION}"
-
       - name: Build package
         if: steps.release.outputs.released == 'true'
         run: uv build
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -161,6 +161,17 @@ ignore_errors = true
 version_toml = ["pyproject.toml:project.version"]
 commit_parser = "angular"
 tag_format = "v{version}"
+# Bump the versioned manifests inside the release commit rather than in a
+# follow-up commit + force-retag. See issue #392.
+build_command = "bash scripts/bump_manifests.sh"
+# PSR stages and commits files listed in `assets` together with pyproject.toml
+# and CHANGELOG.md, so the release tag points at a single commit containing
+# all version-coupled files.
+assets = [
+    "server.json",
+    ".claude-plugin/plugin/.claude-plugin/plugin.json",
+    ".claude-plugin/plugin/.mcp.json",
+]
 
 [tool.semantic_release.changelog]
 changelog_file = "CHANGELOG.md"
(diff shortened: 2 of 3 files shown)


pvliesdonk:
Addressed review feedback:

**Applied (@gemini-code-assist):** OCI identifier now preserves the image base via `sub(":v[^:]+$"; ":v" + $v)` so forks/renames aren't clobbered. Smoke-tested on both the current repo value (`ghcr.io/pvliesdonk/markdown-vault-mcp:v1.23.0`) and a simulated fork (`ghcr.io/someone-else/forked-repo:v1.0.0`) — both preserved their base and updated the tag correctly. f8bdead.

**Not applied (@claude bot):**

- *Trap cleanup for `.tmp` on partial failure* — `set -e` aborts on first `jq` failure and GitHub-hosted runners are ephemeral, so stray `.tmp` files have no lifetime beyond the failed job. The trap would add complexity for zero practical benefit.
- *Missing trailing newline on line 40* — the file already ends with `\n` (`tail -c 3 scripts/bump_manifests.sh | od -c` → `o n \n`). Bot was mistaken.
- *Explicit `jq` dependency note* — `ubuntu-latest` ships `jq`, and the four `jq` calls make the dependency self-evident; a comment would add noise.

PR #394, opened 2026-04-19
fix: rewrite bump_manifests as Python for PSR Docker container

pvliesdonk:
## Summary

The refactor merged in #393 moved the manifest version bump into python-semantic-release's `build_command`. The shell+jq implementation worked locally (where `jq` is pre-installed on ubuntu-latest) but **not** in the PSR Docker action container, where `build_command` actually runs. Release run [24638594957](https://github.com/pvliesdonk/markdown-vault-mcp/actions/runs/24638594957) aborted at the build_command step:

```
scripts/bump_manifests.sh: line 20: jq: command not found
Build command failed with exit code 127
```

`python-semantic-release/python-semantic-release@v10.5.3` is a Docker-based action running inside `python:3.14-slim-trixie`. That image has Python (obviously — PSR is Python) but no `jq`.

## Fix

Replace `scripts/bump_manifests.sh` with `scripts/bump_manifests.py` — same logic using stdlib `json` + `re`. Python is guaranteed available in the PSR container.

## Verification

- **Idempotency:** `NEW_VERSION=1.23.0 python scripts/bump_manifests.py` produces a zero diff against the current manifests — output format byte-identical to the shell version (`json.dump(..., indent=2)` + trailing newline matches `jq` default + existing trailing newline).
- **Smoke test:** `NEW_VERSION=9.9.9-smoke python scripts/bump_manifests.py` updates all three manifests correctly; OCI identifier regex still preserves fork image bases (e.g. `ghcr.io/someone-else/fork:v1.0.0` → `ghcr.io/someone-else/fork:v9.9.9-smoke`).
- `uv run pytest`, `uv run mypy`, `uv run ruff check/format` all clean.

## Test plan after merge

Re-trigger the Release workflow with `force: patch`, `prerelease: true`:

```
gh workflow run release.yml -f force=patch -f prerelease=true
```

Expect the PSR step to run the Python script successfully, commit all four files (`pyproject.toml` + `CHANGELOG.md` + 3 manifests) in one commit, tag it, and push.


pvliesdonk:
Addressed review feedback:

**Applied (@gemini-code-assist inline #2):** `json.dump(..., ensure_ascii=False)` so the script preserves UTF-8 literally, matching `jq`'s default. All three manifests are ASCII-only today so the change is a no-op in practice, but it keeps the "byte-identical to the shell version" promise if a non-ASCII character ever lands in a description. d538eff.

**Declined (@gemini-code-assist inline #1 + @claude bot observation #3 — uv.lock/pyproject mismatch):** This is a misread of the diff. `pyproject.toml` has been at `1.23.0` on main since the v1.23.0 release (verified: `grep ^version pyproject.toml` → `version = "1.23.0"`). The uv.lock was **stale** at `1.22.1` — this PR syncs it *up* to match, which is the bugfix direction, not divergence. The change survives `uv sync`. Only reason the lockfile appears in this PR at all is that I couldn't get a clean pre-commit run without it staged.

**Declined (@claude bot #1 — success log line):** PSR already logs `Running build command bash scripts/bump_manifests.sh` with a green hammer-and-wrench before invoking, and a silent 0-exit is the standard convention for Unix tools that have nothing to say. Extra chatter would be noise.

**Accepted as-is (@claude bot #2, #4):** matches shell-script behavior; no regression.

PR #395, opened 2026-04-20
test: accept PSR prerelease version format in mcpb packaging tests

pvliesdonk:
## Summary

Fallout from the manifest-bump refactor (#393/#394): PSR's `build_command` now runs for every release including RCs, so `main` can carry `X.Y.Z-rc.N` in `plugin.json` and `.mcp.json` between stable releases (this was the intentional behavior change called out in #392).

Two tests in `tests/test_packaging_mcpb.py` baked in a strict `X.Y.Z` check and broke on the first post-refactor RC push ([run 24638882251](https://github.com/pvliesdonk/markdown-vault-mcp/actions/runs/24638882251) on `v1.23.1-rc.1`):

- `test_claude_code_plugin_json_shape` — required 3 dot-separated integer parts
- `test_plugin_mcp_json_pinned_and_matches_plugin_version` — required `markdown-vault-mcp[all]==X.Y.Z`

## Fix

Introduce a shared `_VERSION_RE = re.compile(r"\d+\.\d+\.\d+(?:-rc\.\d+)?")` and use it in both assertions. The other invariants the tests check (no template literal, plugin.json/.mcp.json agreement, the `uvx` / `--from` wiring) are untouched.

Also catches up `uv.lock` to `1.23.1rc1` to match the new `pyproject.toml` version (uv normalizes to PEP 440 form).

## Test plan

- [x] `uv run pytest tests/test_packaging_mcpb.py -v` → 6 pass (against current main at `1.23.1-rc.1`)
- [x] Full suite: `uv run pytest -x -q` → 1397 pass
- [x] `uv run ruff check/format`, `uv run mypy src/` clean

### DIFF
--- a/tests/test_packaging_mcpb.py
+++ b/tests/test_packaging_mcpb.py
@@ -15,6 +15,12 @@ REPO_ROOT = Path(__file__).resolve().parent.parent
 MCPB_DIR = REPO_ROOT / "packaging" / "mcpb"
 PLUGIN_DIR = REPO_ROOT / ".claude-plugin" / "plugin"
 
+# Stable `X.Y.Z` or python-semantic-release prerelease format `X.Y.Z-rc.N`.
+# The release workflow bumps the manifests for every release including RCs
+# (via PSR's build_command), so main may carry an RC version between stable
+# releases.
+_VERSION_RE = re.compile(r"\d+\.\d+\.\d+(?:-rc\.\d+)?")
+
 
 def test_cli_main_import_target_exists() -> None:
     """The mcpb shim imports markdown_vault_mcp.cli.main — make sure it exists."""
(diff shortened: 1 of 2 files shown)


pvliesdonk:
Addressed review feedback:

**Applied (@claude bot docstring nit):** updated `test_plugin_mcp_json_pinned_and_matches_plugin_version` docstring to `<X.Y.Z[-rc.N]>`. b736e6b.

**Declined (@gemini-code-assist, regex "too restrictive"):** Gemini's reasoning assumes the test ever inspects a PEP 440 normalized version like `1.23.1rc1`. It doesn't — the tests only assert against `plugin.json` and `.mcp.json`, which are written by `scripts/bump_manifests.py` using PSR's `NEW_VERSION` verbatim (always `X.Y.Z-rc.N`, never the normalized form). The normalized `1.23.1rc1` only appears in `uv.lock`, which these tests do not read. A broader regex that accepts formats the manifests can't actually contain would hide real bugs (e.g. a truncated or malformed release tag) rather than catch them.

PR #396, opened 2026-04-20
refactor(config): adopt fastmcp-pvl-core env helpers + ServerConfig (MV-PR1)

pvliesdonk:
## Summary

First PR in the fastmcp-pvl-core adoption series. Pulls in the new shared library and starts replacing MV's local infrastructure with it.

- Adds dependency on \`fastmcp-pvl-core==1.0.0rc1\`. The pin will move to subsequent rc.N releases as the rest of the adoption series uncovers gaps; once all 7 MV-PRs land cleanly, we cut \`fastmcp-pvl-core==1.0.0\` stable and re-pin to \`>=1.0.0,<2\`.
- Replaces local \`_env\` / \`_parse_bool\` / \`_parse_list\` / \`_parse_scopes\` with imports/aliases from \`fastmcp_pvl_core\`.
- Adds a \`server: ServerConfig\` field on \`CollectionConfig\`, populated via \`ServerConfig.from_env(_ENV_PREFIX)\` in \`load_config()\`.
- Adds \`[tool.pyright]\` config so editor diagnostics use the uv-managed \`.venv\` (mirrors the same fix already applied in \`fastmcp-pvl-core\`).

Domain-specific fields on \`CollectionConfig\` (auth, OIDC, bearer, event_store_url) are intentionally kept duplicated for now — subsequent MV-PR2/3 migrate consumers (auth builders, middleware) to read from \`config.server.*\` and then drop the duplicates.

## Behaviour notes

- Core \`env()\` strips whitespace and treats empty values as unset. Setting \`MARKDOWN_VAULT_MCP_<BOOL>=\"\"\` on the four bool env vars (\`READ_ONLY\`, \`GIT_LFS\`, \`OIDC_VERIFY_ACCESS_TOKEN\`, \`OLLAMA_CPU_ONLY\`) now falls through to the configured default instead of being parsed as \`False\`. The previous behaviour was an accident of MV's \`_env\` returning the literal empty string.
- Local \`_parse_scopes\` wraps core's \`parse_scopes\` with \`or None\` to preserve MV's "blank input → None" contract — \`build_oidc_auth\` relies on it to fall back to \`['openid']\`.

## Test plan

- [x] \`uv run mypy src/markdown_vault_mcp/config.py\` — no errors
- [x] \`uv run pytest -x -q\` — 1397 passed (matches baseline)
- [ ] Smoke test in editor: imports resolve via Pyright

## Documentation

This PR keeps existing env-var docstrings unchanged because no env vars were added, removed, or renamed. Subsequent MV-PRs will likely add documentation as the migration progresses.

### DIFF
--- a/docs/superpowers/plans/2026-04-20-fastmcp-pvl-core-extraction.md
+++ b/docs/superpowers/plans/2026-04-20-fastmcp-pvl-core-extraction.md
@@ -1,0 +1,3002 @@
+# fastmcp-pvl-core Extraction Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Extract shared FastMCP infrastructure from `markdown-vault-mcp` into a new standalone PyPI package `fastmcp-pvl-core`, then adopt it back in MV as the first consumer, publishing both releases.
+
+**Architecture:** New repo `pvliesdonk/fastmcp-pvl-core` ships a narrow library of generic FastMCP primitives (auth, middleware, logging, config, server-factory helpers, artifact store, CLI helpers). MV imports from it using composition (never inheritance) via a ~30-line `make_server()` that wires building blocks together. No god-factory, no project-specific hooks in core.
+
+**Tech Stack:** Python 3.10+, FastMCP, uv, hatchling, ruff, mypy, pytest, python-semantic-release, PSR → PyPI. Identical toolchain to MV — this is a behavior-preserving extraction, not a rewrite.
+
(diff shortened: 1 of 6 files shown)


pvliesdonk:
Addressed all three review items in 8023942:

1. **Empty-bool behaviour tests** — added `TestEmptyBoolEnvVarsFallToDefault` with one test per affected field (READ_ONLY, GIT_LFS, OIDC_VERIFY_ACCESS_TOKEN, OLLAMA_CPU_ONLY). Locks in the new `empty → default` contract.
2. **`server` docstring entry** — added under `fastembed_cache_dir` in the `Attributes:` section, with a note that domain duplicates remain until MV-PR2/3 migrate consumers.
3. **`_parse_bool` / `_parse_list` re-export comment** — one-line explanation that no shim wrapper is needed because call sites already gate with `is not None`.

Also added two `TestServerConfigComposition` tests covering the default empty `ServerConfig` and `load_config()` populating it from env (`TRANSPORT`, `BEARER_TOKEN`, `BASE_URL`).

Suite is now 1403 passing (was 1397; +6 new). Lint/mypy clean.

PR #397, opened 2026-04-20
refactor(auth): delegate auth builders to fastmcp-pvl-core (MV-PR2)

pvliesdonk:
## Summary

Second PR in the fastmcp-pvl-core adoption series. Replaces MV's local \`resolve_auth_mode\` / \`build_remote_auth\` / \`build_bearer_auth\` / \`build_oidc_auth\` implementations with thin wrappers that delegate to the core library.

- **~280 lines deleted** — httpx discovery, OIDCProxy construction, JWTVerifier wiring, missing-var detection, secret redaction all live in core now.
- A new \`_server_from_collection(config)\` helper synthesises a transient \`ServerConfig\` view from the duplicated auth fields on \`CollectionConfig\`. The wrappers each pass that view to core and return what core returns.
- Public surface (\`resolve_auth_mode\`, \`build_*\`) keeps the same signatures and the same return types so \`mcp_server.create_server\` is unchanged.

## Behavioural notes

1. **\`resolve_auth_mode\` hides core's \`"multi"\`.** Core's resolver has 5 modes (\`none\` / \`bearer\` / \`remote\` / \`oidc-proxy\` / \`multi\`), but MV's caller composes multi-auth itself, so the wrapper re-resolves with the bearer token suppressed and returns only \`"remote"\` / \`"oidc-proxy"\` / \`None\`. Existing branches in \`create_server\` keep working unchanged.
2. **Discovery error class narrowed.** Core catches \`(httpx.HTTPError, ValueError)\`; MV used to catch bare \`Exception\`. A bare \`Exception\` from \`httpx.get\` now propagates by design — tests that simulated \"any failure\" updated to use \`httpx.ConnectError\`.
3. **Log-string format change.** Auth log lines moved from prose (\`\"Bearer auth: BEARER_TOKEN is set\"\`) to core's structured \`key=value\` form (\`\"bearer_auth_enabled token=<redacted>\"\`). Test assertions updated; the security-critical property — secrets never appearing in logs — is preserved verbatim with explicit \`secret-token not in caplog.text\` checks.
4. **3 log-text-only tests deleted as redundant.** \`test_verify_id_token_log_message\`, \`test_verify_access_token_log_message\`, and \`test_debug_logging_on_success\` only asserted log-string format; their behavioural intent (\`verify_id_token\` kwarg flips correctly, \`RemoteAuthProvider\` constructed on success) is already covered by sibling tests inspecting the kwargs passed to mocked classes.

The duplicate auth fields on \`CollectionConfig\` (\`base_url\`, \`bearer_token\`, \`oidc_*\`, \`auth_mode\`) are still populated by \`load_config()\` for now — a later MV-PR migrates \`mcp_server.create_server\` to read from \`config.server\` directly and drops the duplicates.

## Test plan

- [x] \`uv run pytest -x -q\` — 1400 passed (was 1397; -3 dropped log-text tests, +6 from MV-PR1 retained, but rebased against main so net -3 actually shows as -2 from baseline; final count 1400)
- [x] Total coverage 92.98% (≥80% gate)

## Documentation

No env vars added/removed/renamed. Auth public API surface unchanged (\`resolve_auth_mode\`, \`build_bearer_auth\`, etc. still importable with the same shapes from \`markdown_vault_mcp.config\`). Docs updates wait for the consumer-migration PR that drops the duplicates.

### DIFF
--- a/src/markdown_vault_mcp/config.py
+++ b/src/markdown_vault_mcp/config.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 
 import logging
 import os
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, replace
 from pathlib import Path
 from typing import Any
 
@@ -784,8 +784,6 @@ def resolve_auth_mode(config: CollectionConfig) -> str | None:
     Returns:
         ``"remote"``, ``"oidc-proxy"``, or ``None``.
     """
-    from dataclasses import replace
-
     server = _server_from_collection(config)
     mode = _core_resolve_auth_mode(server)
     if mode == "multi":
--- a/src/markdown_vault_mcp/config.py
+++ b/src/markdown_vault_mcp/config.py
@@ -8,12 +8,14 @@ from __future__ import annotations
 
 import logging
 import os
-import sys
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any
 
 from fastmcp_pvl_core import ServerConfig
+from fastmcp_pvl_core import build_bearer_auth as _core_build_bearer_auth
+from fastmcp_pvl_core import build_oidc_proxy_auth as _core_build_oidc_proxy_auth
+from fastmcp_pvl_core import build_remote_auth as _core_build_remote_auth
 from fastmcp_pvl_core import env as _core_env
 
 # Direct re-exports: parse_bool/parse_list match MV's old call shape (take a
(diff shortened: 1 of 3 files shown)


pvliesdonk:
Addressed both review items in 1d0586c:

1. **Hoisted `dataclasses.replace`** to the existing module-level import (no more in-function re-import on every `resolve_auth_mode` call).
2. **Strengthened `test_oidc_debug_does_not_leak_secret`** with `mock_cls.assert_called_once()` — the negative "secret not in logs" assertion alone would have passed for a silent-no-op regression; the positive check now guards that `OIDCProxy` actually gets constructed.

Suite still at 1400 passing; lint/mypy clean.

PR #398, opened 2026-04-20
refactor(server): adopt core wire_middleware_stack + configure_logging_from_env (MV-PR3)

pvliesdonk:
## Summary

Third PR in the fastmcp-pvl-core adoption series. Replaces MV's inline middleware-stack assembly and CLI logging bootstrap with the corresponding helpers from \`fastmcp-pvl-core==1.0.0rc1\`.

### \`mcp_server.py\`

- Drops the local \`ErrorHandling\` / \`Timing\` / \`Logging\` middleware imports + inline \`add_middleware\` calls (~20 lines).
- Single call to \`wire_middleware_stack(mcp, include_traceback=None, transform_errors=False)\`.
  - \`include_traceback=None\` lets the helper infer from the root log level at call time — matches MV's previous \`isEnabledFor(DEBUG)\` behaviour.
  - \`transform_errors=False\` preserves MV's default of letting exceptions propagate to FastMCP's own handlers (rather than synthesizing MCP error responses inside the middleware).
- Removes the now-unused \`os\` import.

### \`cli.py\`

- Drops the inline \`level = DEBUG if -v else INFO\` / \`setLevel\` / \`configure_logging('DEBUG')\` / \`FASTMCP_LOG_LEVEL=DEBUG\` dance in favour of \`configure_logging_from_env(verbose=args.verbose)\`. The helper sets the env var, sets the root level, and calls FastMCP's own \`configure_logging\` — same end-state as the inline code.
- Keeps the **root-StreamHandler fallback inline**: FastMCP's \`configure_logging\` targets its own logger tree, not root, so MV's \`markdown_vault_mcp.*\` loggers still need a root handler when run from a fresh CLI invocation.
- Keeps **httpx/httpcore noise-tuning inline** (per the adoption plan — these are MV-domain-specific dependencies, not core's concern).

### Tests

- \`test_verbose_enables_debug_for_both_logger_trees\` now patches \`configure_logging_from_env\` (kwarg-mode) instead of the bare \`configure_logging\` that was previously imported into the cli module.
- \`test_verbose_sets_fastmcp_log_level_env\` no longer needs to mock \`configure_logging\` — \`configure_logging_from_env\` handles the env-var side effect itself, and the assertion still holds.
- Existing \`TestMiddlewareStack\` tests (3 cases asserting the installed middleware classes + ordering) pass unchanged because \`wire_middleware_stack\` installs the same FastMCP middleware types in the same order.

## Test plan

- [x] \`uv run pytest -x -q\` — 1400 passed
- [x] Total coverage 92.96% (≥80% gate)

## Documentation

No env vars added/removed/renamed. \`FASTMCP_LOG_LEVEL\`, \`FASTMCP_ENABLE_RICH_LOGGING\`, and \`-v\` behave identically. No README/docs changes needed.

### DIFF
--- a/src/markdown_vault_mcp/cli.py
+++ b/src/markdown_vault_mcp/cli.py
@@ -15,7 +15,7 @@ import sys
 from dataclasses import asdict
 from pathlib import Path
 
-from fastmcp.utilities.logging import configure_logging
+from fastmcp_pvl_core import configure_logging_from_env
 
 from markdown_vault_mcp.collection import Collection
 from markdown_vault_mcp.config import _ENV_PREFIX, load_config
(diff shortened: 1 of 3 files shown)


pvliesdonk:
Addressed both review items in d97d5bb:

1. **Claude bot (CLAUDE.md hard rule)**: collapsed the two multi-line comment blocks in `cli.main()` to one short line each, preserving the WHY ("root handler for markdown_vault_mcp.*", "silence httpx/httpcore inline").

2. **Gemini inline**: hardened `test_verbose_sets_fastmcp_log_level_env`'s cleanup — now saves and restores `root.level` and `root.handlers` in addition to `FASTMCP_LOG_LEVEL`. The test intentionally keeps the real `configure_logging_from_env` unmocked because asserting the env-var side effect IS the contract MV depends on (a stronger test than mocking the helper); the tighter cleanup keeps global logging state pristine for siblings.

1400 tests still passing; lint/mypy clean.

pvliesdonk:
Fixed the missed block in b1b2ef9 — collapsed the 6-line comment above `wire_middleware_stack` in `mcp_server.py` to one line, preserving both WHYs (`include_traceback=None` inference from log level + `transform_errors=False` exception propagation).

PR #400, opened 2026-04-20
refactor(artifacts): adopt core ArtifactStore with eager bytes (MV-PR5)

pvliesdonk:
## Summary

Fifth PR in the fastmcp-pvl-core adoption series.

Replaces MV's local \`ArtifactStore\` (stored \`(path, created_at, ttl_seconds)\` and dereferenced through \`Collection\` at serve time) with \`fastmcp_pvl_core.ArtifactStore\` (stores eager bytes + filename + mime_type). This was decision #6 in the extraction plan.

- **\`artifacts.py\`**: now a thin re-export of core's \`ArtifactStore\` + \`TokenRecord\`. Keeps the module-level singleton (\`_artifact_store\`, \`set_artifact_store\`, \`get_artifact_store\`) so tool handlers outside FastMCP's DI context can reach the store used by the HTTP route. Adds \`ARTIFACT_TTL_SECONDS = 3600\` constant (single process-wide TTL — per-token TTL isn't supported upstream). Drops ~200 lines of local \`ArtifactStore\`, \`TokenRecord\`, \`set_collection_store\`, \`_get_collection_from_store\`, and \`make_artifact_handler\` logic.

- **\`mcp_server.py\`**: constructs the \`ArtifactStore\` at \`create_server()\` time (not lifespan) so the HTTP route closure can bind to a concrete instance. Uses \`ArtifactStore.register_route(mcp, store)\` instead of a custom \`mcp.custom_route\` + domain-specific handler.

- **\`_server_tools.py\`** (\`create_download_link\`): eagerly reads bytes via \`asyncio.to_thread(abs_path.read_bytes)\` after path validation, then calls \`store.add(content, filename, mime_type)\`.

- **\`_server_deps.py\`**: lifespan no longer touches the artifact store singleton — the store now lives for the server lifetime.

## Behavioural notes

1. **\`ttl_seconds\` tool arg is now advisory.** Core's \`ArtifactStore\` takes a single TTL at construction; all tokens share it. The tool still validates \`ttl_seconds > 0\` for backward compat, but the actual expiry is always \`ARTIFACT_TTL_SECONDS\` (3600s). The JSON response's \`expires_in_seconds\` reflects the *effective* TTL, not the requested one — this is honest (was 300s before). Docstring updated.

2. **No disk access at serve time.** Bytes are captured at tool-call time; the HTTP handler just returns stored bytes. File deletion after token creation no longer 404s — the token still resolves with the cached bytes until TTL expiry. This is a minor behaviour change from the old "revalidate at serve time" model; acceptable because the TTL is short.

3. **HTTP route closure binds at \`create_server\`.** Previously the handler read \`_collection_store\` via a module-level accessor at request time; now the route closes over the specific store instance, so the lifespan → handler handoff is removed.

## Test updates

\`tests/test_artifacts.py\` rewritten:
- Drops ArtifactStore-internal tests (create/consume/expiry/cleanup) — those are covered in core's own test suite.
- \`TestArtifactHandler\` rewritten to build a minimal Starlette app seeded directly via \`store.add(bytes, filename=, mime_type=)\`, without Collection wiring.
- \`test_expires_in_seconds_matches_ttl\` → \`test_expires_in_seconds_echoes_store_ttl\` (asserts store-wide TTL).
- Adds \`test_raises_runtime_error_when_store_unset\` for \`get_artifact_store\`'s error path.

## Test plan

- [x] \`uv run pytest -x -q\` — **1391 passed** (was 1400; −9 from deleted MV-internal tests)
- [x] \`diff-cover\` on changed lines: **100%**
- [x] Total coverage 93.08% (≥80% gate)

## Documentation

No env vars added/removed/renamed. The \`create_download_link\` tool's docstring now documents that \`ttl_seconds\` is advisory; docs site pages for tools will be refreshed in the MV-PR7 audit PR.

### DIFF
--- a/src/markdown_vault_mcp/_server_deps.py
+++ b/src/markdown_vault_mcp/_server_deps.py
@@ -76,22 +76,15 @@ def make_collection_lifespan(config: CollectionConfig) -> Any:
         # Start background tasks (e.g. git pull loop) after index is built.
         collection.start()
 
-        # Initialise artifact store singletons so the HTTP artifact endpoint
-        # can resolve vault paths outside of FastMCP's DI context.
-        from markdown_vault_mcp.artifacts import (
-            ArtifactStore,
-            set_artifact_store,
-            set_collection_store,
-        )
-
-        set_artifact_store(ArtifactStore())
-        set_collection_store(collection)
+        # Artifact store singleton is wired in create_server(), not here —
+        # the HTTP route captures the store at server-construction time and
+        # tool handlers reach it via get_artifact_store().  Tokens carry
+        # eager bytes now, so the lifespan no longer needs to expose the
+        # Collection to the HTTP handler.
 
         try:
             yield {"collection": collection, "config": config}
         finally:
-            set_artifact_store(None)
-            set_collection_store(None)
             collection.close()
             logger.info("Collection shut down")
 
--- a/src/markdown_vault_mcp/_server_tools.py
+++ b/src/markdown_vault_mcp/_server_tools.py
@@ -1473,14 +1473,20 @@ def _register_download_link_tool(mcp: FastMCP) -> None:
         Args:
             path: Vault-relative path to the file (e.g.
                 ``"notes/report.md"`` or ``"assets/diagram.png"``).
-            ttl_seconds: Link lifetime in seconds (default 300 / 5
-                minutes).
+            ttl_seconds: Requested link lifetime in seconds (default
+                300 / 5 minutes).  The server enforces a single
+                process-wide TTL on its artifact store; the actual
+                expiry returned in ``expires_in_seconds`` reflects that
+                store setting, which may differ from the requested
+                value.
 
         Returns:
             JSON-encoded string with the following fields:
 
             - download_url (str): One-time HTTP URL to download the file.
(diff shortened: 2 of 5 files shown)


pvliesdonk:
Addressed all four review items in 0fe2efc:

1. **Docstring drift (both bots)**: updated `set_artifact_store`'s docstring to match reality — the server owns the store for its lifetime; later `create_server` calls replace the singleton; nothing calls it with `None` in the normal lifecycle.
2. **Gemini: `abs_path.name`** — replaced `path.rsplit('/', 1)[-1]` with the idiomatic pathlib form.
3. **Claude: missing `Content-Disposition` assertion** — restored on `test_serves_attachment_bytes`.
4. **Claude: skip store on stdio** — `ArtifactStore` construction + `set_artifact_store` now guarded on `transport != 'stdio'` (matching the route and tool registration). Also added `test_artifact_route_end_to_end_via_http_app` — an integration-style test that hits the real route mounted by core's `register_route` (via `server.http_app()` + TestClient) rather than `TestArtifactHandler`'s hand-rolled handler, so regressions in core's route implementation can't hide behind MV's local test scaffold.

1392 tests passing (was 1391; +1 new integration test).

PR #401, opened 2026-04-20
refactor(cli): adopt core normalise_http_path + rename --path to --http-path (MV-PR6)

pvliesdonk:
## Summary

Sixth PR in the fastmcp-pvl-core adoption series. Narrower than originally planned because adopting \`make_serve_parser\` as the top parser would break every \`markdown-vault-mcp serve ...\` invocation in docs, Dockerfiles, and user setups — \`serve\` would stop being a subcommand.

### What this PR adopts

- Imports \`normalise_http_path\` from \`fastmcp_pvl_core\`; deletes MV's local \`_normalise_http_path\` and \`_DEFAULT_HTTP_PATH\` constant.
- Renames the \`--path\` flag to \`--http-path\` (matching core's \`make_serve_parser\` shape) and updates \`_cmd_serve\` to read \`args.http_path\`.
- Inline comment in \`_build_parser\`'s serve block notes that the flag set mirrors \`make_serve_parser\` on purpose, so a future \`add_serve_args(parser)\` helper in core can slot in as a drop-in replacement without another round of CLI churn.

### What this PR defers

- \`make_serve_parser\` adoption — blocked on core exposing an \`add_serve_args(parser)\` helper (or an \`add_help=False\` variant usable as an argparse parent). I'll file a follow-up issue on the core repo; when it lands, MV's serve subparser collapses to one call.

## Behavioural notes

- \`--path\` is renamed to \`--http-path\` (no alias kept — the two flag names differ in shape, not meaning). Users on Docker/systemd running \`markdown-vault-mcp serve --path /vault/mcp\` need to update to \`--http-path\`. The env-var fallback (\`MARKDOWN_VAULT_MCP_HTTP_PATH\`) is unchanged.
- \`serve\` remains an explicit subcommand — no \`markdown-vault-mcp <no subcommand>\` behaviour change.

## Test plan

- [x] \`uv run pytest -x -q\` — 1392 passed (was 1391; +1 for the reverted test rewrite path, count stabilises here)
- [x] Total coverage 93.11%

## Documentation

- \`README.md\`: \`--path\` → \`--http-path\` (flag column + subpath-mount example).

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,7 +259,7 @@ markdown-vault-mcp <command> [options]
 Start the MCP server.
 
 ```bash
-markdown-vault-mcp serve [--transport {stdio|sse|http}] [--host HOST] [--port PORT] [--path PATH]
+markdown-vault-mcp serve [--transport {stdio|sse|http}] [--host HOST] [--port PORT] [--http-path PATH]
 ```
 
 | Flag | Default | Description |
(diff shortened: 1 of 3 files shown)


pvliesdonk:
Both review items addressed in f189f8e:

1. **Gemini**: kept `--path` as an alias for `--http-path` so existing Dockerfiles, systemd units, and user scripts keep working across the rename. Added `test_serve_legacy_path_alias_still_works` to guard the alias contract. The argparse dest is still `http_path` — both spellings set the same attribute.

2. **Claude**: fixed the pre-existing README `--host` default discrepancy (was `0.0.0.0`, cli.py defaults to `127.0.0.1`). Added a note that `0.0.0.0` is the typical Docker value. Same row now documents `--path` as the legacy spelling for `--http-path`.

1393 tests passing; lint/mypy clean.

PR #413, opened 2026-04-21
fix(ci): docs deploy on v* tag push + 3-part concurrency

pvliesdonk:
## Summary

Port three docs workflow fixes from scholar-mcp (#154) and the template (pvliesdonk/fastmcp-server-template#20):

1. Add \`tags: [\"v*\"]\` to the push trigger so tag-push fires the workflow.
2. Extend \`github.event_name\` into the concurrency group so tag-push and release-event runs don't cancel each other.
3. Extend the gate on \`upload-artifact\` and the \`deploy\` job to accept tag-push on v* (not only \`event_name == 'release'\`).

## Why

The release-event path can race against tag availability in GitHub Pages' deploy-pages action (scholar v1.8.0-rc.1 + IG v1.8.0 both hit this).  Letting tag-push deploy too gives a deterministic second path — whichever event fires first wins the concurrency group and deploys.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -3,6 +3,13 @@ name: Documentation
 on:
   push:
     branches: [main]
+    # Tag-push also deploys: whichever event fires first (tag-push or
+    # release:published) wins the concurrency group below and performs
+    # the full build+upload+deploy.  The release-event path can race
+    # against tag availability (deploy-pages fetch fails), so tag-push
+    # gives a deterministic second path.  Gates on the `upload-artifact`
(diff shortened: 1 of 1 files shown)


pvliesdonk:
Closing — the actual root cause was the `github-pages` environment deployment-branch policy rejecting `v*` tag refs, not the workflow. Added `v*` tag policy to scholar + IG environments (MV already had it). The existing tag-push-as-smoke + 3-part concurrency from pvliesdonk/fastmcp-server-template#16 plus the ruleset fix is sufficient; the release-event path deploys fine once the environment allows tag refs. No workflow change needed.

PR #411, opened 2026-04-21
fix(release): publish linux packages on prerelease too

pvliesdonk:
## Summary

One-line fix: drop \`!inputs.prerelease\` gate from the \`publish-linux-packages\` job so .deb/.rpm artifacts attach to prerelease GitHub releases too.

## Why

Caught during Step 4 bootstrap-replay smoke: \`v1.26.0-rc.1\` shipped Docker + MCPB but no .deb/.rpm, which defeats the purpose of having an rc to validate the Linux packaging path against.

The gating across publish jobs is now consistent with their actual risk profile:

| Job | Runs on rc? | Reason |
|---|---|---|
| publish-pypi | no | immutable index |
| publish-docker | yes | GHCR is fine for rc tags |
| **publish-linux-packages** | **yes (this PR)** | just artifacts on a GH release |
| publish-mcpb | yes | artifact on the release |
| publish-claude-plugin-pr | no | opens PR to external repo |
| publish-registry | no | immutable index |

Mirrored from pvliesdonk/fastmcp-server-template#11. When that lands and we run \`copier update\` on MV, this file will no-op.

## Test plan

- [x] CI green
- [ ] Next rc release (e.g. \`v1.26.0-rc.2\` or \`v1.27.0-rc.1\`) attaches .deb/.rpm to the GitHub release

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #415, opened 2026-04-22
chore(copier): heavy backfill v1.0.0 → v1.1.1 + bootstrap copier-update workflow

pvliesdonk:
## Summary

Heavy backfill of template v1.0.0 → v1.1.0 for MV. Bootstraps the new \`.github/workflows/copier-update.yml\` workflow (shipped in template v1.1.0), advances \`.copier-answers.yml\` \`_commit:\` from v1.0.0 → v1.1.0, **and** restructures the hybrid files so future \`copier update\` runs do not silently overwrite project-owned divergences.

## Why heavy, not lazy

Testing a literal \`copier update\` from v1.0.0 → v1.1.0 showed it was destructive: license line silently flipped MIT → LicenseRef-Proprietary, argparse CLI replaced by typer scaffold. Without the backfill, every future weekly run would re-attempt those destructive merges and require manual undo.

## Divergence comments

- \`pyproject.toml\` — \`license = \"MIT\"  # project-owned — keep across copier updates\` and the same comment on the MIT classifier. Breaks line-level equality with template baseline so future patches conflict (operator decides) instead of silently overwriting.
- \`packaging/nfpm.yaml\` — same comment on \`license: \"MIT\"\`.

## Sentinel zones

- \`pyproject.toml\` — wrapped domain deps in \`# PROJECT-DEPS-START / # PROJECT-DEPS-END\` and optional-deps in \`# PROJECT-EXTRAS-START / # PROJECT-EXTRAS-END\`. Matches template layout; future template additions to deps land above the sentinel block and won't collide with MV's domain deps.

## CLI shim

\`src/markdown_vault_mcp/cli.py\` is now a 3-line re-export:

\`\`\`python
from markdown_vault_mcp._cli_impl import main
__all__ = [\"main\"]
\`\`\`

The actual implementation moved to \`_cli_impl.py\` (project-owned, not in template). Future \`copier update\` patches against \`cli.py\` have no shared context lines to anchor on, so all hunks reject cleanly into a \`.rej\` sidecar that the operator triages. This frees MV from the template's typer-rewrite trajectory.

## Related

- **Template PR #23** switches the \`copier-update.yml\` workflow from \`--conflict=rej\` (silent destruction + \`.rej\` documentation) to \`--conflict=inline\` (real 3-way merge with standard git conflict markers). Required for the divergence comments here to actually work on the weekly cron.
- Scholar PR #156 did the equivalent backfill via a real \`copier update\` (template v1.0.4 → v1.1.0 had no pyproject delta so no destructive merge risk).

## Test plan

- [x] \`uv run ruff check .\` → clean
- [x] \`uv run pytest -x -q\` → 1394 passed
- [x] Simulated future copier update w/ \`--conflict=inline\`: divergence on \`license\` line produces proper 3-way conflict marker instead of silent overwrite
- [ ] CI green
- [ ] Post-merge of template PR #23 + v1.1.1: re-render \`copier-update.yml\` here to pick up the \`--conflict=inline\` fix and advance \`_commit\` to v1.1.1


pvliesdonk:
Rebased onto origin/main (picked up PRs #411, #412, #414 that merged while this was under review) and addressed review feedback from the third pass:

- ✅ **PR title** updated from "v1.0.0 → v1.1.0" to "v1.0.0 → v1.1.1" to match \`_commit\`.
- ⏭ **Unpinned \`uvx copier\`** — deferring per the review's explicit "not a blocker for this PR since it's template-sourced, but worth a follow-up issue." Will open a template-side follow-up to pin \`copier\` to a minor version in the \`.jinja\` workflow, so the pin propagates on the next \`copier update\` rather than being re-clobbered each week.
- ℹ **\`print()\` in \`_cli_impl.py\` handlers** — left as-is. They're CLI user-facing terminal output (not server-side operational logging), which is the reasonable carve-out from the CLAUDE.md no-\`print()\` rule that the reviewer themselves called out.
- ℹ **Conflict marker grep pattern anchored on copier's exact wording** — noted as informational-only by the reviewer; low risk.

The prior-round \`--conflict=rej\` silent-destruction concern was already resolved by the v1.1.1 refresh commit (switches to \`--conflict=inline\`, copier exits 0, inline markers committed for operator review).

All gates still green locally:
- \`uv run ruff check .\` → clean
- \`uv run pytest -x -q\` → 1394 passed

PR #449, opened 2026-05-10
fix(ci): hotfix copier-update.yml — path-scoped Write for Jobs B and C

### LINKED ISSUE #448: Hotfix: copier-update.yml — Agent Jobs B/C lack Write permission, can't persist output JSON
## Summary

`.github/workflows/copier-update.yml`'s `claude_args` for Agent Jobs B (line 308) and C (line 342) declare:

```
--allowed-tools "Read,Bash(cat:*),Bash(git -C /tmp/template:*)"
```

Both jobs' prompts require writing JSON to `/tmp/agent-job-{b,c}.json`. Without `Write` in the allowlist every Write call is denied; the JSON files don't exist; the PR-body aggregator renders both sections as "⚠️ Agent failed".

Job A (line 274) already includes `Write`, which is why Job A produces real output.

## Evidence

- Run: https://github.com/pvliesdonk/markdown-vault-mcp/actions/runs/25635456782
- Resulting PR: #434
- Job B / Job C result: `"is_error": false`, `"permission_denials_count": 18`, no JSON written.

## Fix

Add `Write` to both allowlists.

## Note

Same hotfix-then-self-heal pattern as #446: applied directly to `.github/workflows/copier-update.yml`; once the upstream template fix in pvliesdonk/fastmcp-server-template lands and a new template version ships, the next `copier update` will overwrite this file with the (already-fixed) template version.

pvliesdonk:
## Summary

- Adds path-scoped Write — `Write(//tmp/agent-job-{b,c}.json)` — to the `--allowed-tools` allowlist for Agent Jobs B and C in `.github/workflows/copier-update.yml`.
- Job A's allowlist already had bare Write (it edits arbitrary working-tree files during conflict resolution); Jobs B and C had no Write at all, so their attempts to persist output JSON were denied.

## Why

Each Agent Job's prompt requires writing JSON output that the aggregator step then consumes. With Write missing, every write was denied (`permission_denials_count: 18` in run 25635456782) and the resulting PR body's two enrichment sections rendered as "⚠️ Agent failed". Path-scoping (rather than bare Write) preserves the prompt-injection defense — the agents read attacker-controllable inputs (cloned-template CHANGELOG, diff content) and bare Write would let injection redirect output to clobber other paths.

Closes #448.

## Note: hotfix temporarily diverges from the template

The same fix is being applied upstream in pvliesdonk/fastmcp-server-template#123 (which already closes #122 OIDC and #124 Write together). Once that lands and a new template version is released, the next `copier update` will overwrite this workflow file with the (already-fixed) template version, so no rollback or merge handling is needed.

The stale "Jobs B/C: read-only" comment block in this same workflow file is left untouched here on purpose — it's corrected in the upstream template; updating it here would create extra divergence that copier would re-conflict on.

## Test plan

- [ ] Trigger a manual `Copier Update` run after merge; confirm all three agent enrichment sections populate in the resulting PR body (not just Job A).

### DIFF
--- a/.github/workflows/copier-update.yml
+++ b/.github/workflows/copier-update.yml
@@ -305,7 +305,7 @@ jobs:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
           prompt: ${{ steps.prepare-prompt-b.outputs.prompt }}
           claude_args: |
-            --allowed-tools "Read,Bash(cat:*),Bash(git -C /tmp/template:*)"
+            --allowed-tools "Read,Write(//tmp/agent-job-b.json),Bash(cat:*),Bash(git -C /tmp/template:*)"
 
       - name: Pre-render Job C prompt
         id: prepare-prompt-c
@@ -339,7 +339,7 @@ jobs:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
           prompt: ${{ steps.prepare-prompt-c.outputs.prompt }}
           claude_args: |
-            --allowed-tools "Read,Bash(cat:*),Bash(git -C /tmp/template:*)"
+            --allowed-tools "Read,Write(//tmp/agent-job-c.json),Bash(cat:*),Bash(git -C /tmp/template:*)"
 
       - name: Ensure PR labels exist
         if: steps.changes.outputs.changed == 'true'


pvliesdonk:
On the double-slash path observation in the claude-review:

`Write(//tmp/agent-job-b.json)` is the **documented** Claude Code permission syntax for absolute paths. From [code.claude.com/docs/en/permissions](https://code.claude.com/docs/en/permissions) "Permission rule syntax":

| Pattern | Meaning | Example |
|---|---|---|
| `//path` | Absolute path from filesystem root | `Edit(//tmp/scratch.txt)` |
| `/path` | Path relative to project root | `Edit(/src/**/*.ts)` |

Single-slash `/tmp/agent-job-b.json` would be interpreted as **project-relative** (`<repo-root>/tmp/agent-job-b.json`), not as the absolute path the agent writes to. The double-slash is load-bearing.

pvliesdonk:
**Retraction notice (post-merge).**

[Earlier on this PR](https://github.com/pvliesdonk/markdown-vault-mcp/pull/449#issuecomment-4416028059) I defended the `Write(//tmp/...)` path-scoped form against claude-review's soft "double-slash seems suspicious" concern, citing the Claude Code permissions docs at code.claude.com/docs/en/permissions. The docs ARE consistent (they document `//absolute-path` as the correct absolute-path syntax), but the SDK shipping in `claude-code-action@v1` empirically doesn't reconcile that form with the agent's single-slash invocation. Diagnostic run #452 captured:

```
Agent invokes: Write(file_path="/tmp/agent-job-b.json")
SDK denies with: "Claude requested permissions to write to /tmp/agent-job-b.json,
                  but you haven't granted it yet."
```

The bot reviewer was right in practice. PR #454 (in flight) reverts to bare `Write` for Jobs B/C plus reverts #452's diagnostic flag. An upstream bug at `anthropics/claude-code-action` documenting the docs/SDK reconciliation gap is still to be filed.

PR #456, opened 2026-05-11
perf(tracker): pass exclude_patterns to detect_changes()

### LINKED ISSUE #257: perf: pass exclude_patterns to ChangeTracker.detect_changes()
## Problem

`ChangeTracker.detect_changes()` scans all `**/*.md` files under `source_dir` without filtering against `exclude_patterns`. Excluded files (e.g. `.claude/**`, `.obsidian/**`) are hashed on every `reindex()` call and reported as "added" — because they exist on disk but are never saved to tracker state (state is rebuilt from the FTS index, which correctly excludes them).

PR #256 added filtering in `reindex()` so these files are never parsed or indexed, and added a stale-doc purge pass for persistent indexes. The tracker churn is functionally harmless but wastes I/O on vaults with large excluded subtrees.

## Proposed Change

Add an `exclude_patterns` parameter to `ChangeTracker.detect_changes()`:

```python
def detect_changes(
    self,
    source_dir: Path,
    glob_pattern: str = "**/*.md",
    exclude_patterns: list[str] | None = None,  # NEW
) -> ChangeSet:
```

Apply `fnmatch` filtering (or delegate to `Collection._is_path_excluded()` extracted in PR #256) during the disk scan loop, before hashing.

Wire it in `Collection.reindex()`:
```python
changes = self._tracker.detect_changes(
    self._source_dir,
    exclude_patterns=self._exclude_patterns,
)
```

## Acceptance Criteria

- [ ] `detect_changes()` accepts `exclude_patterns` parameter
- [ ] Excluded files are not hashed or reported in `ChangeSet`
- [ ] Existing callers without `exclude_patterns` are unaffected (default `None`)
- [ ] Test: vault with excluded dir, verify `changes.added` does not contain excluded paths

## Verification

```bash
# After fix, excluded paths should not appear in ChangeSet:
uv run python -c "
from markdown_vault_mcp.tracker import ChangeTracker
from pathlib import Path
import tempfile

with tempfile.TemporaryDirectory() as d:
    p = Path(d)
    (p / 'note.md').write_text('# Note')
    (p / '.claude').mkdir()
    (p / '.claude/test.md').write_text('# Excluded')
    t = ChangeTracker(p / 'state.json')
    changes = t.detect_changes(p, exclude_patterns=['.claude/**'])
    assert '.claude/test.md' not in changes.added
    print('PASS')
"
```

## Documentation

No doc changes needed — internal optimization only.

---

Follow-up from PR #256 review (claude[bot] comment).

SYNTYZE:
## Summary

- Implements the perf follow-up tracked in #257 (originated as a review comment on PR #256 / issue #255).
- `ChangeTracker.detect_changes()` now accepts an optional `exclude_patterns` argument and skips matching files during the disk scan, so they are never hashed.
- Excluded paths are also kept out of the `deleted` bucket when a stale entry for them exists in `stored_state`, so the caller's purge pass in `IndexManager.reindex()` remains the single source of truth for cleaning up previously indexed (and now excluded) docs.
- Wires `self._exclude_patterns` through to the call site in `IndexManager.reindex()`.

## Motivation

`reindex()` already filters excluded files correctly thanks to #256, but the tracker still walks every `**/*.md` under `source_dir` and SHA-256-hashes them on each call. On a vault with a large excluded subtree (real-world trigger: a `brain/_archive/**` tree that contains a vendored `node_modules/` with thousands of READMEs/CHANGELOGs, several of which exceed the OpenAI embeddings 8192-token cap), each reindex pays significant I/O for files that will be thrown away. This change pushes the filter one layer down so excluded files are not even hashed.

## Behaviour

| Before | After |
|---|---|
| Every `**/*.md` under `source_dir` is hashed, even if it will be filtered out by `IndexManager`. | Files matching `exclude_patterns` are skipped before hashing. Disk I/O proportional to the *kept* set, not the total. |
| Stale `stored_state` entries pointing at excluded files appear in `ChangeSet.deleted`. | They are kept out of `deleted`; the index manager's stale-doc purge pass is the single owner of that cleanup. |
| `detect_changes(source_dir)` (no exclude arg) — every file reported. | Identical (default `exclude_patterns=None` preserves v1.6.0 semantics). |

## Test plan

- [x] New `TestExcludePatterns` test class in `tests/test_tracker.py` with 5 cases:
  - excluded files not in `added`
  - excluded files not hashed (spy on `_compute_hash`, verifies the I/O saving)
  - excluded files not in `modified` after on-disk content change
  - stale excluded `stored_state` entries kept out of `deleted`
  - legacy call without `exclude_patterns` unchanged
- [x] Manually verified on a real vault: previously failing `reindex()` (OpenAI 400 from a >8192-token excluded archive section) now completes cleanly; only kept files are hashed.

## Docs

- Updated the `tracker.py` signature in `docs/design.md` to include the new parameter.

## Notes for reviewers

- Default value is `None`, so the public signature stays additive — no existing caller breaks.
- The `deleted`-bucket exclusion is intentional and documented in the docstring. Happy to revert that half if you'd prefer `detect_changes` stay symmetric and instead handle the stale-entry case purely in `IndexManager`; the current shape just felt safer because it makes "excluded paths never appear in the ChangeSet" a single, testable invariant.

Fixes #257.


ghost:
Closing — pulling this back for internal review on our side first. Will re-open if/when ready. Thanks!

PR #457, opened 2026-05-11
feat(uploads): adopt register_file_exchange_upload (#443)

### LINKED ISSUE #443: Add an upload-link tool: HTTP POST one-time URL for local agents to push files into the vault
## Context

Today the inbound paths to the vault are:

- `write(content)` for markdown — fine
- `write(content_base64)` for binaries — context-bound, broken above ~1 MB (see #442)
- `fetch(url)` — pulls from a URL; unusable when the file is on the local agent's disk or has no public URL.

A local coding agent that wants to drop a screenshot, a generated PDF, or any binary into the vault has no good option.

## Approach

Adopt **`register_file_exchange_upload`** from `fastmcp-pvl-core` v2.1.0 (closed pvliesdonk/fastmcp-pvl-core#64) via the **`fastmcp-server-template` v1.6.0** scaffold (closed pvliesdonk/fastmcp-server-template#117).  No local POC.

### Step 1 — Copier update v1.5.1 → v1.6.0

`copier update` brings in:
- `DOMAIN-FILE-EXCHANGE-START / END` sentinel block in `server.py` wrapping the existing `register_file_exchange(...)` call.
- Fully-commented `register_file_exchange_upload(...)` scaffold inside the same sentinel, with stub `_upload_receiver` and `_validate_upload_target`.
- `fastmcp-pvl-core>=2.1.0,<3` pin in `pyproject.toml`.
- 4 new env vars documented in `docs/configuration.md`.
- "Uploading files (`receiver=`)" section in `docs/guides/file-exchange.md`.
- `## File Exchange` section in `CLAUDE.md`.

### Step 2 — Uncomment the scaffold

Mechanical: remove the `#` prefix on the `register_file_exchange_upload(...)` call inside the sentinel.

### Step 3 — Supply the real receiver + validator

In `src/markdown_vault_mcp/uploads.py` (or fold into `_server_deps.py`):

```python
def _vault_upload_receiver(record: UploadRecord, body: bytes) -> dict[str, Any]:
    """Commit uploaded bytes to the vault.

    Dispatches to ``Collection.write`` for ``.md`` paths or
    ``Collection.write_attachment`` for binaries based on the
    target_id's extension.
    """
    collection = get_collection_singleton()
    if record.target_id.endswith(".md"):
        collection.write(record.target_id, content=body.decode("utf-8"))
    else:
        collection.write_attachment(record.target_id, body)
    return {"path": record.target_id, "size_bytes": len(body)}


def _validate_upload_target(target_id: str) -> None:
    """Reject path traversal, disallowed extensions, etc. before link creation.

    Runs at link-creation time via pvl-core's ``pre_link_validator=``,
    so the agent gets the rejection in-band as a tool error rather
    than after a wasted HTTP POST.
    """
    collection = get_collection_singleton()
    collection._validate_path(target_id)  # path traversal + relative-path guard
    if not target_id.endswith(".md"):
        collection._check_attachment_extension(target_id)  # allowlist
```

The `pre_link_validator=` raises `ValueError` for invalid input — pvl-core surfaces it as in-band tool error per the v0.4 spec contract.

### Buffered vs streaming receiver

pvl-core ships both `receiver=` (buffered) and `stream_receiver=` (chunked) variants.  MV uses the buffered form — Collection's write/write_attachment APIs take `bytes`, and the size cap (default 10 MB, configurable via `MARKDOWN_VAULT_MCP_FILE_EXCHANGE_UPLOAD_MAX_BYTES`) keeps the buffer bounded.

## Documentation

Most docs come for free via the v1.6.0 template adoption.  MV-specific additions:

- `README.md` — new `create_upload_link` row in the MCP Tools table.
- `docs/tools/index.md` — full tool entry with parameters, returns, security notes.
- `docs/guides/claude-desktop.md` — the local-coding-agent upload flow as an example (paired with `curl -X POST --data-binary @...`).

## Tests

- `tests/test_uploads.py` — MV-side concerns: receiver dispatches to the right `Collection` method by extension, path validation rejects traversal at link creation, attachment allowlist enforced at link creation, end-to-end (`create_upload_link` → curl POST → file appears in vault → `read` returns it).
- The pvl-core helper's own behavior (token lifecycle, HTTP status codes, size enforcement) is tested upstream — MV doesn't re-test it.

## Sequencing

1. Copier update PR (v1.5.1 → v1.6.0).  Standalone — picks up the scaffold + pin bump + sentinel.  Behaviorally inert (scaffold stays commented).
2. This PR — uncomments the scaffold + adds the real receiver/validator + tests + MV-specific docs.

Could combine into one PR; splitting keeps the copier update reviewable on its own.

pvliesdonk:
Closes #443.

Adopts `fastmcp-pvl-core` 2.1.0's `register_file_exchange_upload` via the v1.6.1 template scaffold (landed in #455). Gives a local agent a one-time HTTPS POST URL (`create_upload_link`) for pushing files into the vault without round-tripping bytes through MCP context.

## Summary

- **New tool `create_upload_link(target_id, ttl_seconds)`** — mints a one-time `POST /markdown-vault-mcp/uploads/{token}` URL (HTTP/SSE only; requires `MARKDOWN_VAULT_MCP_BASE_URL`). Tagged `write` — hidden in read-only mode.
- **New `_vault_upload_receiver`** in `markdown_vault_mcp.uploads` — extension-dispatched commit: `.md` → `Collection.write` (utf-8 decoded), other → `Collection.write_attachment` (raw bytes).
- **New `_validate_upload_target`** in same module — runs as `pre_link_validator=` so path-traversal attempts and disallowed extensions raise `ValueError` at `create_upload_link` call time, not after a wasted POST.
- **New `get_collection_singleton` accessor** in `_server_deps.py` — the HTTP route handler runs outside FastMCP's `Depends(get_collection)` injection; same pattern as `artifacts.py`'s singleton.
- **End-to-end test**: mint URL → POST bytes via httpx → file lands → readable via `read`.
- **Regression test** for the `transport=\"auto\"` footgun (mirrors image-gen-mcp's pattern: derive `fx_transport` from `make_server`'s arg and pass explicitly, since `\"auto\"` reads env vars the CLI doesn't set).
- **Docs**: `create_upload_link` row in README + full entry in `docs/tools/index.md` with curl example, validation details, and an admonition about the per-extension size cap; new \"Uploading a local file to the vault\" workflow in `docs/guides/claude-desktop.md`; updated `CLAUDE.md`/`README.md`/`docs/configuration.md`/`docs/guides/file-exchange.md` to reflect that upload IS now wired (download still deferred to #431).

## Test plan

- [x] \`uv run pytest -x -q\` — 1485/1485 green (8 new tests in \`test_uploads.py\` + 2 in \`test_server_deps.py\`)
- [x] \`uv run mypy src/ tests/\` clean
- [x] \`uv run pre-commit run --all-files\` clean
- [x] \`diff-cover ... --fail-under=80\` — 100% on patch
- [x] Local review circus (3 rounds, 6 dispatches): both \`pr-review-toolkit:code-reviewer\` and \`pr-review-toolkit:silent-failure-hunter\` APPROVED on \`49adfa8\`
- [x] Regression-test sanity-check: temporarily reverted server.py to confirm the new \`test_create_upload_link_registered_via_make_server_arg\` test FAILS on the old wiring, PASSES on the new
- [ ] CI green
- [ ] Bot reviewers LGTM

## Follow-ups deferred to separate issues

- #458 — Bypass the in-Collection \`MAX_ATTACHMENT_SIZE_MB\` cap for non-\`.md\` uploads. Currently both \`write(content_base64=...)\` and \`create_upload_link\` for attachments are bounded by the same 1-MiB default, defeating the upload path's intent for large binaries.
- #459 — Widen the \`fx_transport\` mapping to include \`\"streamable-http\"\` (or fail loudly on unknown transport values). CLI argparse \`choices\` make this unreachable in production today.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -172,15 +172,27 @@ across `copier update`, so opt-in customisations (`produces=`,
 download-direction `register_file_exchange(...)` call itself) survive
 subsequent template updates.
 
-> **Project-specific override:** unlike the template default, **neither
-> direction is currently wired in markdown-vault-mcp**. The download
-> direction is deferred to #431 because the spec-compliant
-> `create_download_link(origin_id, ttl_seconds)` tool collides on name
-> with MV's existing `create_download_link(path, ttl_seconds)` (registered
-> via `ArtifactStore` in the `DOMAIN-WIRING` block). The upload direction
-> ships fully commented-out and is intended to be uncommented alongside
-> the #431 migration. See [`docs/guides/file-exchange.md`](docs/guides/file-exchange.md)
-> for the wiring pattern.
+> **Project-specific override:** unlike the template default, the two
+> directions are wired asymmetrically.
+>
+> - **Upload direction is wired** as of #443:
+>   `register_file_exchange_upload(...)` is active in
+>   `server.py`, with `_vault_upload_receiver` /
+>   `_validate_upload_target` from `markdown_vault_mcp.uploads` as the
+>   receiver and pre-link validator. The route auto-mounts only when
+>   transport is HTTP/SSE *and* `MARKDOWN_VAULT_MCP_BASE_URL` is set.
+> - **Download direction is still deferred to #431** because the
+>   spec-compliant `create_download_link(origin_id, ttl_seconds)` tool
+>   collides on name with MV's existing
+>   `create_download_link(path, ttl_seconds)` (registered via
+>   `ArtifactStore` in the `DOMAIN-WIRING` block). Do not add
+>   `register_file_exchange(mcp, ...)` to the sentinel block until #431
+>   resolves the collision.
+>
+> See [`docs/guides/file-exchange.md`](docs/guides/file-exchange.md)
+> for the wiring pattern, and the
+> [`create_upload_link` tool entry](docs/tools/index.md) for the
+> agent-facing contract.
 
 ## Shared Infrastructure
 
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Point it at a directory of Markdown files (an Obsidian vault, a docs folder, a Z
 - **Attachment support** — read, write, delete, and list non-markdown files (PDFs, images, etc.)
 - **Git integration** — optional auto-commit and push on every write via `GIT_ASKPASS`
 - **OIDC authentication** — optional token-based auth for HTTP deployments (Authelia, Keycloak, etc.)
-- **MCP tools** — 28 LLM-visible tools including search, read, write, edit, delete, rename, git history, and admin operations; plus 6 app-only tools for MCP Apps clients
+- **MCP tools** — 29 LLM-visible tools including search, read, write, edit, delete, rename, git history, file-exchange uploads, and admin operations; plus 6 app-only tools for MCP Apps clients
 - **MCP resources** — 9 resources exposing vault configuration, statistics, tags, folders, document outlines, similar notes, recent notes, and an interactive SPA
 - **MCP prompts** — 6 prompt templates including template-driven note creation
 <!-- DOMAIN-END -->
@@ -170,17 +170,19 @@ The server registers a built-in `get_server_info` tool (via `fastmcp_pvl_core.re
 ### File exchange
 
 A `DOMAIN-FILE-EXCHANGE-START` / `DOMAIN-FILE-EXCHANGE-END` sentinel
-block in `server.py` reserves space for the
+block in `server.py` wires the
 [MCP File Exchange](docs/guides/file-exchange.md) helpers from
-`fastmcp-pvl-core`. **Neither direction is currently wired in this
-server** — the download-direction `register_file_exchange(...)` call is
-deferred to #431 (name collision with MV's existing
-`create_download_link` tool registered via `ArtifactStore`); the
-upload-direction `register_file_exchange_upload(...)` ships fully
-commented-out and is intended to be uncommented alongside the #431
-migration. See the guide for producing / consuming / uploading
-patterns, or [`CLAUDE.md`](CLAUDE.md#file-exchange-register_file_exchange--opt-in-upload)
-for the wiring pattern.
+`fastmcp-pvl-core`. The **upload direction is wired** (#443) — agents
+mint a one-time HTTPS POST URL via `create_upload_link(target_id,
+ttl_seconds)` and push bytes outside the MCP context, with the route
+mounted only on HTTP/SSE transports when `MARKDOWN_VAULT_MCP_BASE_URL`
+is set. The **download direction remains deferred to #431** because
+the spec-compliant `create_download_link(origin_id, ttl_seconds)` tool
+collides on name with MV's existing `create_download_link(path,
+ttl_seconds)` registered via `ArtifactStore`. See the guide for the
+upload-flow walkthrough and the producer / consumer patterns waiting on
+#431, or [`CLAUDE.md`](CLAUDE.md#file-exchange-register_file_exchange--opt-in-upload)
+for the wiring overview.
 
 ## Configuration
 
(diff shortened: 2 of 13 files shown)


pvliesdonk:
Thanks for the review!

**Addressed in [\`0979aa7\`](https://github.com/pvliesdonk/markdown-vault-mcp/commit/0979aa7):**

1. ✅ **UnicodeDecodeError → 400 verification.** Added \`test_invalid_utf8_md_upload_returns_400\` that POSTs invalid UTF-8 to a \`.md\` upload via httpx and asserts HTTP 400. The local silent-failure-hunter pre-PR review had already verified the mapping by reading pvl-core's \`_file_exchange_runtime.py:825-831\` (catches \`ValueError\` → 400 with \`str(exc)\` body), and \`UnicodeDecodeError\` is a \`ValueError\` subclass — so the docs claim is accurate. The new test pins it: if pvl-core ever changes the mapping, the test fails loudly with the actual status code in the assertion message.

2. ✅ **Slash/\`..\` rejection source.** Added a one-line note to \`docs/tools/index.md\`'s Validation paragraph clarifying that rule (a) (segment-rule rejections) is enforced by pvl-core's \`ExchangeURI.validate_segment\` *upstream* of MV's pre-link callback. Operators debugging \"why is the error message phrased differently\" now have a pointer.

**Defended in writing (no code change):**

3. **Test access to private \`_collection_singleton\`** — leaving as-is. The save/restore pattern in \`tests/test_uploads.py\` mirrors the established precedent in \`tests/test_artifacts.py\` (\`TestGetArtifactStoreUninitialised\`) for the inverse direction. Adding a \`get_collection_singleton_or_none()\` helper would expand the public API surface for test ergonomics; the local code-quality reviewer flagged this on Task 1 but explicitly deferred (\"Stylistic; leave as-is\"). If we add a similar helper, both \`artifacts.py\` and \`_server_deps.py\` should get one in a separate hygiene PR.

4. **\`fx_transport\` doesn't cover \`\"streamable-http\"\`** — filed as #459. The CLI's argparse \`choices=[\"stdio\", \"sse\", \"http\"]\` makes this unreachable in production today; #459 widens the mapping (or fails loud) for third-party embedders.

Local circus re-ran clean on \`0979aa7\` (both \`pr-review-toolkit:code-reviewer\` and \`pr-review-toolkit:silent-failure-hunter\` APPROVED, no findings).

pvliesdonk:
Thanks — agreed APPROVED.

On the residual observation about \`tests/test_server_deps.py:37\` using direct assignment vs \`set_collection_singleton(saved)\`: I'll leave it as-is for this PR. Mild defense for the asymmetry: in \`test_server_deps.py\` the singleton accessor IS the system-under-test, so the cleanup uses raw assignment to keep the restore step from indirectly going through code-under-test. In \`test_uploads.py\` the singleton is just plumbing for the receiver-under-test, so the public setter is fine. The two patterns are intentional, not drift.

If we ever revisit (e.g. as part of #458's potential helper-API expansion), we can pick one or add a \`get_collection_singleton_or_none()\` helper that lets both files use it cleanly.

Flipping to ready.

pvliesdonk:
Thanks both — gemini APPROVED with no comments, claude-review APPROVED with 3 non-blockers + 1 action item.

**Action item (Dependency Audit gate):** filed as #460 — pre-existing pip CVE in the build environment, applies to \`main\` and every PR since #433, not introduced by this diff.  Out-of-scope for #443.

**Observations (non-blockers, no action this PR):**
1. Private \`Collection._validate_path\` / \`_validate_attachment_path\` coupling — agreed, defensible for now; if/when #458 reshapes the validator surface, can promote to public.
2. \`read_only\` re-check in receiver — agreed by-design (tag-driven \`mcp.disable\` gates the tool at registration); could add a docstring sentence in a future PR but not load-bearing.
3. Singleton save/restore asymmetry — defended on thread above (intentional: test_server_deps tests the accessor itself, so cleanup uses raw assignment to keep restore independent of code-under-test).
4. document.py error message — confirmed correct (was the explicit fix in \`ea85614\`).

Ready for human merge.  All required CI green; \`Dependency Audit\` failure is pre-existing and out-of-scope (tracked in #460).

PR #496, opened 2026-05-15
fix(chunker): word-budget fallback so no chunk exceeds max_chunk_words

### LINKED ISSUE #489: Ollama embeddings: oversize chunks silently truncated past model context limit
## Symptom

When `OllamaProvider` is the embedding backend (e.g. `nomic-embed-text-v1.5`), large chunks are silently truncated by Ollama because the chunk exceeds the model's native training context window. The vault still indexes "successfully" — but vectors for any oversize chunk represent only the *beginning* of the content, degrading semantic-search quality with no error or warning surfaced.

Re-indexing the same vault with FastEmbed (which handles long inputs differently / has different tokenizer limits) does not exhibit the problem, which is what confirmed the root cause.

## Evidence

`ai-ollama` (ollama/ollama:0.20.4) logs during index build, vault running `nomic-embed-text-v1.5`:

```
time=2026-05-14T06:07:10.593Z level=WARN source=server.go:169
  msg="requested context size too large for model" num_ctx=8192 n_ctx_train=2048
```

Ollama logs the warning and proceeds; it will still accept the request but the model only attends to the first `n_ctx_train` (2048 here) tokens of each input. Our side never sees an error.

## Root cause

Two compounding factors in `markdown-vault-mcp`:

### 1. Chunker has no hard upper bound

`HeadingChunker._refine_oversize` (`src/markdown_vault_mcp/scanner.py:223–255`) recursively descends H1→H6 to split oversize leaves, but when the deepest matching level is reached and a section *still* exceeds `max_chunk_words` with no deeper headings inside, the chunk is emitted as-is:

```python
sub_chunks = self._split_at_levels(
    sub_lines, levels=(next_level,), base_line=chunk.start_line + 1
)
if not sub_chunks:
    # No headings of next_level inside; cannot split further.
    out.append(chunk)
    continue
```

Default `max_chunk_words=400` (`collection.py:152`) — around 500–600 tokens for typical English, fits in 2048 with room to spare. But a flat 5000-word section with no sub-headings (a long prose journal entry, an LLM-generated note, an academic abstract, an imported web page) goes to the embedding provider as one chunk and gets silently truncated.

### 2. `OllamaProvider` has no context-window awareness

`src/markdown_vault_mcp/providers.py:80–146` — the provider:
- Never queries the model's context via `/api/show` (which exposes `parameters.num_ctx` and `model_info.<arch>.context_length`)
- Never validates input length before POSTing to `/api/embed`
- Never sets `options.num_ctx`, so it inherits whatever Ollama defaults to (`num_ctx=8192` in current Ollama builds — which itself triggers the WARN above on `n_ctx_train=2048` models)
- Treats a 200 OK as success regardless of whether the input was truncated upstream

Combined, an oversize chunk produces a "successful" embedding of the first ~2048 tokens. The remainder is invisible to semantic search.

## Suggested mitigations

Order of cheapness-to-implement; pick one minimal fix + one defensive logging step:

1. **Hard fallback split in `_refine_oversize`** (cheap, deterministic) — when no deeper heading exists but the chunk still exceeds `max_chunk_words`, fall back to a paragraph-then-word-budget splitter. Guarantees no chunk exceeds the configured budget. Closes the bug at the source for *all* providers, not just Ollama.
2. **Pre-flight token-budget log** — before submitting a batch, log a `WARNING` if any input's word count exceeds a sane budget (e.g. ~500 words ≈ 1500 tokens) so operators see something rather than nothing.
3. **Provider-aware context probe** (followup) — `OllamaProvider` calls `/api/show` once on init to learn `model_info.<arch>.context_length`, sets `options.num_ctx = min(default, n_ctx_train)` on requests (silences the upstream WARN), and rejects/truncates inputs above that bound with an explicit log.
4. **Sanity bound on the env var path** — `MARKDOWN_VAULT_MCP_MAX_CHUNK_WORDS` is already configurable; documenting recommended ceilings per known embedding model in `docs/configuration.md` would help operators tune defensively before #1 lands.

(1) is the right minimum fix — the chunker's invariant should be "no chunk exceeds the budget", full stop. The recursive heading split is best-effort for *semantic* boundaries; a final word-budget split is the safety net.

## Documentation

- `docs/configuration.md` — add a note on `MAX_CHUNK_WORDS` interacting with the chosen embedding model's context window, with a table of common models (`nomic-embed-text-v1.5` n_ctx_train=2048, `all-MiniLM-L6-v2` 512, etc.)
- `docs/guides/` — if there's an embeddings guide, mention silent-truncation as a failure mode to watch for
- `CHANGELOG.md` — entry under fix once a PR lands

## Verification

After fix:

```bash
# Synthetic vault with one note containing a single H1 and 10000-word body.
uv run pytest tests/test_scanner_chunker.py -k oversize_leaf_fallback
# Confirms every emitted chunk has <= max_chunk_words.

# Manual: run with OllamaProvider against nomic-embed-text-v1.5, watch
# ollama logs — should no longer see "requested context size too large
# for model" once OllamaProvider sets num_ctx (mitigation 3), and even
# without that, no chunk should be large enough to be truncated.
```

## Provenance

Found during real-world use of an Obsidian vault against this MCP server. The Ollama log line above is from the running `ai-ollama` container; the chunker behaviour confirmed by reading `scanner.py:223–255` and `collection.py:152`.

pvliesdonk:
## Summary

- New `_budget_split` helper in `HeadingChunker` (`src/markdown_vault_mcp/scanner.py`) that fragments any oversize chunk on paragraph and word boundaries when heading-based refinement can't make further progress.
- Wired at every chunker exit path: `_refine_oversize` leaves (heading present, no deeper level), preamble chunks (heading=None), the short-doc bypass when the single chunk exceeds budget, and the no-headings-at-all return.
- Result: `len(chunk.content.split()) <= max_chunk_words` now holds for every emitted chunk regardless of source structure — eliminating silent truncation at the embedding step for providers with context limits (e.g. Ollama serving `nomic-embed-text-v1.5` with `n_ctx_train=2048`, or FastEmbed's default `BAAI/bge-small-en-v1.5` with a 512-token context).

Closes #489.

## Doc updates

- `docs/configuration.md` — `MAX_CHUNK_WORDS` now described as a hard cap with embedding-provider context-window guidance. Leads with the default FastEmbed model (`BAAI/bge-small-en-v1.5`, 512 tokens) and notes that the default `400` words already exceeds it — recommends lowering to ~300 if staying on the default model. Frames `nomic-embed-text-v1.5` (8192 native; Ollama `n_ctx_train=2048` default) as the long-context alternative.
- `docs/design.md` — Adaptive heading-level chunking sections updated to describe the fallback and the hard-cap invariant.

## Migration

A reindex is required for existing vaults to benefit (already documented in the upgrade note under MAX_CHUNK_WORDS). Operators on the default FastEmbed model should consider lowering MAX_CHUNK_WORDS to ~300 alongside the reindex.

## Related

- `OllamaProvider` context-window probe (logging the model's `n_ctx_train` at startup, optionally setting `options.num_ctx`) is the natural follow-up — the chunker fix removes silent truncation at the source for all providers, which is the load-bearing piece; the probe is polish.
- The budget-split can fragment a single section into several rows in the FTS `sections` table. `read(path, section=heading)` returns only the first fragment (existing `ORDER BY start_line ASC LIMIT 1` semantics) — this is the same limitation tracked under #490 (section-read brittleness); not a regression but more reachable now.

## Test plan

- [x] `uv run pytest` full suite — 1570 pass
- [x] `uv run ruff check .` / `ruff format --check .` — clean
- [x] `uv run mypy src/ tests/` — clean
- [x] `diff-cover` — 98.5% patch coverage (only uncovered line is a defensive whitespace-only early return that's unreachable through current callers)
- [x] Local code-review circus — both reviewers clean after one polish iteration (default-model doc fix, nomic context-value reconciliation, bin-packing flush test)
- [ ] CI green
- [ ] Bot reviews (claude-review, gemini-code-assist) clean on first pass


pvliesdonk:
## Re: gemini finding — word-split fragments share `start_line` (search.py ordering)

Investigated. The finding is technically valid but its correct fix is **out of scope for this chunker PR** — deferred to **#498**.

**Why it can't be fixed correctly inside the chunker:**

- Word-split fragments of a single oversize line genuinely *begin on the same source line* — there is no integer line number between line N and line N+1 to distribute them across.
- Giving them distinct integer `start_line` values collides with the real subsequent lines (fragment K of line N would land on line N+K, interleaving with genuine content there).
- Renumbering downstream content to make room makes `start_line` diverge from real source line numbers for genuine lines — and `start_line` is surfaced in `SearchResult` / `SectionHit` and consumed by `read(path, section=)`.

**The correct fix is a search-layer change:** thread `sections.id` (rowid, monotonic with chunker emission order) through the FTS + vector result pipelines and add it as the final tiebreaker in `_group_by_path`'s sort, plus `, rowid` in the FTS5 `ORDER BY` for upstream determinism. That touches the query SELECTs, the row protocols, and the grouping helper — a coherent search-layer change, not a chunker change. Filed as **#498** with the full scope.

**Practical blast radius until #498 lands:** near-zero. The tie requires (a) two chunks from the *same source line* — which only the word-split fallback produces, for a pathological single line longer than the entire `max_chunk_words` budget — *and* (b) those fragments tying on bm25/cosine score. `_group_by_path` already uses a stable sort, so the order is deterministic given a deterministic upstream query; the only genuine gap is FTS5's unspecified secondary order on exact rank ties. No data corruption, no silent truncation — at worst, two adjacent fragments of one pathological line could appear swapped in a search result.

This PR's scope (the chunker hard-cap) is unchanged and complete.

PR #502, opened 2026-05-15
fix(edit): report the real divergent line in multi-line edit diagnostics

### LINKED ISSUE #501: edit diagnostic is first-line-only: misleading 'first_diff_at_char' on multi-line old_text
## Problem

When `edit()` fails to match a multi-line `old_text`, the `EditConflictError` diagnostic produced by `find_closest_match()` (`utils/text.py`) is misleading.

Observed in production logs:

```
error=old_text not found in TNO/PMC Cybersecure Products and Systems - Theory of Change.md
closest_match_line: 289
first_diff_at_char: 17
expected: '### The six links'
found: '### The six links'
```

`expected` and `found` are byte-identical, yet the edit failed — which reads as a false negative.

## Root cause

`find_closest_match()` only ever inspects the **first line** of `old_text` (`old_text.split("\n", 1)[0]`). For a multi-line `old_text` whose first line matches the file perfectly:

1. It anchors on that perfectly-matching file line (here line 289, `### The six links`).
2. The char-diff loop `while diff_pos < min_len and a[i] == b[i]` exits with `diff_pos = min_len` when the two lines are *equal* — so `first_diff_at_char: 17` just means "identical for all 17 chars", not "differs at 17".
3. `expected_snippet` / `found_snippet` are therefore two identical strings.

The actual divergence is on some *later* line of `old_text`, which the diagnostic never examines. The log gives zero information about the real mismatch.

## Fix

Make `find_closest_match()` multi-line aware: anchor on the first line, then walk `old_text`'s lines against the file region to report the **first line that genuinely diverges** — its file line number, the char offset within that line, and snippets of that line pair. Non-printable / invisible characters are already escaped at the call site via `{...!r}` in `_server_tools.py`, so a real invisible-char difference on the divergent line becomes visible once the correct line is reported.

Scoped to the diagnostic only — matching/normalization behaviour is unchanged.

## Documentation

- `docs/design.md` — clarify diagnostic-field semantics for multi-line `old_text`.

## Verification

- New tests in `tests/test_utils_text.py` covering multi-line `old_text` where the first line matches but a later line diverges, and where `old_text` extends past the file region.

pvliesdonk:
## Summary

`find_closest_match()` (the source of `EditConflictError`'s diagnostic) only ever inspected the **first line** of `old_text`. When that line matched the file exactly — common for a multi-line `old_text` — the char-diff loop returned `min_len` and the diagnostic emitted two byte-identical snippets with a meaningless `first_diff_at_char`, reading as a false negative:

```
closest_match_line: 289
first_diff_at_char: 17
expected: '### The six links'
found: '### The six links'
```

- Anchor on the first line, then walk `old_text`'s lines against the file region to report the **first line that genuinely diverges** (file line number + char offset within that line). Invisible-character differences become visible because `_server_tools.py` already `repr()`s the snippets.
- Return an empty dict (no diagnostic) when every line of `old_text` matches the file region — instead of falling back to the last line with two identical snippets, which would reintroduce the very bug being fixed. Consistent with the existing no-anchor `{}` return.
- Docstrings (`find_closest_match`, `EditConflictError`) and `docs/design.md` updated to match.

Closes #501

## Test plan

- [x] `tests/test_utils_text.py::TestFindClosestMatch` — new tests: later-line divergence, perfectly-matching first line with a U+200B difference on a later line, `old_text` extending past the file region, single-line and multi-line all-lines-match returning `{}`
- [x] `uv run pytest -x -q` — 1583 passed
- [x] `uv run ruff check` / `ruff format --check` / `uv run mypy src/ tests/` — clean
- [x] `diff-cover` patch coverage — 100%

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -1024,11 +1024,14 @@ one normalized match is found, the original byte range is replaced and
 if the file does not exist. Raises `EditConflictError` if `old_text` is
 not found (after both exact and normalized matching) or appears more than
 once. When both exact and normalized match fail, `EditConflictError`
-includes diagnostic fields: `closest_match_line`, `first_diff_char`,
+carries optional diagnostic fields: `closest_match_line`, `first_diff_char`,
 `expected_snippet`, `found_snippet`. For a multi-line `old_text` these
 locate the *first line that genuinely diverges* from the file — the
 diagnostic anchors on the first line, then walks subsequent lines so a
 later-line mismatch is reported rather than a perfectly-matching first line.
+The fields are omitted (left `None`) when no divergence can be localized:
+no file line is similar enough to anchor on, or every line of `old_text`
+matches the file region.
 
 **`delete()` behavior**: removes the file from disk, deletes FTS and embedding
 entries, triggers `on_write`. Raises `DocumentNotFoundError` if not found.
--- a/docs/design.md
+++ b/docs/design.md
@@ -1025,7 +1025,10 @@ if the file does not exist. Raises `EditConflictError` if `old_text` is
 not found (after both exact and normalized matching) or appears more than
 once. When both exact and normalized match fail, `EditConflictError`
 includes diagnostic fields: `closest_match_line`, `first_diff_char`,
-`expected_snippet`, `found_snippet`.
+`expected_snippet`, `found_snippet`. For a multi-line `old_text` these
+locate the *first line that genuinely diverges* from the file — the
+diagnostic anchors on the first line, then walks subsequent lines so a
+later-line mismatch is reported rather than a perfectly-matching first line.
 
 **`delete()` behavior**: removes the file from disk, deletes FTS and embedding
 entries, triggers `on_write`. Raises `DocumentNotFoundError` if not found.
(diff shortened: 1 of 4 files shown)


pvliesdonk:
Re: the non-blocking observation about an untested "fuzzy anchor + later-line divergence" path — that path is unreachable, so no test is added.

The walk only advances past `offset 0` when `old_lines[0]` is **byte-identical** to the anchor line (`old_line != file_line` is the break condition). If `old_lines[0]` only *fuzzy*-matches the anchor (ratio < 1.0), it is by definition not identical, so the walk breaks at `offset 0` and the divergence is reported on the first line. A later-line divergence therefore always implies an exact (ratio 1.0) anchor.

The suggested example confirms this — running it gives `closest_match_line == 2` (divergence at the fuzzy first line), not `3`:

```
find_closest_match("helllo world\nthis line is wrong",
                   "preamble\nhello world\nthis line is right\n")
# -> {'closest_match_line': 2, 'first_diff_char': 4, ...}
```

The two reachable combinations are already covered: exact anchor + later-line divergence (`test_multiline_reports_divergent_line`, `test_multiline_first_line_match_not_reported_as_diff`) and fuzzy anchor + first-line divergence (`test_close_match`, `test_diff_position_reported`).

PR #510, opened 2026-05-21
Avoid eager embedding build during MCP startup

### LINKED ISSUE #509: MCP handshake timeout on large vaults: startup scan blocks initialize response
## Summary

With large vaults (e.g. >5,000 documents / >500,000 chunks), `markdown-vault-mcp serve` fails to respond to the MCP `initialize` message within the client's internal handshake timeout (~30 s). Claude Desktop shows **"Could not attach to MCP server markdown-vault-mcp"** approximately 2 minutes after launch — the timeout fires while the server is still scanning the vault.

## Root cause

The FastMCP lifespan in `_server_deps.py` calls `collection.build_index()` and `collection.build_embeddings()` **synchronously before `yield`**, meaning the server cannot handle any MCP message (including `initialize`) until both calls complete.

For a large vault this takes minutes:

1. **`collection.build_index()`** — calls `IndexManager.build_index()` which runs `scan_directory()` over the entire vault on **every startup**, even when an existing on-disk FTS index is already fully populated. The `Collection._initialized` flag that would short-circuit this is always `False` at construction time, so the guard in `Collection.build_index()` (lines 492–507) never fires on first boot.

2. **`collection.build_embeddings()`** — when the `.npy` vector file does not yet exist on disk, `_load_vectors()` creates an empty `VectorIndex` (count=0), causing `build_embeddings()` to attempt embedding the **entire corpus inline** rather than deferring to an offline script.

The `timeout` field in `claude_desktop_config.json` does **not** extend Claude Desktop's internal MCP handshake deadline, so increasing it has no effect on this error.

## Reproduction

1. Configure `markdown-vault-mcp` pointing at a vault with >5,000 markdown files and an embedding provider (Ollama / bge-m3 or similar).
2. Build the FTS index once via the CLI or the overnight indexing script so the `.db` file exists, but leave the vector files (`.npy` / `.json`) absent.
3. Restart Claude Desktop. After ~30 s the "Could not attach" banner appears.

### 1. `collection.py` `__init__` — detect pre-populated DB at construction time

After `FTSIndex` is constructed (and therefore the SQLite connection is open), probe whether the on-disk DB is already populated:

```python
# Fast-path: if the on-disk FTS index already contains documents,
# mark as initialised immediately so build_index(force=False) skips
# the full startup scan.
self._initialized = False
if index_path is not None and index_path.exists():
    try:
        cur = self._fts._conn.execute("SELECT 1 FROM documents LIMIT 1")
        if cur.fetchone() is not None:
            self._initialized = True
    except Exception:
        pass  # Malformed DB: fall through to normal init
```

This lets the existing guard in `Collection.build_index()` fire correctly on every startup once the DB has been built at least once.

### 2. `collection.py` `build_index()` — replace `list_notes()` with `COUNT(*)`

In the no-op path the current code calls `self._fts.list_notes()` which materialises **all** document rows. Replace with a single-row count query:

```python
if not force and self._initialized:
    row = self._fts._conn.execute("SELECT COUNT(*) FROM documents").fetchone()
    doc_count = row[0] if row else 0
    if doc_count > 0:
        return IndexStats(documents_indexed=doc_count, chunks_indexed=0, skipped=0)
```

### 3. `managers/index.py` `build_embeddings()` — skip if `.npy` absent

When `force=False` and the vector file does not yet exist on disk, return 0 immediately rather than embedding the whole corpus inline. The initial build belongs in an offline/overnight script; once the file exists, subsequent startups load it in milliseconds.

```python
npy_path = Path(str(self._embeddings_path) + ".npy")
if not npy_path.exists():
    logger.info(
        "build_embeddings: %s.npy missing — skipping eager build to avoid "
        "blocking MCP startup. Run the offline indexing script first.",
        self._embeddings_path,
    )
    return 0
```

## Impact

With all three patches applied, startup time drops from **~2 minutes (→ timeout)** to **< 1 second**. FTS search is immediately available; semantic search becomes available after the offline indexing script completes and the `.npy` file is created.

## Notes

- The vault uses a **symlink farm** pointing at multiple source directories; this multiplies the effective file count seen by `scan_directory()` and makes the scan particularly slow.
- The `IndexManager.build_index()` method (line 246 in v1.28.0) runs `source_dir.glob("**/*.md", recurse_symlinks=True)` **after** the exclude filter to count skipped files — this traverses the entire tree regardless and adds additional latency even when most files are excluded.

## Version

`markdown-vault-mcp` 1.28.0, Python 3.14, macOS, Claude Desktop (MCP client).

Baijack-star:
## Summary
- seed Collection initialization from an existing persistent FTS DB so warm indexes skip startup scans after process restart
- use COUNT(*) for the build_index no-op path instead of materializing every note
- skip eager build_embeddings() during server lifespan when the vector .npy sidecar is absent, leaving first corpus embedding to explicit offline commands/tools

## Tests
- uv run --extra embeddings-api pytest tests/test_collection.py::TestBuildIndexNoOp tests/test_server_deps.py -q
- uv run ruff check src/markdown_vault_mcp/collection.py src/markdown_vault_mcp/_server_deps.py tests/test_collection.py tests/test_server_deps.py

Fixes #509

### DIFF
--- a/src/markdown_vault_mcp/_server_deps.py
+++ b/src/markdown_vault_mcp/_server_deps.py
@@ -111,7 +111,8 @@ def make_collection_lifespan(config: CollectionConfig) -> Any:
         # Build embeddings eagerly only when a vector sidecar already exists.
         # Initial corpus embedding can take minutes or hours and must not block
         # the MCP initialize handshake; use the explicit indexing command/tool.
-        if kwargs.get("embedding_provider") is not None:
+        embeddings_path = kwargs.get("embeddings_path")
+        if kwargs.get("embedding_provider") is not None and embeddings_path is not None:
             if collection.has_embedding_index_sidecar():
                 chunks_embedded = await asyncio.to_thread(collection.build_embeddings)
                 logger.info("Embeddings ready: %d chunks", chunks_embedded)
@@ -119,7 +120,7 @@ def make_collection_lifespan(config: CollectionConfig) -> Any:
                 logger.info(
                     "Skipping eager embedding build because %s.npy is missing; "
                     "run the indexing command/tool to build vectors offline",
-                    kwargs.get("embeddings_path"),
+                    embeddings_path,
                 )
 
         # Start background tasks (e.g. git pull loop) after index is built.
--- a/src/markdown_vault_mcp/_server_deps.py
+++ b/src/markdown_vault_mcp/_server_deps.py
@@ -108,12 +108,19 @@ def make_collection_lifespan(config: CollectionConfig) -> Any:
             stats.chunks_indexed,
         )
 
-        # Build embeddings eagerly when an embedding provider is configured.
-        # build_embeddings() skips work if the vector index already exists on disk,
-        # so this is safe to call on every startup.
+        # Build embeddings eagerly only when a vector sidecar already exists.
+        # Initial corpus embedding can take minutes or hours and must not block
+        # the MCP initialize handshake; use the explicit indexing command/tool.
         if kwargs.get("embedding_provider") is not None:
-            chunks_embedded = await asyncio.to_thread(collection.build_embeddings)
-            logger.info("Embeddings ready: %d chunks", chunks_embedded)
+            if collection.has_embedding_index_sidecar():
+                chunks_embedded = await asyncio.to_thread(collection.build_embeddings)
+                logger.info("Embeddings ready: %d chunks", chunks_embedded)
+            else:
+                logger.info(
+                    "Skipping eager embedding build because %s.npy is missing; "
+                    "run the indexing command/tool to build vectors offline",
+                    kwargs.get("embeddings_path"),
+                )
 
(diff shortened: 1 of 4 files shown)


Baijack-star:
Addressed the review feedback in `88d5042`:

- only probe/build startup embeddings when both `embedding_provider` and `embeddings_path` are configured, avoiding the misleading `None.npy` log path;
- simplified the SQLite `COUNT(*)` helper since `fetchone()` always returns the aggregate row;
- added a regression that a provider without `embeddings_path` does not probe the sidecar or build embeddings during startup.

Verified again:

```bash
uv run --extra embeddings-api pytest tests/test_collection.py::TestBuildIndexNoOp tests/test_server_deps.py -q
uv run ruff check src/markdown_vault_mcp/collection.py src/markdown_vault_mcp/_server_deps.py tests/test_collection.py tests/test_server_deps.py
```

pvliesdonk:
Completely fixed in v2.0.0-rc1 with a complete refactor

PR #515, opened 2026-05-26
feat(server): non-blocking background indexing (closes #513)

### LINKED ISSUE #513: Non-blocking startup: background-build FTS index and embeddings (supersedes #509)
## Summary

### New `Collection` state

```python
```

### New `Collection` methods

```python

```

```python

collection.start()
```

### Failure handling

### Shutdown

## CLI unchanged

## Acceptance criteria

## Out of scope

## References

pvliesdonk:
## Summary

Closes #513 (supersedes #509; replaces the abandoned PR #510).

Moves FTS index and embedding builds off the startup hot path. The MCP `initialize` handshake now completes within seconds regardless of vault size; a daemon thread runs `reindex()` then `build_embeddings()` after the handshake. Status surfaces via `stats://vault → index_status`.

Preserves the warm-start fast path: pre-built persistent state loads in sub-second; the background reindex confirms no drift. Preserves the PR #256 `exclude_patterns` purge invariant via `Collection.build_index()`'s warm-start no-op-with-purge logic (the regression that killed PR #510).

(Re-opened from PR #514 which was closed when its branch was renamed — `+` in the original branch name caused 403s in the CI checkout step. Same commits.)

## Highlights

- **`Collection.initialize_async()`** — synchronous persistent-state probe; sub-second; no filesystem scan.
- **`Collection.schedule_background_reindex()`** — spawns a daemon thread running `reindex()` then `build_embeddings()`. Status transitions `indexing → embedding → ready`; failure → `failed`. Shutdown via `_index_shutdown` + 30s join in `Collection.close()`.
- **`_fts_done_event` / `_index_done_event` split** — foreground tool calls unblock as soon as FTS phase completes, not waiting for embeddings (preserves the "partial FTS results during indexing" promise).
- **`Collection.build_index()` warm-start fast path** — replaces the prior `list_notes()` materialization with `COUNT(*)` + DB-only `_purge_excluded_rows()`; keeps PR #256's exclude purge invariant.
- **`Collection.build_embeddings(skip_if_missing=)`** — opt-in non-blocking variant for direct callers.
- **`Collection._ensure_initialized`** — waits on `_fts_done_event` with bounded 60s timeout (warning + inline fallback on timeout); self-thread guard prevents worker self-deadlock.
- **`Collection.stats()`** — no longer calls `_ensure_initialized`; non-blocking during cold-start so `stats://vault` stays a fast status probe.
- **`stats://vault` resource** — gains `index_status` field with shape `{status, indexed, error}` (typed via `IndexStatusPayload` TypedDict).
- **Lifespan** — `_server_deps.py` rewired: `initialize_async()` + `schedule_background_reindex()` replace the blocking `build_index` + `build_embeddings`.

## Test plan

- [x] 1613 tests pass locally (was 1456 on main; +157 new tests cover the state machine, lifespan rewire, stats non-blocking, warm-start purge, etc.)
- [x] `ruff check`, `ruff format --check src/ tests/`, `mypy src/ tests/` — all clean
- [ ] CI passes on Python 3.11, 3.12, 3.13, 3.14
- [ ] `claude-review` and `gemini-code-assist` bot bodies LGTM
- [ ] CHANGELOG entry auto-generated from conventional commits

## Notes

- 18 commits with clean conventional-commit granularity — PSR will compute the next minor bump from the `feat:` commits. Prefer merge (not squash) per repo policy so PSR walks individual messages.
- Documentation updated: `docs/guides/claude-desktop.md`, `docs/configuration.md`, `docs/design.md`, `docs/resources.md`.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -13,7 +13,7 @@ All configuration is via environment variables. Most use the `MARKDOWN_VAULT_MCP
 | `MARKDOWN_VAULT_MCP_STATE_PATH` | path | `{SOURCE_DIR}/.markdown_vault_mcp/state.json` | No | Path to the change-tracking state file |
 | `MARKDOWN_VAULT_MCP_INDEXED_FIELDS` | csv | — | No | Comma-separated frontmatter fields to promote to the tag index for structured filtering |
 | `MARKDOWN_VAULT_MCP_REQUIRED_FIELDS` | csv | — | No | Comma-separated frontmatter fields required on every document; documents missing any are excluded from the index |
-| `MARKDOWN_VAULT_MCP_EXCLUDE` | csv | — | No | Comma-separated glob patterns to exclude from scanning (e.g. `.obsidian/**,.trash/**`) |
+| `MARKDOWN_VAULT_MCP_EXCLUDE` | csv | — | No | Comma-separated glob patterns to exclude from scanning (e.g. `.obsidian/**,.trash/**`).  Changes take effect on the next background reindex (server start or periodic git pull); removed patterns may take longer to surface previously-excluded files than added patterns take to purge them |
 | `MARKDOWN_VAULT_MCP_TEMPLATES_FOLDER` | string | `_templates` | No | Relative folder path used by the `create_from_template` prompt to discover/read template files |
 | `MARKDOWN_VAULT_MCP_PROMPTS_FOLDER` | path | — | No | Path to a directory of `.md` prompt files that extend or override built-in prompts |
 
--- a/docs/design.md
+++ b/docs/design.md
@@ -467,6 +467,19 @@ attachment write operations.
 `read()` validates the path inline rather than via `_validate_path()`: if the
 resolved path escapes `source_dir`, it returns `None` instead of raising.
 
+### Startup sequence (server lifespan)
+
+The MCP server lifespan delegates the slow parts of indexing to a background thread so the MCP handshake is never blocked:
+
+1. Construct `Collection` — probes the persistent FTS DB and seeds `_initialized`.
+2. `sync_from_remote_before_index()` — pull from the configured remote (if any).
+3. `collection.initialize_async()` — no-op probe so warm-start callers see the persistent state.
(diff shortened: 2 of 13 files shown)


pvliesdonk:
@gemini-code-assist Fix applied in 9ae54e9 per your high-priority finding. The background worker now gates the embedding phase on both `_embedding_provider` AND `_embeddings_path` being set, with an INFO log when skipping due to the missing path. Added `test_schedule_background_reindex_skips_embeddings_without_path` to lock in the contract. PTAL.

pvliesdonk:
Closing per CLAUDE.md's "Know when to abandon the branch" rule. This PR took 6 fix rounds and the diff turned into sediment of patches against a structural design flaw, not isolated bugs.

**Root cause** (captured for the next attempt): I made `Collection._ensure_initialized` wait on the background thread's `_fts_done_event`. That foreground/background coupling generated every concentric problem — the two-event split, the 60 s timeout fallback that itself races on `_write_lock`, the broad-`except Exception` masks in `_count_documents`/`_fts_has_documents` (damage control for "we can't tell uninitialized from corrupted"), the `stats()` bypass of `_ensure_initialized` that created a dual-invariant universe, and the `skip_if_missing` parameter that became dead production code.

**Cleaner shape for the next attempt**:
- `Collection` stays strictly sync, no internal threading state. `_ensure_initialized` either raises a clear "not built yet" error or runs `build_index()` inline. No waits, no events, no shutdown flag.
- A separate `BackgroundIndexer` class wraps a Collection: owns the daemon thread, status state machine, shutdown event, bounded join in close().
- Lifespan composes them: build Collection → probe persistent state → spawn BackgroundIndexer → yield.
- Foreground tool calls hit whatever state the Collection currently has — zero rows during cold start is a valid signal (FTS returns empty results); operators see indexing progress via `stats://vault.index_status` from the BackgroundIndexer.
- Task 2's warm-start fast path with `exclude_patterns` purge in `Collection.build_index()` stays — that part is sound.

The first attempt was a spike, not lost work. Findings carry over to the re-implementation; only the messy diff is discarded.

Reopens #513.

PR #516, opened 2026-05-26
feat(server): non-blocking background indexing, second attempt (closes #513)

### LINKED ISSUE #513: Non-blocking startup: background-build FTS index and embeddings (supersedes #509)

pvliesdonk:
## Summary

Replaces the abandoned PR #515 with a structurally different approach:
indexing stays in `Collection`, but the public API gains
`start_background_reindex()` (fire-and-forget daemon spawn) and
`index_status()` (read-only snapshot) — and the foreground waiter
`_ensure_initialized()` is **removed entirely** (22 call sites stripped).

The no-foreground-waiter rule that PR #515 violated is now executable:
`test_foreground_reads_never_block_on_background` asserts `search` /
`list` / `stats` / `index_status` each return within 500ms while a
deliberately slow background embedding pass is in flight. Future code
re-introducing a waiter will fail this test.

Spec: `docs/superpowers/specs/2026-05-26-background-indexing-v2-design.md`
Plan: `docs/superpowers/plans/2026-05-26-background-indexing-v2.md`

Closes #513.

## Architecture

| | |
|---|---|
| Sync mode | `build_index()` / `reindex()` — unchanged; used by CLI + `reindex` MCP tool |
| Background mode | `start_background_reindex()` — daemon thread runs `reindex()` then `build_embeddings()` (if provider configured); idempotent; returns immediately |
| Status | Nested in `CollectionStats.index_status` (LLM accesses via `stats` tool) and surfaced under `get_server_info`'s upstream slot |
| Shutdown | `close()` signals shutdown, joins with 30s bounded timeout, daemon-thread fallback; SQLite WAL handles partial uncommitted state on next startup |
| Rule | No public method on `Collection` may block on background indexing state. Empty-during-cold-start is a valid result, not an error to wait out. |

## Local review

Two rounds of the `preflight-circus` skill (5 core lenses + 4 supplementary)
ran before push. All findings at confidence ≥ 80 addressed in commits
`1b4744b` and `b9fe6e0`.

## Deferred follow-ups (filed as comments below or to be split into issues post-merge)

These were surfaced by the local review at sub-threshold or as architectural
deltas, intentionally not bundled:

1. **Type-design**: `CollectionStats.index_status` is typed `dict[str, Any]`
   and `_background_phase` is `str | None` with an implicit
   `{"indexing", "embedding", None}` domain. A frozen `IndexStatus` dataclass
   + `Literal["indexing", "embedding"]` would give compile-time guarantees.
   Separate refactor.
2. **Test coverage**: Concurrent foreground writes during background
   reindex, and `close()` during in-flight writes, are not directly
   exercised. Real production paths; require a slow-callback test
   harness.
3. **Observability**: Categorising specific exception types in the
   daemon worker (`MemoryError`, `sqlite3.DatabaseError`, `OSError`)
   for severity escalation. Currently all go through the broad
   `except Exception` boundary.
4. **`_closed` flag**: `start_background_reindex()` races trivially
   with `close()` — adding a `_closed` flag would harden it. Single-shot
   shutdown in practice, but worth a follow-up if anyone hits it.

## Test plan

- [x] `uv run pytest -x -q` — 1601 passed, 1 skipped
- [x] `uv run mypy src/ tests/` — clean
- [x] `uv run ruff check . && uv run ruff format --check .` — clean
- [x] Foreground reads return < 500ms during slow background embedding (the load-bearing test)
- [x] `_ensure_initialized` removed from collection.py and all 22 call sites
- [x] Bounded `close()` join: 30s timeout, daemon-thread fallback
- [x] `index_status` exposed via both `stats` and `get_server_info`
- [x] Two rounds of preflight-circus (5 core lenses + 4 supplementary) cleared at ≥80 confidence

### DIFF
--- a/.claude/scheduled_tasks.lock
+++ b/.claude/scheduled_tasks.lock
@@ -1,0 +1,1 @@
+{"sessionId":"b0c2a392-87fc-41c9-8e2c-d7a596000a80","pid":254519,"procStart":"543893","acquiredAt":1778439832390}
--- a/.claude/scheduled_tasks.lock
+++ b/.claude/scheduled_tasks.lock
@@ -1,1 +1,0 @@
-{"sessionId":"b0c2a392-87fc-41c9-8e2c-d7a596000a80","pid":254519,"procStart":"543893","acquiredAt":1778439832390}
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,4 @@ site/
 .claude/worktrees/
 .claude/settings.local.json
 docs/superpowers/
+.claude/
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ Point it at a directory of Markdown files (an Obsidian vault, a docs folder, a Z
 > **Upgrading.** As of this release, `search` returns query-relevant snippets in the `content` field by default (approximately 200 words). Pass `snippet_words=0` to recover the prior full-chunk behaviour, or use `read(path, section=heading)` to fetch a specific chunk after seeing a snippet. Documents are also re-chunked on next `reindex` to honour the adaptive `MARKDOWN_VAULT_MCP_MAX_CHUNK_WORDS` threshold (default 400).
 - **Frontmatter-aware** — indexes YAML frontmatter fields, supports required field enforcement
 - **Incremental reindexing** — hash-based change detection, only re-processes modified files
+- **Non-blocking startup** — the server starts immediately and builds its index in the background; tools are usable straight away and clients can check progress via the `stats` tool's `index_status` field
 - **Write operations** — create, edit, delete, rename documents with automatic index updates
 - **Attachment support** — read, write, delete, and list non-markdown files (PDFs, images, etc.)
 - **Git integration** — optional auto-commit and push on every write via `GIT_ASKPASS`
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -13,7 +13,7 @@ All configuration is via environment variables. Most use the `MARKDOWN_VAULT_MCP
 | `MARKDOWN_VAULT_MCP_STATE_PATH` | path | `{SOURCE_DIR}/.markdown_vault_mcp/state.json` | No | Path to the change-tracking state file |
 | `MARKDOWN_VAULT_MCP_INDEXED_FIELDS` | csv | — | No | Comma-separated frontmatter fields to promote to the tag index for structured filtering |
 | `MARKDOWN_VAULT_MCP_REQUIRED_FIELDS` | csv | — | No | Comma-separated frontmatter fields required on every document; documents missing any are excluded from the index |
-| `MARKDOWN_VAULT_MCP_EXCLUDE` | csv | — | No | Comma-separated glob patterns to exclude from scanning (e.g. `.obsidian/**,.trash/**`) |
+| `MARKDOWN_VAULT_MCP_EXCLUDE` | csv | — | No | Comma-separated glob patterns to exclude from scanning (e.g. `.obsidian/**,.trash/**`). Changes take effect on the next background reindex (server start or periodic git pull). |
 | `MARKDOWN_VAULT_MCP_TEMPLATES_FOLDER` | string | `_templates` | No | Relative folder path used by the `create_from_template` prompt to discover/read template files |
 | `MARKDOWN_VAULT_MCP_PROMPTS_FOLDER` | path | — | No | Path to a directory of `.md` prompt files that extend or override built-in prompts |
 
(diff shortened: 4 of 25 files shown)


pvliesdonk:
Abandoning this PR per CLAUDE.md's "two or three rounds" rule. This is the second attempt at #513 (after PR #515 was also abandoned), and round 3 of the local preflight-circus introduced a real bug while trying to fix earlier-round findings — the `_closed` flag I added has a TOCTOU race because the write isn't inside `_background_state_lock`. That's exactly the "sediment of fixes-of-fixes" failure mode the rule exists to catch.

Closing without merging. Issue #513 stays open. A fresh attempt should start from `main` with no commits from this branch.

## What was tried (for the next attempt to evaluate, NOT to adopt blindly)

- Pure-sync `Collection` + `start_background_reindex()` method on `Collection` itself (no wrapper class).
- `_ensure_initialized()` waiter removed entirely (22 call sites stripped) on the theory that the no-foreground-waiter rule should be a structural property, not discipline.
- An executable regression test (`test_foreground_reads_never_block_on_background`) using a deliberately slow embedding provider, asserting `search/list/stats/index_status` return within 500ms during background work.
- Frozen `IndexStatus` dataclass + `BackgroundPhase = Literal["indexing", "embedding"]` for typed status.
- `_closed` flag for shutdown-race prevention (this is the one that turned out to be broken).
- Status surfaced via two paths: nested in `CollectionStats.index_status` (LLM-callable via `stats` tool) and through `register_server_info_tool`'s upstream slot (`upstream_label="index_status"`).
- Lifespan: fire-and-forget `start_background_reindex()`, no synchronous build at startup.

**Whether any of the above is the right shape is for the next attempt to re-decide.** This branch is one data point, not a validated design. PR #515's data point is the other.

## What this attempt's preflight surfaced (the next attempt should not recur these)

1. **TOCTOU race on `_closed`**: writing the flag outside the same lock that the reader uses doesn't actually serialise the post-close-spawn check. Whatever shutdown-race prevention the next attempt builds, the write-set and the read-set must share their locking.
2. **`_IDLE_INDEX_STATUS` overloading `last_error`** to carry "collection not yet initialised" conflates lifecycle state with run-failure state — every monitor that alerts on `last_error != None` will false-positive during startup. A distinct lifecycle field is cleaner; or use the all-`None`-timestamps shape as the implicit signal.
3. **Lifespan fire-and-forget** removes the previous synchronous-build's loud-startup-failure mode. A misconfigured embedding provider, unreadable source dir, etc., now manifest as "server up, every query returns empty, only `index_status.last_error` says why." The next attempt should decide whether to keep that contract or restore an explicit fail-fast path for clear configuration errors.
4. **Worker doesn't categorise exception types**: `MemoryError`, `sqlite3.DatabaseError`, `OSError` all go through the same broad `except Exception` boundary. Fine as a daemon-thread guard but defers severity escalation that production might need.
5. **Test gaps repeatedly flagged but never closed in this branch**: concurrent foreground write during background reindex; close() during in-flight callback; state-lock invariant specifically during the indexing phase (the load-bearing test only exercised the embedding phase). The next attempt should design these in from the start, not retrofit.
6. **Documentation rot from the strip**: the `_ensure_initialized` removal left stale references in `Collection.__init__` comment, `IndexManager.build_index` docstring, `docs/design.md` (two paragraphs), `_server_deps.py` lifespan comment, and `tests/test_collection.py`. Any future strip-style refactor should grep + sweep call sites and docstrings as one task, not two.
7. **MCP-client tests need a "wait for background ready" helper**: this branch added `mcp_client_ready` as a context manager in `tests/conftest.py`. Whether the next attempt needs this depends on its lifespan design, but if tests use `Client(server)` and immediately call tools, they'll hit empty results during cold start. Plan for this in the test design.
8. **Pyright (stricter than mypy) flags `dict[str, Any]` returns and `from tests.conftest import X` imports**. The diff passed mypy but pyright drew attention to several places. The next attempt could decide whether to tighten types beyond mypy's defaults.

## Local review

Three rounds of `preflight-circus` (5 core lenses + 4 supplementaries) ran on this branch. Each round produced findings; round 3 found a fix from round 3 was broken. That convergence failure is the abandon signal.

Spec and plan files at `docs/superpowers/specs/2026-05-26-background-indexing-v2-design.md` and `docs/superpowers/plans/2026-05-26-background-indexing-v2.md` are committed to this branch only; they will be lost with the branch. The next attempt should write its own spec and plan informed by both this and the PR #515 lessons.

PR #517, opened 2026-05-27
refactor(collection): make _ensure_initialized a fast probe (refs #513)

### LINKED ISSUE #513: Non-blocking startup: background-build FTS index and embeddings (supersedes #509)

pvliesdonk:
## Summary

`Collection._ensure_initialized()` no longer triggers `build_index()`. It now only flips `self._initialized = True`. The 22 existing call sites remain unchanged.

**Production behavior is unchanged today** because the lifespan in `_server_deps.py` still eagerly awaits `build_index()` (and `build_embeddings()`) before yielding, so the lazy fallback in `_ensure_initialized` was already dead. This PR is a **safety predicate** for the upcoming `BackgroundIndexer` landing (issue #513, attempt 4): without it, a tool call landing while a future background daemon is mid-`build_index()` would trigger a concurrent second `build_index()` race.

## Why ship this alone

The four-PR sequence for issue #513 (spec at `docs/superpowers/specs/2026-05-27-issue-513-background-indexer-design.md`):
1. **This PR** — `_ensure_initialized` fast probe.
2. `BackgroundIndexer` + lifespan switch (the actual handshake unblock).
3. `get_index_status` MCP tool.
4. Docs reframe; closes #513.

Each PR has one reviewer surface. The first three attempts at #513 (PRs #510, #515, #516) were abandoned because each tried to land 5–7 changes at once.

## Scope discipline

Not in this PR:
- Stripping `_ensure_initialized` from the 22 call sites (this was a major factor in PR #515 failing to converge).
- Any threading, `IndexStatus`, or lifespan changes.

Doc rot caught locally before push (pitfall #6 from PR #516's abandon notes): the `Collection` class docstring, two adjacent inline comments, the section banner, two `docs/design.md` paragraphs, and the test class name `TestLazyInitialisation` all still referenced the now-defunct lazy-build contract and have been updated in this PR.

## Test plan

- [x] New `test_ensure_initialized_does_not_build_index` monkeypatches `Collection.build_index` to raise — directly proves the invariant (not just that the method returned).
- [x] `TestExplicitInitialisation` (renamed from `TestLazyInitialisation`) — tightened assertions: `search()` and `list()` on an unbuilt index return empty, and a positive control test verifies `list()` returns documents after explicit `build_index()`.
- [x] Full suite green: 1595 passed, 1 skipped (pre-existing Python 3.13+ skip).
- [x] Lint, format, mypy clean.
- [x] Local preflight-circus clean (5 lenses + supplementary code-reviewer + comment-analyzer; 0 findings at ≥80 confidence).

Refs #513

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -365,10 +365,11 @@ Two methods manage the index:
   adds/modifies/deletes since the last scan and applies only the delta.
   Applies `exclude_patterns` filtering and purges stale excluded documents.
 
-**Lazy initialization**: on first call to `search()`, `list()`, or `read()`,
-`Collection` lazily builds the FTS index from `source_dir` if no pre-built
-`index_path` was provided. `build_index()` can be called explicitly to
-pre-warm the index or to force a rebuild.
+**Explicit initialization**: callers (the MCP lifespan, the CLI, tests)
+must call `build_index()` before querying. Tool methods on an unbuilt
+`Collection` return empty results; they do not trigger a filesystem
+scan. `build_index()` can also be called to force a rebuild
+(`force=True`).
 
 ### Error Handling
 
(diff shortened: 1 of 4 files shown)


pvliesdonk:
Closing in favor of a restart from main.

The seven commits on this branch accumulated through five rounds of preflight-circus iteration, each finding documentation imprecisions that I then attempted to fix with more documentation. Honest diagnosis: I was shipping claims I hadn't fully traced and relying on reviewers to catch what I missed, which is the failure mode — not a workflow. Combined with that, the PR's value rested on forward references to an unmerged PR2 with no anchored spec material, which made the PR unreviewable in isolation regardless of how carefully each individual docstring was worded.

A fresh plan with anchored forward references and pre-traced claims is being drafted. Same architectural target (Gemini's BackgroundIndexer remedy, issue #513).

No code from this PR will be reused.

PR #526, opened 2026-05-28
refactor(collection): remove lazy-init; per-bucket readiness policy

### LINKED ISSUE #525: Remove Collection lazy-init pattern (_ensure_initialized) — prerequisite for #513
## Background

Issue #513 (non-blocking MCP handshake via background indexing) has been attempted five times, all abandoned. Attempt 5 (2026-05-28) was abandoned before any code was written when an independent code-first audit found that the design's "Collection stays purely synchronous; lifespan kicks off `build_index()` on a background thread" approach cannot deliver the sub-second handshake while `Collection`'s lazy-init pattern exists.

This issue scopes the prerequisite refactor.

## Current state (verified against main, 2026-05-28)

`Collection` carries a `_initialized: bool` flag (`src/markdown_vault_mcp/collection.py:210`) set to `True` only inside `build_index()` at line 514. A method `_ensure_initialized()` (`collection.py:366-369`) gates lazy initialization:

```python
def _ensure_initialized(self) -> None:
    """Build the FTS index on first access if it has not been built yet."""
    if not self._initialized:
        self.build_index()
```

`_ensure_initialized()` is called at the top of **23 method bodies**:

- Read/search: `search` (420), `read` (450), `list` (477), `list_folders` (563), `list_tags` (577), `get_toc` (596), `get_backlinks` (613), `get_outlinks` (633), `get_broken_links` (646), `get_similar` (671), `get_recent` (690), `get_context` (721), `get_orphan_notes` (736), `get_most_linked` (749), `get_connection_path` (773), `stats` (905)
- Index management: `reindex` (524), `build_embeddings` (541)
- Write: `write_attachment` (1026), `write` (1061), `edit` (1102), `delete` (1132), `rename` (1168)

Because `_initialized` is an instance attribute (not persisted), it is `False` on every fresh server start — even when the FTS database file is already populated. The short-circuit inside `build_index()` at line 500 (`if not force and self._initialized: existing = self._fts.list_notes(); if existing: return ...`) therefore never fires on the first call after a process start, so warm-restart `build_index()` always re-scans and re-upserts every file.

## Why this blocks #513

#513's design moves `build_index()` and `build_embeddings()` off the lifespan critical path onto a background daemon thread. With the lazy-init pattern in place, every foreground tool call before the background worker completes hits `_ensure_initialized()` → `build_index()` on the calling thread. The handshake completes immediately but the *first* tool call blocks for the full scan duration — defeating the point of the move.

Adding a foreground wait on a background completion event is the cascade pattern that abandoned attempts 1 (PR #515) and 2 (PR #516); we will not re-attempt that.

## Scope

1. **Delete** `_initialized` attribute and `_ensure_initialized()` method from `Collection`.
2. **Delete** the 23 `self._ensure_initialized()` call sites listed above.
3. **Simplify** `build_index()` short-circuit at `collection.py:500` to drop the `self._initialized` half of the conjunction. (Net behavior: warm restarts get a fast `build_index()` by way of the `_fts.list_notes()` populated check.)
4. **Update** the `Collection` class docstring at `collection.py:86-101` to drop the "lazy initialisation handle it" framing and document the new contract: callers must call `build_index()` explicitly before searches/reads return non-empty results.
5. **Update** any tests that construct a `Collection` and immediately call read/search/list methods without first calling `build_index()`. (Grep will surface the set.)
6. **Update** `docs/design.md` indexing section to reflect the new contract.

## Out of scope

- Background indexing (#513) itself — this issue does not introduce a background thread.
- Provider-construction fail-fast (separate issue to be filed).
- Renaming `build_index()` or restructuring its short-circuit beyond removing the `_initialized` conjunct.

## Verification (per CLAUDE.md issue-writing discipline)

After this lands, the following shell checks must report clean:

```
grep -rn "_ensure_initialized" src/ tests/                  # expect: no matches
grep -rn "self\._initialized" src/                          # expect: no matches in collection.py
uv run pytest -x -q                                          # expect: all green
```

## Documentation

- `docs/design.md` — drop the "Lazy initialisation" framing; document that `build_index()` is a call-out on the caller side.
- `Collection` class docstring — update to reflect the new contract.
- README — no changes expected (README does not currently describe the lazy-init).

## Relationship to other issues

- Closes a hidden prerequisite for #513.
- Memory record: see `project_issue_513_attempt_5_abandon.md` in `~/.claude/projects/-mnt-code-markdown-mcp/memory/` for the abandon context.

pvliesdonk:
## Summary

Closes #525 (prerequisite for #513).

Drops `Collection._ensure_initialized()` lazy-bootstrap gate and replaces
it with an explicit per-bucket readiness contract derived from the
design discussion on #525:

- **Bucket 1** (`read`, `write`, `edit`, `delete`, `rename`,
  `write_attachment`) — never block. Disk is the source of truth.
- **Bucket 2** (`search`, `list`, `list_folders`, `list_tags`,
  `get_recent`, `get_orphan_notes`, `get_most_linked`,
  `get_broken_links`, `get_toc`, `stats`) — query current index state.
  Return empty on cold start, no implicit build.
- **Bucket 3** (`get_backlinks`, `get_outlinks`, `get_similar`,
  `get_context`, `get_connection_path`) — raise the new
  `IndexNotReadyError` on unbuilt indices, since the answers would
  be silently wrong.
- **Bucket 4** (`reindex`, `build_embeddings`) — raise
  `IndexNotReadyError`. `build_index()` short-circuit now keys on
  FTS state alone, so warm restarts (new process, same DB) are O(1).

The 23 `self._ensure_initialized()` callsites are removed (bucket 1+2)
or replaced with `self._require_index_ready()` (bucket 3+4). New
`Collection.wait_for_index_ready(timeout=None)` primitive raises
pre-#513; the `timeout` parameter is reserved for the
background-indexer variant (#513) where it will block on a completion
event. Internal flag renamed `_initialized` → `_index_built` and is
set only after a successful `build_index()` — a failed build leaves
the Collection not-ready.

## Contract change (callers, take note)

The warm-restart short-circuit is keyed on FTS contents alone and does
**not** detect config drift. To apply a configuration change
(`exclude_patterns`, `required_frontmatter`) to a pre-existing index,
call:

```python
collection.build_index(force=True)
collection.build_embeddings(force=True)  # if embeddings configured
```

Docstring and `docs/design.md` updated accordingly. Two pre-existing
tests (`test_build_index_purges_stale_excluded_docs[_with_embeddings]`)
updated to use `force=True` — they were exercising the lazy-init
re-scan side-effect that no longer fires.

## Verification (per #525)

```
$ grep -rn "_ensure_initialized" src/ tests/   # 0 matches
$ grep -rn "self\._initialized" src/           # 0 matches
$ uv run pytest -x -q                           # 1631 passed, 1 skipped
$ uv run ruff check . && uv run ruff format --check .  # clean
$ uv run mypy src/ tests/                       # 0 errors
$ uv run diff-cover coverage.xml --compare-branch=main --fail-under=80  # 100%
```

## TDD discipline

Failure modes per bucket enumerated up front (issue thread + commit
message). RED phase: 20 tests in `tests/test_indexing_readiness.py`
fail against unchanged code — 15 fail on bucket 2/3/4 + primitive +
cross-cutting contracts; 5 bucket-1 tests pass (regression checks for
the contract that disk ops don't need the index). GREEN phase
flipped all 20 with one set of edits. One missed failure mode
surfaced during full-suite verification (warm-restart purge-on-config-
change side-effect of the old `_initialized` flag) — surfaced as an
explicit contract change rather than patched away.

## Test plan
- [x] All existing tests pass (1631 / 1)
- [x] 20 new tests in `test_indexing_readiness.py` cover all four
      buckets + the readiness primitive + cross-cutting failed-build
      semantics
- [x] Patch coverage 100%
- [x] Verification grep clean (no `_ensure_initialized`, no
      `self._initialized` in src)

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -358,8 +358,15 @@ Two methods manage the index:
 
 - **`build_index(force=False)`**: initial population. Scans `source_dir` and
   builds the FTS index. Short-circuits as a no-op when the persisted FTS
-  database already contains documents (warm restart on the same
-  `index_path`). `force=True` drops and rebuilds from scratch. When a
+  database contains documents **and** carries the completeness sentinel
+  written by a prior clean build (warm restart on the same `index_path`).
+  A database with rows but no sentinel — residue of a process that
+  crashed mid-build, since `IndexManager.build_index` commits per-document
+  in its own transaction — is treated as cold and triggers a full
+  rebuild. The sentinel is the `build_completed_at` row in the FTS
+  `meta` table, cleared by `Collection.build_index` before any
+  destructive rebuild and written only after `_index_mgr.build_index`
+  returns cleanly. `force=True` drops and rebuilds from scratch. When a
   persistent `index_path` contains documents that now match
   `exclude_patterns`, they are purged from the FTS and vector indexes
   after the scan — but only when a scan actually runs (i.e. on a cold
--- a/docs/design.md
+++ b/docs/design.md
@@ -371,19 +371,21 @@ Two methods manage the index:
 
 **Readiness contract (issue #525)**: `Collection.__init__` does not
 populate the index. Callers must invoke `build_index()` explicitly
-before bucket-3 relational queries (`get_backlinks`, `get_outlinks`,
-`get_similar`, `get_context`, `get_connection_path`) or the bucket-4
-coordinators (`reindex`, `build_embeddings`); otherwise
-`IndexNotReadyError` is raised. Bucket-1 file operations (`read`,
-`write`, `edit`, `delete`, `rename`, `write_attachment`) and bucket-2
-aggregate queries (`search`, `list`, `stats`, `list_folders`,
-`list_tags`, `get_recent`, `get_orphan_notes`, `get_most_linked`,
-`get_broken_links`, `get_toc`) work on an unbuilt index — bucket-1
-hits disk directly; bucket-2 queries whatever is currently in the
-index (empty on cold start). `wait_for_index_ready(timeout=None)` is
-the readiness primitive: it raises `IndexNotReadyError` pre-#513;
-once the background indexer (#513) lands it will block on a
-completion event.
+before bucket-3 relational/FTS-backed queries (`get_backlinks`,
+`get_outlinks`, `get_similar`, `get_context`, `get_connection_path`,
+`get_toc`) or the bucket-4 coordinators (`reindex`,
+`build_embeddings`); otherwise `IndexNotReadyError` is raised.
+`start()` must also be called after `build_index()` because its git
+pull loop wires `reindex` as the `on_pull` callback. Bucket-1 file
+operations (`read`, `write`, `edit`, `delete`, `rename`,
+`write_attachment`) and bucket-2 aggregate queries (`search`, `list`,
+`stats`, `list_folders`, `list_tags`, `get_recent`,
+`get_orphan_notes`, `get_most_linked`, `get_broken_links`) work on an
+unbuilt index — bucket-1 hits disk directly; bucket-2 queries
+whatever is currently in the index (empty on cold start).
+`wait_for_index_ready(timeout=None)` is the readiness primitive: it
+raises `IndexNotReadyError` pre-#513; once the background indexer
+(#513) lands it will block on a completion event.
 
 To apply a configuration change (e.g. new `exclude_patterns`,
 `required_frontmatter`) to a pre-existing index, call
--- a/docs/design.md
+++ b/docs/design.md
@@ -357,18 +357,39 @@ its source document path, enabling bulk deletion when a document is reindexed.
 Two methods manage the index:
 
 - **`build_index(force=False)`**: initial population. Scans `source_dir` and
-  builds the FTS index. If the index already has data and `force=False`, this
-  is a no-op. `force=True` drops and rebuilds from scratch. When a persistent
-  `index_path` contains documents that now match `exclude_patterns`, they are
-  purged from the FTS and vector indexes after the scan.
+  builds the FTS index. Short-circuits as a no-op when the persisted FTS
+  database already contains documents (warm restart on the same
+  `index_path`). `force=True` drops and rebuilds from scratch. When a
+  persistent `index_path` contains documents that now match
+  `exclude_patterns`, they are purged from the FTS and vector indexes
+  after the scan — but only when a scan actually runs (i.e. on a cold
+  index or with `force=True`); a warm-restart short-circuit does not
+  apply config changes.
 - **`reindex()`**: incremental update. Uses `ChangeTracker` to detect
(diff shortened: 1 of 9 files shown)


pvliesdonk:
Review findings addressed in 79ddc5e:

- **#1 (get_toc bucket misclassification)**: `get_toc` is FTS-backed (queries the `documents` row); cold-start returns no row, so the underlying manager raised a misleading `ValueError("Document not found")`. Moved to bucket 3 — raises `IndexNotReadyError` first. Regression test added.
- **#2 (failed force=True leaves stale flag)**: `build_index()` now resets `_index_built=False` before delegating to `_index_mgr.build_index()`, so a mid-rebuild exception leaves the Collection visibly not-ready. Regression test added.
- **#3 (docstrings missing IndexNotReadyError)**: Added to `Raises` sections of all bucket-3 methods (get_toc, get_backlinks, get_outlinks, get_similar, get_context, get_connection_path) and bucket-4 coordinators (reindex, build_embeddings).
- **#4 (start() ordering)**: Documented on `start()` docstring and the class-level contract. The implicit requirement is now explicit: `build_index()` must precede `start()` because the git pull loop wires `reindex` as `on_pull`.

Test count 1631 → 1633, all green; lint/mypy/patch-coverage (100%) clean.

pvliesdonk:
Round 2 findings addressed in 8fd6ad5, but framed as the two root causes rather than two patches:

**Root cause A — incomplete external-callsite audit.** Lazy-init was a safety net for every caller; I only grepped inside `Collection`. The full grep across `src/` finds one remaining broken external caller (`_cmd_reindex`) and three safe ones (server lifespan, `_pause_and_reindex`, the `start()` `on_pull` wiring documented in round 1). Fix: `_cmd_reindex` calls `build_index()` first (O(1) on the warm path). Real-Collection integration test added — the existing mocked tests can't catch this regression class.

**Root cause B — `_index_built` flag conflated two propositions.** Reader semantics: "FTS is the coherent representation of source_dir." Writer in the warm-restart short-circuit set it based on "FTS has rows," which proves only "some rows exist" and not "prior build completed cleanly" (per-doc upserts each commit separately, so a crash mid-loop persists partial state). Fix: persisted completeness sentinel — new `meta` table, written by `Collection.build_index()` only after `_index_mgr.build_index()` returns cleanly, cleared before any destructive rebuild. Warm-restart short-circuit now requires both row presence AND the sentinel. Reindex / write / edit / delete don't touch it because they preserve coherence by construction.

Discipline note: I saved a memory file (`feedback_failure_mode_enumeration_completeness`) recording that I need to enumerate "what behaviors does the removed code give callers" and "what failure paths must each new invariant survive" before writing code — not discover them through review iterations.

1633 → 1636 tests, lint/mypy/100% patch coverage clean.

PR #527, opened 2026-05-28
feat(server): MARKDOWN_VAULT_MCP_DISABLE_APPS_UI env var to hide MCP-Apps tools

Finomosec:
## Summary

Adds an `apps-ui` tag to the `browse_vault` and `show_context` tools and a tag-based disable pass in `make_server` that fires when `MARKDOWN_VAULT_MCP_DISABLE_APPS_UI` is truthy (`1`/`true`/`yes`/`on`).

Composes with the existing `write` and `git-managed` disable passes — set-union on disabled tags. No behavior change when the env var is unset.

## Motivation

Both tools are designed to open an interactive MCP Apps panel for the user. Their docstrings already tell the LLM not to call them when no UI is rendered (preferring `search` / `read` / `list_documents` / `get_context` instead) — but they still appear in `tools/list`, costing tokens in every session.

Some clients (e.g. a CLI-only consumer) cannot render MCP Apps panels at all. For those, hiding the two tools entirely is cleaner than keeping them and relying on the docstring guidance.

## Changes

- `_server_apps.py`: add `tags={"apps-ui"}` to `browse_vault` and `show_context` decorators.
- `server.py`: read `MARKDOWN_VAULT_MCP_DISABLE_APPS_UI`; call `mcp.disable(tags={"apps-ui"})` when truthy. Documented inline.

## Test plan

- [x] Container build + run with the env var **unset** → both tools listed.
- [x] Container run with `MARKDOWN_VAULT_MCP_DISABLE_APPS_UI=true` → `tools/list` returns neither.
- [x] Runtime check inside the container confirms 27 tools visible vs. 29 without the var (delta = `browse_vault` + `show_context`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,6 +212,7 @@ All configuration is via environment variables with the `MARKDOWN_VAULT_MCP_` pr
 | `MARKDOWN_VAULT_MCP_HTTP_PATH` | `/mcp` | HTTP endpoint path for streamable HTTP transport (used by `serve --transport http`) |
 | `MARKDOWN_VAULT_MCP_EVENT_STORE_URL` | `file:///data/state/events` | Event store backend for HTTP session persistence. `file:///path` (default) survives restarts; `memory://` for dev (lost on restart). |
 | `MARKDOWN_VAULT_MCP_APP_DOMAIN` | (auto) | Override the Claude app domain used for MCP Apps iframe sandboxing. Auto-computed from `BASE_URL` when not set. |
+| `MARKDOWN_VAULT_MCP_DISABLE_APPS_UI` | `false` | Hide MCP-Apps UI tools (`browse_vault`, `show_context`) from the tool listing for clients that don't render MCP Apps panels. |
 | `FASTMCP_LOG_LEVEL` | `INFO` | Log level for FastMCP internals (`DEBUG`, `INFO`, `WARNING`, `ERROR`). App loggers default to `INFO`. `-v` overrides both to `DEBUG`. |
 | `FASTMCP_ENABLE_RICH_LOGGING` | `true` | Set to `false` for plain/structured JSON log output instead of Rich-formatted output. |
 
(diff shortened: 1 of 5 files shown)


pvliesdonk:
Thanks for the contribution — the feature itself is valid and the `tags={"apps-ui"}` additions in `_server_apps.py` are exactly right.

The env-var reading in `server.py` bypasses the project's established patterns, and those patterns already live in `fastmcp-pvl-core`. The utilities you need are already imported in `config.py`:

```python
from fastmcp_pvl_core import env as _core_env      # via _env()
from fastmcp_pvl_core import parse_bool as _parse_bool
```

**What needs to change:**

**1. Add the field to `CollectionConfig` and `load_config()` in `config.py`** (same pattern as `read_only`):

```python
# in CollectionConfig dataclass
disable_apps_ui: bool = False

# in load_config()
raw_disable_apps_ui = _env("DISABLE_APPS_UI")
disable_apps_ui = _parse_bool(raw_disable_apps_ui) if raw_disable_apps_ui is not None else False
```

**2. Use `config.disable_apps_ui` in `server.py`** — no `import os`, no inline truthy set:

```python
if config.disable_apps_ui:
    mcp.disable(tags={"apps-ui"})
```

This puts the env-var reading in the one place it belongs (config.py), uses the pvl-core `parse_bool` that already handles `"1"/"true"/"yes"/"on"`, and keeps `server.py` as a pure wiring layer (matching how `read_only` and `git_repo_url` already work).

**3. Document the new env var** in `docs/configuration.md` and `README.md` alongside the other `MARKDOWN_VAULT_MCP_*` vars. The project's documentation discipline requires this before merge (see CLAUDE.md).

The `tags={"apps-ui"}` lines in `_server_apps.py` need no changes.

pvliesdonk:
Note that I'm happy to take it from here, but I have no idea yet how quick that will be.

Finomosec:
I made the suggested changes. Hope this is cool now.

Finomosec:
Redid the PR as #702 ontop of current master.

PR #528, opened 2026-05-28
feat(server): cold-start background FTS build (#513 PR1)

pvliesdonk:
## Summary

PR1 of three planned PRs closing #513. On cold start (no completeness sentinel from PR #526), the MCP `initialize` handshake completes sub-second; the FTS build runs in a background daemon thread. Bucket-1 and bucket-2 callers work immediately against the (empty) index. Bucket-3 and bucket-4 callers block on the completion event via the upgraded `wait_for_index_ready(timeout=None)` primitive. Warm restarts keep PR #526's O(1) short-circuit and never spawn the background thread.

## What changed

- **Lifespan rewire** (`_server_deps.py`): three paths via the new `Collection.should_use_background_build()` predicate — warm on-disk DB → synchronous short-circuit; cold on-disk DB → `start_background_build_index()`; in-memory DB → synchronous (no persistent state to short-circuit from).
- **`Collection` additions:**
  - State: `_background_build_thread`, `_background_build_done` (pre-set), `_background_build_error`, `_background_started`.
  - Methods: `is_index_ready()`, `start_background_build_index()`, `get_index_status()`, `should_use_background_build()`.
  - `wait_for_index_ready(timeout=None)` is now event-based — waits on the completion event, raises `IndexBuildFailedError` if the background build captured an exception, raises `IndexNotReadyError` on timeout or never-scheduled.
  - `_require_index_ready()` (PR #525) delegates to `wait_for_index_ready(timeout=None)` — bucket-3/4 callsites naturally inherit blocking semantics; no callsite changes.
  - `close()` joins the background thread with a 30s bounded timeout.
- **New `IndexBuildFailedError`** in `exceptions.py`, parallel to `IndexNotReadyError`. Carries the original via `__cause__`.
- **New MCP tool `get_index_status`** in `_server_tools.py` — non-blocking `{status, documents_indexed, error}` for operator polling.
- **Docs** — `docs/design.md` and the `Collection` class docstring document the cold-start contract; bucket-3/4 method `Raises` sections gain `IndexBuildFailedError`.

## What's intentionally NOT in this PR

- **PR2:** embeddings background build. On cold start with embeddings configured, the foreground `build_embeddings()` is a no-op against the empty FTS — semantic search returns empty until the operator runs CLI `markdown-vault-mcp index` or PR2 lands.
- **PR3:** warm-start drift reindex (catch disk changes that happened while the server was down).
[...]

PR #529, opened 2026-05-28
feat(server): cold-start background FTS via tool-layer wait (#513 PR1, attempt 7)

pvliesdonk:
## Summary

PR1 of three planned PRs closing #513. Supersedes the abandoned PR #528 (attempt 6) which collapsed under the uniform "library delegates to wait" abstraction.

**Corrected boundary:** library raises (`_require_index_ready` stays as PR #525 left it — raise-immediately, no delegation). Blocking semantics live at the MCP tool/resource layer in a `@needs_index_ready` decorator applied to 7 tools + 2 resources. Internal callers (lifespan, git pull loop, CLI, direct library users) handle "not ready" with caller-appropriate logic — never block.

## What changed

- **Lifespan** routes cold on-disk through `start_background_build_index()`; warm on-disk and in-memory through synchronous `build_index()`. Embeddings gated on `is_index_ready()` — skipped + logged on cold start (PR2 backgrounds them).
- **`Collection` additions:** state (4 fields), `is_index_ready()`, `start_background_build_index()`, `should_use_background_build()`, `get_index_status()`. `wait_for_index_ready(timeout)` event-based with 4-step control flow including the never-scheduled guard.
- **`close()`** joins the background thread under `_write_lock`, before any resource teardown.
- **New `IndexBuildFailedError`** in `exceptions.py`. Surfaces at the decorator (not at the library).
- **`needs_index_ready` decorator** in new `_server_readiness.py`. Applied via two-commit split — Commit A applied to `get_backlinks` only + preflight test; Commit B applied to the other 8 surfaces.
- **`get_index_status` MCP tool** for operator polling without making a real bucket-3/4 call.
- **Env var `MARKDOWN_VAULT_MCP_READY_TIMEOUT_S`** (default 60s) configures the decorator's wait.

## Out of scope (PR2 / PR3)

- Embeddings background build → PR2.
- Warm-start drift reindex → PR3.
- Cooperative cancellation inside `IndexManager.build_index` (close relies on daemon + bounded join).

## Failure-mode coverage

Every row in the design's failure-mode table maps to a test in `tests/test_background_fts.py`. The canonical regression test against attempt 6's hole is `test_require_index_ready_raises_immediately_not_blocks` — asserts a bucket-3/4 *library* call during in-flight background raises within 0.1s wall-clock (not blocks).

## Discipline

This attempt was preceded by:
- Closing PR #528 unmerged with a retrospective.
- Memory entries: `feedback_uniform_delegation_design_smell.md` (when "every X behaves the same way" requires per-caller exceptions, the contract is wrong, not the callers) and `project_issue_513_attempt_6_abandon.md`.
- A fresh brainstorm with the corrected boundary as the fixed anchor.
- A spec that went through 4 independent review passes before any code was written.
- Subagent-driven implementation with two-stage review per task and a structurally-enforced commit split (Commit A: decorator preflight on `get_backlinks` only; Commit B: rollout to remaining 8 surfaces) so the decorator pattern was verified end-to-end before broad application.

## Test plan
- [x] `pytest -q` — 1674 passed, 1 skipped
- [x] `ruff check` + `ruff format --check` — clean
- [x] `mypy src/ tests/` — clean
- [x] `diff-cover --fail-under=80` — 95% patch coverage
- [x] Verification grep — `_require_index_ready` stays raise-only on main + this branch (no delegation re-introduced)

Refs: #513 (PR2 + PR3 to follow).

### DIFF
--- a/README.md
+++ b/README.md
@@ -380,6 +380,7 @@ markdown-vault-mcp reindex [--source-dir PATH] [--index-path PATH]
 | `stats` | Get collection statistics (document count, chunk count, link health metrics, etc.) |
 | `build_embeddings` | Build or rebuild vector embeddings for semantic search |
 | `embeddings_status` | Check embedding provider and index status |
+| `get_index_status` | Check background FTS build state (`ready` / `building` / `failed`) |
 | `get_backlinks` | Find all documents that link to a given document |
 | `get_outlinks` | Find all links from a document, with existence check |
 | `get_broken_links` | Find all links pointing to non-existent documents |
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -17,6 +17,19 @@ All configuration is via environment variables. Most use the `MARKDOWN_VAULT_MCP
 | `MARKDOWN_VAULT_MCP_TEMPLATES_FOLDER` | string | `_templates` | No | Relative folder path used by the `create_from_template` prompt to discover/read template files |
 | `MARKDOWN_VAULT_MCP_PROMPTS_FOLDER` | path | — | No | Path to a directory of `.md` prompt files that extend or override built-in prompts |
 
+## Index Readiness
+
+### `MARKDOWN_VAULT_MCP_READY_TIMEOUT_S`
+
+Default: `60` (seconds).
+
+Maximum time the MCP-layer `needs_index_ready` decorator waits for
+the FTS index to become ready before raising `IndexNotReadyError`
+to the client. Applied to bucket-3/4 tool and resource calls during
+a cold-start background FTS build. Increase for very large vaults
+where the initial scan takes longer; decrease for tighter feedback
+on stuck builds.
+
 <!-- DOMAIN-CONFIG-VARS-START -->
 ## Server Identity
 
--- a/docs/design.md
+++ b/docs/design.md
@@ -394,6 +394,31 @@ whatever is currently in the index (empty on cold start).
 raises `IndexNotReadyError` pre-#513; once the background indexer
 (#513) lands it will block on a completion event.
 
+**Cold-start background FTS (issue #513 PR1, tool-layer wait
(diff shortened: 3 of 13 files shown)


pvliesdonk:
Review findings addressed in bf10a05:

- **Decorator docstring direction (claude-review)**: `_server_readiness.py`'s docstring said "place `@needs_index_ready(...)` ABOVE `@mcp.tool(...)`" but the actual code (correctly) has `@mcp.tool` outer / `@needs_index_ready` inner — Python applies bottom-up, so the wrap-then-register order matters. Code was right, docstring lied. Changed to "BELOW."
- **`build_index()` recovery path (gemini-code-assist)**: `Collection.build_index()` set `_index_built=True` on success but didn't clear `_background_build_error`. After a failed background build, an in-process recovery via synchronous `build_index()` would leave `is_index_ready()` stuck at False and bucket-3/4 calls would keep surfacing `IndexBuildFailedError`. Now clears the captured error after a clean sync build. Regression test: `test_synchronous_build_index_clears_prior_background_error`.

Minor items deferred (acknowledged trade-offs, not behavior bugs):
- `needs_index_ready` wrapper reads `collection` from kwargs only. FastMCP's `Depends()` always injects as kwarg (verified by the working preflight); positional call hits a diagnostic `RuntimeError`. Defer.
- Wrapper hardcoded `async def` with no `iscoroutinefunction` guard. All current handlers are async; defer.
- `get_index_status` `"building"` covers both "in-flight" and "never scheduled" — acknowledged in spec + library docstring as intentional from operator perspective.

The discipline gap surfaced by Fix 2 is the same pattern as before: my failure-mode enumeration covered the start_background_build_index → wait_for_index_ready → bucket-3/4 path but didn't cover the synchronous recovery path that joins back to the same error state. Saving the lesson in memory for the next PR2 brainstorm.

1674 → 1675 tests, lint/mypy clean.

PR #531, opened 2026-05-29
feat(server): embeddings background build (#513 PR2)

pvliesdonk:
## Summary

PR2 of three planned PRs closing #513. Builds on PR1 (PR #529 merged 2026-05-28). Same boundary as PR1: library raises immediately; MCP-layer decorator waits.

The existing background daemon thread now runs two phases — FTS then embeddings — when an `embedding_provider` is configured. The decorator gains an `embeddings=True` argument; `get_similar`, `vault_similar`, and `reindex` switch to it. `get_index_status` extends to a nested shape with `fts` and `embeddings` substates.

## What changed

- **`Collection`:** new state (`_embeddings_build_done` event, `_embeddings_build_error`, `_embeddings_built`); new methods `is_embeddings_ready()` and `wait_for_embeddings_ready(timeout)`; `build_embeddings()` facade gains the recovery-path state updates (mirrors PR1's R2 fix for `build_index`); `start_background_build_index()` worker extends to phase 2 with the FTS-fail-unblocks-embeddings-waiter critical fix and the thread.start-fail-unblocks-both-events parallel fix; new `has_embedding_provider` property (uniform refinement discovered during implementation — see below).
- **`@needs_index_ready(embeddings=False)`:** new parameter; when True, waits for both phases. The embeddings wait is gated on `collection.has_embedding_provider` at the decorator level (uniform refinement, not per-caller carve-out — closes the I14 vector race for provider-configured deployments, no-op for no-provider). Worst-case wait `2 × MARKDOWN_VAULT_MCP_READY_TIMEOUT_S` (default 120s).
- **Tools/resources switched:** `get_similar` (semantic), `vault_similar` (semantic resource), `reindex` (mutates vector sidecar; prevents corruption race).
- **`get_index_status`:** nested shape `{status, fts: {...}, embeddings: {...}}`; PR1's flat shape is replaced. Backward-incompatible but PR1 only shipped four days ago.
- **Lifespan:** cold start with provider configured logs the embeddings-in-background notice; the synchronous warm/in-memory path keeps embeddings inline as before.

## Implementation refinement worth noting

The spec assumed every deployment uses `embeddings=True` decorators would have an `embedding_provider` configured. In practice the majority of deployments don't configure one — without a guard, switching `reindex` to `embeddings=True` would silently regress (it would raise `IndexNotReadyError(\"never scheduled\")` for the no-provider case). The fix is a uniform Collection-level guard: `has_embedding_provider` is checked at the decorator level. It applies identically to all three decorated handlers (not a per-caller carve-out). When a provider is configured, the I14 race is closed; when not, the wait is a no-op. Reviewer-accepted.

## Out of scope

- PR3: warm-start drift reindex.
- Cooperative cancellation inside `IndexManager.build_embeddings` (close keeps PR1's daemon + 30s join).
- Re-entrancy guard for concurrent `build_embeddings` MCP calls (idempotent via `force=False` short-circuit; documented limitation).
- Internal `reindex` callsites (git pull loop, `_reindex_after_pull`) racing with background embeddings on the vector sidecar — accepted as documented limitation (operator recovery via CLI \`index\`).
- `search` (semantic/hybrid mode) stays bucket-2; returns empty during embeddings-building window.

## Discipline

Brainstorming + 4 spec review rounds before code (R1 RED → 2 criticals; R2 RED → 2 new findings including thread.start parallel hang; R3 YELLOW → over-engineered exception list; R4 GREEN). The CRITICAL regression tests — \`test_fts_failure_skips_embeddings_phase_with_provider\` and \`test_thread_start_failure_with_provider_releases_embeddings_waiter\` — lock in the two hang-bug fixes by asserting wall-clock < 0.05s.

## Test plan

- [x] \`pytest -q\` clean (1611 → 1705)
- [x] \`ruff check/format\` clean
- [x] \`mypy\` clean
- [x] \`diff-cover --fail-under=80\` (92% patch)
- [x] Boundary preservation: \`_require_index_ready\` unchanged on this branch
- [x] Exactly 3 surfaces decorated with \`embeddings=True\` (\`get_similar\`, \`reindex\`, \`vault_similar\`)

Refs: #513 (PR3 to follow).

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -423,11 +423,16 @@ without deadlocking on internal blocking.
 thread that handles FTS extends to a second phase when an
 ``embedding_provider`` is configured. After a successful FTS phase,
 the worker calls ``Collection.build_embeddings()`` to populate the
-vector sidecar. Bucket-3 surfaces that need vectors —
-``get_similar``, ``vault_similar``, and ``reindex`` (which directly
-mutates the vector sidecar via ``IndexManager.reindex``) — use the
-new ``@needs_index_ready(embeddings=True)`` decorator argument and
-wait for both phases at the MCP layer. The library never blocks on
+vector sidecar. Bucket-3/4 surfaces that need vectors —
+``get_similar``, ``vault_similar``, ``reindex`` (which directly
+mutates the vector sidecar via ``IndexManager.reindex``), and
+``build_embeddings`` (which calls ``IndexManager.build_embeddings``
+directly — the same code the worker's phase-2 runs) — use the new
+``@needs_index_ready(embeddings=True)`` decorator argument and wait
+for both phases at the MCP layer. The mutator surfaces require the
+embeddings wait to serialise against the still-running worker
+phase-2 (``IndexManager.build_embeddings`` is lock-free across its
+``_load_vectors``/``vectors.save`` critical region). The library never blocks on
 embeddings; ``_require_index_ready()`` is unchanged from PR1.
 ``get_index_status`` returns a nested shape with ``fts`` and
 ``embeddings`` substates plus a top-level reduction.
--- a/docs/design.md
+++ b/docs/design.md
@@ -419,6 +419,33 @@ The library's `_require_index_ready()` is unchanged from PR #525
 pull loop and lifespan's embeddings path handle "not ready"
 without deadlocking on internal blocking.
 
+**Embeddings background build (issue #513 PR2).** The same daemon
+thread that handles FTS extends to a second phase when an
+``embedding_provider`` is configured. After a successful FTS phase,
+the worker calls ``Collection.build_embeddings()`` to populate the
+vector sidecar. Bucket-3 surfaces that need vectors —
+``get_similar``, ``vault_similar``, and ``reindex`` (which directly
+mutates the vector sidecar via ``IndexManager.reindex``) — use the
+new ``@needs_index_ready(embeddings=True)`` decorator argument and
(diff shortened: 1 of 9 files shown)


pvliesdonk:
Abandoning PR2 attempt.

Six rounds of bot review found legitimate bugs at each round, but the trajectory exposed an underlying design drift rather than a polishable surface:

- R1–R3 found real bugs and were cleanly addressed.
- R4 surfaced doc-rot I introduced in the R3 commit.
- R5 surfaced a write-write race on the vector sidecar; the proposed fix (`@needs_index_ready(embeddings=True)` on `build_embeddings`) was applied.
- R6 then showed that very fix permanently blocks `build_embeddings(force=True)` recovery after a background phase-2 failure — i.e., the R5 fix and the R2 facade recovery path are in conflict.

The root cause is that `wait_for_embeddings_ready` is overloaded as both "wait until the writer is done" and "raise if the build errored", and different surfaces need those two semantics gated differently (`build_embeddings` is a recovery surface; `get_similar` / `vault_similar` are not). Patching the contract round-by-round produces drift; the right move is to revisit the design with that split in mind from the start.

PR2 is closed without merge. A fresh attempt should:

1. Split the "writer-busy" predicate from the "errored" predicate at the library level.
2. Decide per-surface which gating each MCP tool/resource opts into.
3. Re-validate the race fix and the recovery path can coexist in the new contract.

Issue #513 stays open. PR1 (#529) remains merged and operational.

PR #542, opened 2026-05-30
feat!(server): rename status ready->queryable + remove 'ready' vocabulary (#538)

### LINKED ISSUE #538: Rename get_index_status status "ready" → "queryable" + priority flip
## Background

This issue is the first of a four-issue split that replaces #533. Order chosen so the rename lands first, in isolation, forcing precision about what the status word means before any unwind work touches it. See #533's closing comment for the discipline retrospective.

## Scope

Single, focused change to ``Collection.get_index_status`` + its MCP wire shape + tests + docs.

- Rename status value ``"ready"`` to ``"queryable"`` everywhere it appears as a wire value: ``collection.py`` get_index_status body, the MCP tool docstring in ``_server_tools.py``, the lifespan tests at ``tests/test_background_fts.py`` that assert on ``status``, the README quick-reference table, ``docs/tools/index.md``, ``docs/resources.md`` if applicable.
- Flip the priority order in get_index_status so ``is_index_ready()`` is checked first (a readable index reports ``"queryable"`` even if some captured background error is present from a prior attempt). ``"failed"`` only when NOT queryable AND event is set AND a captured error exists. ``"building"`` is the catch-all.
- Document the status word semantics in the MCP tool docstring: ``"queryable"`` is the in-process snapshot that the FTS preconditions are met; runtime corruption is discovered at use time and is NOT this field's job.
- Update CHANGELOG with a BREAKING entry for the wire-value rename.

## Out of scope (deferred to sibling issues)

- ``IndexBuildFailedError`` deletion (→ #539)
- Captured-error step in ``wait_for_index_ready`` (→ #539)
- ``is_index_ready()`` simplification (→ #539)
- ``IndexNotReadyError.reason`` discriminator (→ #540)
- ``sqlite3.OperationalError`` catch in decorator (→ #541)
- Drift detection / status subdivision (→ #534)

## Verification

```bash
grep -nE 'status\\s*=\\s*"ready"' src/markdown_vault_mcp/collection.py
# expect: 0 matches

grep -rn "\\"ready\\"" docs/tools/index.md docs/resources.md README.md | grep -i status
# expect: 0 matches in the status enums

uv run pytest -x -q
# expect: green; lifespan tests asserting status="queryable"
```

## Documentation

- README.md tools table
- docs/tools/index.md get_index_status entry + table header
- docs/resources.md if it references the status enum
- CHANGELOG.md BREAKING entry
- Collection class docstring forward-pointer to #534's expected subdivision (``"up_to_date"`` / ``"degraded"``)

## Conventional commit

``feat!:`` (status enum value change on the wire)

## Relationship

Closes — part of #533's replacement set. Sibling: #539, #540, #541. Downstream consumers (#534, #535, #536) update their "depends on #533" pointers to reference this issue's PR plus its siblings.

pvliesdonk:
## Summary

Closes #538. First of four issues replacing the abandoned #533.

Eliminates the loaded word "ready" from the FTS-index readiness
vocabulary at every load-bearing identifier and prose surface.
Replaces with target words chosen for what they actually describe:

- Status wire value `"ready"` → `"queryable"`
- Method `is_index_ready()` → `is_queryable()`
- Method `_require_index_ready()` → `_require_built()` (it only
  checks `_index_built`)
- Method `wait_for_index_ready()` → `wait_until_queryable()`
- Decorator `needs_index_ready` → `needs_queryable`
- Module `_server_readiness.py` → `_server_queryable.py`
- Exception `IndexNotReadyError` → `IndexUnavailableError`
- Env var `MARKDOWN_VAULT_MCP_READY_TIMEOUT_S` →
  `MARKDOWN_VAULT_MCP_BUILD_TIMEOUT_S`

Plus the priority flip in `get_index_status`: a built index with a
captured background error from a prior attempt now reports
`status="queryable"` (with the diagnostic in `error`), not
`status="failed"`.

Design spec: `docs/superpowers/specs/2026-05-30-issue-538-status-rename-design.md` (gitignored, local-only).

## Breaking changes

See CHANGELOG. External consumers must rename references; no
deprecation shims ship (CLAUDE.md "don't build backward-compat
layers for internal refactors"). Running deployments that set the
old env var will silently fall back to the 60-second default;
update operator configs.

## Test plan

- [x] New `test_get_index_status_queryable` (TDD-driven rename of
      `test_get_index_status_ready`).
- [x] New `test_get_index_status_queryable_when_built_with_captured_error`
      (TDD-driven priority-flip test).
- [x] All existing test assertions sweep from `"ready"` to
      `"queryable"`.
- [x] Comprehensive grep verifies no surviving "ready" identifiers
      outside CHANGELOG.
- [x] CI hard gates (pytest, ruff, mypy, pre-commit) all green.
- [x] Patch coverage ≥80% on changed files.

## Follow-ups in chain

This is PR 1 of 4 replacing #533:
- #538 (this PR) — rename + priority flip.
- #539 — delete `IndexBuildFailedError`; simplify
  `wait_until_queryable`; simplify `is_queryable()` to the
  two-field check.
- #540 — add `IndexUnavailableReason` Literal + required `reason`
  field on `IndexUnavailableError`.
- #541 — MCP decorator catches `sqlite3.OperationalError` and adds
  the third `reason` value (`"broken"`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,3 +2,37 @@
 
 All notable changes to this project will be documented in this file.
 Format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
+
+## [Unreleased]
+
+### Changed (BREAKING)
+
+- `Collection.get_index_status` `status` field value renamed from
+  `"ready"` to `"queryable"`. MCP clients pattern-matching on the
+  old value will silently treat the new value as unknown until
+  updated.
+- `Collection.get_index_status` priority order flipped: a built
+  index with a captured background error from a prior attempt now
+  reports `status="queryable"` (with the diagnostic in `error`),
+  not `status="failed"`. `"failed"` now means "preconditions do not
+  hold AND a captured error exists" — i.e., the index is not
+  readable.
+- `Collection.is_index_ready()` renamed to `Collection.is_queryable()`.
+- `Collection._require_index_ready()` (private) renamed to
+  `Collection._require_built()` — matches what it actually checks
+  (only `_index_built`, single field).
+- `Collection.wait_for_index_ready()` renamed to
+  `Collection.wait_until_queryable()`.
+- MCP decorator `needs_index_ready` renamed to `needs_queryable`.
+  Module `_server_readiness.py` renamed to `_server_queryable.py`.
+- Public exception `IndexNotReadyError` renamed to
+  `IndexUnavailableError`.
+- Environment variable `MARKDOWN_VAULT_MCP_READY_TIMEOUT_S` renamed
+  to `MARKDOWN_VAULT_MCP_BUILD_TIMEOUT_S`. Running deployments that
+  set the old variable will silently fall back to the 60-second
+  default after upgrading; update operator configs (compose files,
+  systemd units, `.env` files) accordingly.
+
+External consumers that previously imported `IndexNotReadyError`,
+called `is_index_ready()`/`wait_for_index_ready()`, or set the old
+env var must rename their references. No deprecation shims ship.
--- a/README.md
+++ b/README.md
@@ -380,7 +380,7 @@ markdown-vault-mcp reindex [--source-dir PATH] [--index-path PATH]
 | `stats` | Get collection statistics (document count, chunk count, link health metrics, etc.) |
 | `build_embeddings` | Build or rebuild vector embeddings for semantic search |
 | `embeddings_status` | Check embedding provider and index status |
-| `get_index_status` | Check background FTS build state (`ready` / `building` / `failed`) |
+| `get_index_status` | Check background FTS build state (`queryable` / `building` / `failed`) |
 | `get_backlinks` | Find all documents that link to a given document |
 | `get_outlinks` | Find all links from a document, with existence check |
 | `get_broken_links` | Find all links pointing to non-existent documents |
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -17,14 +17,14 @@ All configuration is via environment variables. Most use the `MARKDOWN_VAULT_MCP
 | `MARKDOWN_VAULT_MCP_TEMPLATES_FOLDER` | string | `_templates` | No | Relative folder path used by the `create_from_template` prompt to discover/read template files |
 | `MARKDOWN_VAULT_MCP_PROMPTS_FOLDER` | path | — | No | Path to a directory of `.md` prompt files that extend or override built-in prompts |
 
-## Index Readiness
+## Index Build Timeout
 
-### `MARKDOWN_VAULT_MCP_READY_TIMEOUT_S`
+### `MARKDOWN_VAULT_MCP_BUILD_TIMEOUT_S`
 
 Default: `60` (seconds).
 
-Maximum time the MCP-layer `needs_index_ready` decorator waits for
-the FTS index to become ready before raising `IndexNotReadyError`
+Maximum time the MCP-layer `needs_queryable` decorator waits for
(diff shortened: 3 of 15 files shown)


pvliesdonk:
Thanks for the review. Responding to each finding:

**Finding 1 (Notable: `get_index_status` / `@needs_queryable` divergence)** — deliberately not adding a docstring/CHANGELOG note. Tracing the state machine, the `_index_built=True AND _background_build_error is not None` state isn't actually reachable via production paths:

- Initial `build_index()` success clears the error.
- Background worker's `_worker` calls `build_index()` which resets `_index_built=False` at start (collection.py line 802); if scan raises, `_index_built` stays False → status reports `"failed"`, not `"queryable"`.
- Foreground `build_index(force=True)` same reset+raise pattern.
- `start_background_build_index`'s reset block clears both flags before spawning.

The reviewer's exact scenario is logically sound but the state is only reachable via direct test poking (which `test_get_index_status_queryable_when_built_with_captured_error` does deliberately as a unit test of the priority-flip branch logic). Adding a docstring warning about a state production code can't produce would itself be misleading — operators following the warning would look for the condition and not find it. And once the follow-up PR aligns `is_queryable()`'s body with `get_index_status`'s 2-field check, the warning would become flat-out wrong and need removal — exactly the forward-reference-fueled doc rot the project's no-forward-references rule warns against.

The in-code comment at the divergence point (`collection.py` `get_index_status` body) documents the divergence for source readers. That's the right surface.

**Finding 2 (Minor: `wait_until_queryable` docstring "per `is_queryable`")** — deferring per your own "Low priority since #539 will touch this file anyway" assessment. The follow-up PR rewrites this method body wholesale (removes the `IndexBuildFailedError` branch); the docstring will be rewritten at the same time to describe the simplified semantics. Polishing it now and then again in the follow-up creates churn.

**Finding 3 (Nit: `docs/design.md` future-tense)** — fixed in cb3483a. The "once the background indexer (#513) lands it will block on a completion event" framing was stale — PR #529 already shipped that. Rewrote to present-tense description of what the method does today.

Re: the positive callouts — agreed, and the `_require_built()` rename was actually surfaced during brainstorm when we realized the old `_require_index_ready()` name mislabeled a single-field check.

pvliesdonk:
Re-review on cb3483a brought me back to Findings 2 and 3. On reflection my earlier deferral of Finding 2 was the forward-reference trap — "next PR will fix it" doesn't apply when the current PR's docstring is misleading at HEAD. Within #538 alone, `wait_until_queryable`'s docstring claimed delegation to `is_queryable()` but the implementation duplicates the field checks. A reader expects a delegation; not finding one is the bug regardless of what #539 does.

Fixed in acc11e4:

- **Finding 2:** `wait_until_queryable` docstring dropped the "per `is_queryable`" specifier. The Control-flow section already enumerates the three raise paths; the simpler "Block until the FTS index is queryable, or raise" is honest without implying delegation.

- **Finding 3:** added `assert col.is_queryable() is False` to `test_get_index_status_queryable_when_built_with_captured_error` with a comment describing the intentional asymmetry between the two surfaces. The divergence is now explicit in test code rather than only in `get_index_status`'s body comment — a future change that touches `is_queryable()`'s body will force a deliberate update to this test.

- **Finding 1:** maintain deferral. The argument was correctness, not scope: tracing the state machine, every production path that captures `_background_build_error` also resets `_index_built=False` (worker via `build_index()`'s pre-rebuild reset at line 802; foreground `build_index(force=True)` same; warm-restart short-circuit clears the error at line 778). The `_index_built=True AND error set` state is only reachable via direct test injection. Adding a docs note for a state production code can't produce would itself be misleading.

PR #567, opened 2026-06-01
feat(git): use OIDC claim values as git commit identity per request

### LINKED ISSUE #565: feat(git): use OIDC claim values as git commit identity via GIT_COMMIT_NAME_CLAIM / GIT_COMMIT_EMAIL_CLAIM
## Summary

In multi-user OIDC-authenticated deployments, every auto-commit lands with the server's static configured identity (`MARKDOWN_VAULT_MCP_GIT_COMMIT_NAME` / `MARKDOWN_VAULT_MCP_GIT_COMMIT_EMAIL`). There is no way to attribute commits to the individual authenticated user.

## Observed

When multiple users share a vault over OIDC auth, all write operations (create, edit, delete, rename) produce git commits with the same server-bot identity regardless of which user is authenticated. The authenticated user's identity, available in the OIDC token's claims (e.g. `name`, `email`, `preferred_username`), is not used.

Originally reported by an external contributor in #524, who implemented a YAML-based subject→identity mapping. The simpler direction: two new env vars that name the OIDC claim key to read, with the static configured values as fallback.

## Expected

When `MARKDOWN_VAULT_MCP_GIT_COMMIT_NAME_CLAIM` and/or `MARKDOWN_VAULT_MCP_GIT_COMMIT_EMAIL_CLAIM` are set (e.g. `name`, `email`), the server reads the named claim from the current request's OIDC access token and uses it as the git commit identity for that request. When no token is present or the named claim is absent, the existing static `GIT_COMMIT_NAME` / `GIT_COMMIT_EMAIL` values are used unchanged.

## Scope boundary

This issue covers the git commit identity (both author and committer fields). The author/committer distinction — where the committer stays as the bot and only the author field reflects the OIDC user — is a separate concern and should be tracked separately.

## Open questions

- `[unverified]`: Whether fastmcp's access-token contextvar propagates correctly into the WriteCallback thread (called inside `asyncio.to_thread`). Needs an integration test.
- `[unverified]`: Which OIDC claim names are standard across common providers (Auth0, Keycloak, Authelia). The env var approach leaves this to the operator; no normalization is attempted.

pvliesdonk:
## Summary

- Adds `MARKDOWN_VAULT_MCP_GIT_COMMIT_NAME_CLAIM` and `MARKDOWN_VAULT_MCP_GIT_COMMIT_EMAIL_CLAIM` env vars
- When set (e.g. `name`, `email`), each commit reads those OIDC claim keys from the current request's access token and uses the values as the git author name/email for that commit
- Falls back to the static `GIT_COMMIT_NAME` / `GIT_COMMIT_EMAIL` values when no token is present, the claim is absent, or the claim value is not a non-empty string
- Non-OIDC deployments (no claim env vars set) are completely unaffected — backward-compatible by design

Simpler alternative to the YAML mapping approach proposed in #524. Closes #565.

## How it works

`_extract_claim(claim_key)` reads `fastmcp.server.dependencies.get_access_token()`, which is a `contextvars.ContextVar` that propagates correctly into the `WriteCallback` thread spawned by `asyncio.to_thread()`. No new dependency or pvl-core version bump required.

## Test plan

- [x] `TestExtractClaim` (7 unit tests): all `_extract_claim` failure modes — no token, absent claim, empty string, non-string value, non-dict claims, `None` key
- [x] `TestOidcClaimGitIdentity` (6 integration tests): actual git commits verified via `git log --format=%aN/%aE` — name claim, email claim, both claims, no-token fallback, absent-claim fallback, backward-compat (no claim config)
- [x] `TestGitClaimConfig` (4 tests): env var loading, None defaults, claim keys passed to `GitWriteStrategy`
- [x] Full suite: 1747 passed, 1 pre-existing skip, 0 failures

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,6 +254,8 @@ Backward compatibility: `MARKDOWN_VAULT_MCP_GIT_TOKEN` without `GIT_REPO_URL` st
 | `MARKDOWN_VAULT_MCP_GIT_PUSH_DELAY_S` | `30` | Seconds of write-idle time before pushing; `0` = push only on shutdown |
 | `MARKDOWN_VAULT_MCP_GIT_COMMIT_NAME` | `markdown-vault-mcp` | Git committer name for auto-commits; **set this in Docker** where `git config user.name` is empty |
 | `MARKDOWN_VAULT_MCP_GIT_COMMIT_EMAIL` | `noreply@markdown-vault-mcp` | Git committer email for auto-commits |
+| `MARKDOWN_VAULT_MCP_GIT_COMMIT_NAME_CLAIM` | — | OIDC claim key to use as the commit author name (e.g. `name`); overrides `GIT_COMMIT_NAME` per-request when an OIDC token is present |
+| `MARKDOWN_VAULT_MCP_GIT_COMMIT_EMAIL_CLAIM` | — | OIDC claim key to use as the commit author e-mail (e.g. `email`); overrides `GIT_COMMIT_EMAIL` per-request when an OIDC token is present |
 | `MARKDOWN_VAULT_MCP_GIT_LFS` | `true` | Enable Git LFS — runs `git lfs pull` on startup to fetch LFS-tracked attachments (PDFs, images). Set to `false` for repos without LFS. |
 
 ### Attachments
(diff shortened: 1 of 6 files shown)


pvliesdonk:
**Re Gemini's `try...except LookupError` suggestion:** not needed.

`fastmcp.server.dependencies.get_access_token()` already suppresses `LookupError` internally via `contextlib.suppress(LookupError)` when reading the MCP SDK request context (verified: `server/dependencies.py` line 373). It returns `None` when called outside a request context — it never raises. Our existing `if token is None: return None` guard handles that case.

Added the two non-blocking items from the Claude review: a comment on `_write_conflict_files` explaining the static identity rationale, and a `logger.debug` line after the effective identity is resolved.

PR #569, opened 2026-06-01
feat(git): split commit author from committer for per-user attribution

### LINKED ISSUE #568: feat(git): split commit author vs committer identity for per-user attribution
## Summary

Auto-commits currently use the same identity for both git author and committer. In multi-user deployments where the MCP server commits on behalf of authenticated users, the git author field should reflect the human user while the committer field stays as the configured server identity — but there is no way to express this split today.

## Observed

`_stage_and_commit` sets `user.name` and `user.email` via `-c` flags on the `git commit` command. These flags set **both** the author and the committer to the same value. Whether using the static configured identity or the OIDC claim override introduced in #567, there is no way to make git record "authored by Alice, committed by vault-bot" — both fields always end up identical.

This is visible in `git log --format="%aN <%aE> / %cN <%cE>"`: author and committer are always the same string.

[verified: by reading `_stage_and_commit` in `src/markdown_vault_mcp/git.py` and confirming only `-c user.name` / `-c user.email` flags are used — commit e.g. 2502b9c]

## Expected

When a server-level committer identity (the bot/service account) is configured alongside a per-request author identity (e.g. from OIDC claims), git should record:

- **author** — the human who made the change (from per-request identity)
- **committer** — the service account that committed it (from the static server config)

This is the standard git distinction for "authored by X, committed by Y" workflows (e.g. how GitHub records commits applied via the API).

## Context

Surfaced during review of #565 / #567, which added OIDC claim-based identity override. The current implementation replaces the full identity (both author and committer) with the claim values. The author/committer split was explicitly deferred from that PR's scope.

## Scope boundary

This issue is about the git author/committer field distinction only. It does not cover push credentials, OIDC configuration, or how the per-request identity is sourced (static env vars, OIDC claims, or other means).

## Open questions

- `[unverified]`: Whether git's `--author "Name <email>"` flag (as opposed to `-c user.name`) is the right mechanism, or whether separate `-c author.name` / `-c author.email` overrides exist.
- `[unverified]`: Whether the current `WriteCallback` signature can carry both identities without a breaking change, or whether a different approach is needed.

pvliesdonk:
## Summary

- `_stage_and_commit` gains `author_name` / `author_email` optional parameters
- When provided and different from `commit_name` / `commit_email`, a `--author "Name <email>"` flag is added to `git commit` so git records separate author and committer fields
- `GitWriteStrategy.__call__` now routes the static configured identity to the committer and the OIDC-claim-resolved effective identity to the author only
- Backward-compatible: when no claim is configured, `author_name`/`author_email` are `None`, no `--author` flag is added, and `%aN == %cN` as before

Closes #568. Builds on #567.

## Behavior change from #567

#567 replaced **both** author and committer with the claim values. This PR corrects that: the committer stays as the static server identity (`GIT_COMMIT_NAME` / `GIT_COMMIT_EMAIL`) and only the author field reflects the per-request OIDC identity. This is the standard "authored by human, committed by bot" git pattern.

## Test plan

- [x] `TestStageAndCommitAuthorSplit` (5 tests): both fields differ, only name differs, only email differs, no author args (backward compat), author == committer (no `--author` flag added)
- [x] `TestOidcClaimAuthorCommitterSplit` (4 tests): claim → author=claim/committer=static; no claim → author==committer==static; no token → fallback; asymmetric name-only claim
- [x] Existing `TestCommitterIdentityInCommit` stub updated to accept new kwargs — assertions unchanged
- [x] Full suite: 1769 passed, 0 failures

### DIFF
--- a/src/markdown_vault_mcp/git.py
+++ b/src/markdown_vault_mcp/git.py
@@ -616,8 +616,14 @@ class GitWriteStrategy:
                     self._git_root,
                     path,
                     operation,
-                    commit_name=effective_name,
-                    commit_email=effective_email,
+                    commit_name=self._commit_name,
+                    commit_email=self._commit_email,
+                    author_name=effective_name
+                    if effective_name != self._commit_name
+                    else None,
+                    author_email=effective_email
+                    if effective_email != self._commit_email
+                    else None,
                 )
             if self._enable_push:
                 self._schedule_push()
@@ -2515,6 +2521,8 @@ def _stage_and_commit(
     operation: Literal["write", "edit", "delete", "rename"],
     commit_name: str = GitWriteStrategy.DEFAULT_COMMIT_NAME,
     commit_email: str = GitWriteStrategy.DEFAULT_COMMIT_EMAIL,
+    author_name: str | None = None,
+    author_email: str | None = None,
 ) -> None:
     """Stage and commit a single file change (no push).
 
@@ -2524,6 +2532,13 @@ def _stage_and_commit(
         operation: The write operation type.
         commit_name: Git committer name (overrides git config).
         commit_email: Git committer email (overrides git config).
+        author_name: Git author name.  When provided and different from
+            *commit_name* (or when *author_email* also differs), the commit
+            is recorded with a separate ``--author`` field so the author and
+            committer identities are distinct.  Falls back to *commit_name*
+            when ``None``.
+        author_email: Git author e-mail.  Same split semantics as
+            *author_name*.  Falls back to *commit_email* when ``None``.
     """
     root = str(git_root)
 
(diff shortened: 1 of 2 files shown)


pvliesdonk:
Addressed both findings:

**Finding 1 (sanitization):** Added `_sanitize_git_identity()` which strips `\r`, `\n`, `<`, `>` from values before interpolating into the `--author` string. The newline stripping is the key protection — it prevents injected text from becoming a new commit-object header line. `<`/`>` are stripped to preserve the `Name <email>` format. Added two regression tests (newline stripped, angle brackets stripped). Switched the `author_name or commit_name` fallback to `author_name if author_name is not None else commit_name` as flagged.

**Finding 2 (docstring clarity):** Updated the `author_name`/`author_email` docstring to explicitly state that the fields are evaluated independently and that a `None` value falls back to the committer field, producing a mixed identity string.

**Re Gemini:** no findings — nothing to address.

pvliesdonk:
**Finding 1 fixed.** The comparison now uses `_sanitize_git_identity(commit_name)` / `_sanitize_git_identity(commit_email)` on both sides, closing both modes:
- Mode A: `commit_name` with angle brackets no longer triggers spurious `--author` when no OIDC claim is configured.
- Mode B: an OIDC claim that sanitizes to the same value as the committer is correctly treated as equal (no `--author` added).

**Finding 2 fixed.** Added `logger.debug("git_identity_sanitized ...")` when sanitization materially changes an `author_name` or `author_email` value.

Two new regression tests: one for each mode.

PR #589, opened 2026-06-03
fix(indexing): don't conflate cancellation with failure in async done-callbacks

### LINKED ISSUE #584: Async reindex/build_embeddings done-callbacks report CancelledError as a job failure in get_index_status
## Summary

The async writer-job done-callbacks catch `BaseException`, so a cancelled (or interrupted) reindex/embeddings Future is recorded and surfaced via `get_index_status` as if it were a genuine job failure.

## Observed

In `src/markdown_vault_mcp/indexing/coordinator.py`, `_on_reindex_done` and `_on_build_embeddings_done` (registered as `Future.add_done_callback`) record the outcome like this:

```python
try:
    fut.result()
    self._last_reindex_error = None
except BaseException as exc:
    self._last_reindex_error = exc
    logger.exception("Async reindex job failed")
```

Because the catch is `BaseException` (not `Exception`), a `concurrent.futures.CancelledError` — or a `KeyboardInterrupt` / `SystemExit` propagated into the Future result — is stored in `_last_reindex_error` / `_last_build_embeddings_error` and reported through `get_index_status()` as `last_reindex_error="..."` / `last_build_embeddings_error="..."`. A caller reading status after a cancellation sees a phantom "error" that is indistinguishable from a real indexing failure.

Cause: the `except BaseException` capture conflating cancellation with failure is `[verified: read at coordinator.py _on_reindex_done / _on_build_embeddings_done, PR #582 commit b4ca995]`.

## Expected

Cancellation should be distinguishable from a genuine job failure in the surfaced status (or not surfaced as an error at all).

## Context

- Verified by reading `src/markdown_vault_mcp/indexing/coordinator.py` at PR #582 (commit `b4ca995`).
- Pre-existing behaviour: moved verbatim from `collection.py` (the #561 async-error-capture code); **not introduced by #582**. Surfaced by the silent-failure review lens while reviewing #582 and deferred as a behaviour-changing follow-up.

## Scope boundary

Not about `get_index_status`'s `list_notes` fallback — that is a separate observation (filed separately). Not part of #576, which was a behaviour-preserving extraction.

## Open questions

- `[unverified]` whether any production / MCP path actually cancels these Futures — the MCP layer fires-and-forgets, so cancellation may be rare in practice.
- `[unverified]` whether `KeyboardInterrupt` / `SystemExit` can realistically reach these callbacks given the writer thread's own `BaseException` handling.

### LINKED ISSUE #590: build_index_async._on_done records a shutdown-drain CancelledError as status=failed
## Summary

`build_index_async`'s `_on_done` callback records a writer-shutdown-drain
`CancelledError` as a build failure, so `get_index_status()` can report
`status="failed"` on a normal shutdown rather than a genuine build error.

## Observed

In `src/markdown_vault_mcp/indexing/coordinator.py`, `_on_done`
(the `build_index_async` done-callback) catches the outcome of
`fut.result()` with `except BaseException as exc: ... fail_build(exc)`.

When the writer terminates on a `BaseException`, its close-and-drain block
(`index_writer.py:333-334`) cancels every still-pending future regardless
of kind — `pending_future.cancel()` / `set_exception(_CancelledError())`.
A `BuildIndex` future still queued at that moment therefore completes with
`concurrent.futures.CancelledError`; `_on_done` then calls
`self._readiness.fail_build(CancelledError)`, and `ReadinessState.status_fields`
(`readiness.py:99-101`) surfaces that as `status="failed"` with the
cancellation as the reported `error`.

This is the same conflation #584 fixed for the reindex/embeddings
callbacks (`_on_reindex_done` / `_on_build_embeddings_done`), which now
carve out `except CancelledError` ahead of the catch-all. `_on_done` was
explicitly out of #584's scope (#584 names only the reindex/embeddings
callbacks).

Cause: [verified: traced `_on_done` (coordinator.py:287-308), the writer
drain (index_writer.py:323-334), and `ReadinessState.status_fields`
(readiness.py:99-101)].

## Expected

A `BuildIndex` future cancelled by the writer-shutdown drain should not be
surfaced as a build failure — `get_index_status()` should not report
`status="failed"` solely because the index build was cancelled during
shutdown.

## Context

- Commit: 2bed46f (PR #589 branch) / present on `main` since #588.
- Reachability is narrow: a `BuildIndex` future must still be queued when
  an earlier job raises a true `BaseException` on the writer thread.
  `close()` alone drains queued jobs through the FIFO and does not cancel
  them; only the writer's `except BaseException` branch cancels pending
  futures.

## Scope boundary

Not about the reindex/embeddings callbacks (fixed in #584/#589). Only the
`build_index_async._on_done` path and the `status="failed"` surface.

## Open questions

- [unverified]: whether `mark_done()` in `_on_done`'s `finally` interacts
  with a recorded cancellation in a way that affects `wait_until_queryable`
  (see #586 for the related `reason` conflation).

pvliesdonk:
Closes #584, closes #590.

## Problem

All three async done-callbacks caught a writer-shutdown-drain cancellation
as a failure. During a drain, `IndexWriter` cancels pending futures
(`index_writer.py:333-334`), so `fut.result()` in the callback raises
`concurrent.futures.CancelledError` — an `Exception` subclass — which the
catch-all treated as a job failure.

## Fix

Split `except CancelledError` (debug-log, not recorded) ahead of the
catch-all in all three callbacks:

- **`_on_reindex_done` / `_on_build_embeddings_done` (#584):** the recorded
  error is a diagnostic-only `get_index_status` field. Catch-all stays
  `except BaseException`: these run on the writer thread inside
  `Future.set_exception` (`index_writer.py:306`, inside the writer's own
  `except BaseException`), so an escaping exception would skip the writer's
  close-and-drain and strand pending futures. Recording is harmless;
  propagating strands.

- **`_on_build_index_done` (#590):** higher impact — it calls `fail_build()`,
  which mutates the readiness state machine (drives top-level `status`), not
  just a diagnostic field. A drain-cancelled `BuildIndex` future previously
  flipped status to `"failed"`; now the cancellation skips `fail_build`
  (`finally: mark_done()` still fires so waiters unblock to `never_built`).
  Extracted from a closure to a method so all three callbacks are uniform
  and directly testable; #585's `set_build_completed` `except Exception` and
  BaseException-propagation are preserved unchanged.

## Tests

Symmetric across all three callbacks: cancellation-ignored,
genuine-`Exception`-recorded, non-cancellation-`BaseException`
(reindex/embeddings: recorded without propagating; build_index: drives
`fail_build`, with the `except BaseException` breadth mutation-pinned). Plus
`test_on_build_index_done_cancellation_unblocks_waiters` pins the
`finally: mark_done()` liveness invariant (mutation-verified).

Full suite: 1885 passed, 1 skipped.

## Docs

Internal bugfix to private callbacks — no user-facing surface change
(`get_index_status` field shape is unchanged; a phantom value is merely no
longer written), so no docs/ or README update is required.

### DIFF
--- a/src/markdown_vault_mcp/indexing/coordinator.py
+++ b/src/markdown_vault_mcp/indexing/coordinator.py
@@ -14,7 +14,7 @@ from __future__ import annotations
 import logging
 import threading
 import time
-from concurrent.futures import Future
+from concurrent.futures import CancelledError, Future
 from typing import TYPE_CHECKING, Any
 
 from markdown_vault_mcp.exceptions import IndexUnavailableError
@@ -284,38 +284,47 @@ class IndexWriteCoordinator:
             self._readiness.fail_build(exc)
             raise
 
-        def _on_done(fut: Future[IndexStats]) -> None:
-            try:
-                try:
-                    fut.result()
-                except BaseException as exc:
-                    logger.exception("Async index build failed")
-                    self._readiness.fail_build(exc)
-                    return
-                try:
-                    self._fts.set_build_completed()
-                except Exception as exc:
-                    logger.exception(
-                        "Async index build: set_build_completed failed after build"
-                    )
-                    self._readiness.fail_build(exc)
-                    return
-                self._readiness.mark_built()
-            finally:
-                # Guarantee waiters unblock even if an unexpected BaseException
-                # escapes set_build_completed: it propagates (so the worker can
-                # respond to the signal), but the done-event must still be set.
-                self._readiness.mark_done()
-
-        future.add_done_callback(_on_done)
+        future.add_done_callback(self._on_build_index_done)
         return future
+
+    def _on_build_index_done(self, fut: Future[IndexStats]) -> None:
+        """Finalize the readiness state machine from the async build Future (#585)."""
+        try:
+            try:
+                fut.result()
+            except CancelledError:
+                # Drain cancellation is not a build failure — skip fail_build (#590).
+                logger.debug("async_build_index_cancelled")
+                return
+            except BaseException as exc:
+                logger.exception("Async index build failed")
+                self._readiness.fail_build(exc)
+                return
+            try:
+                self._fts.set_build_completed()
+            except Exception as exc:
+                logger.exception(
+                    "Async index build: set_build_completed failed after build"
+                )
+                self._readiness.fail_build(exc)
+                return
+            self._readiness.mark_built()
+        finally:
+            # Guarantee waiters unblock even if an unexpected BaseException
+            # escapes set_build_completed: it propagates (so the worker can
+            # respond to the signal), but the done-event must still be set.
+            self._readiness.mark_done()
 
     def _on_reindex_done(self, fut: Future[ReindexResult]) -> None:
         """Capture async reindex outcome for visibility via get_index_status (#561)."""
         try:
             fut.result()
             self._last_reindex_error = None
+        except CancelledError:
+            # Cancellation (writer-shutdown drain) is not a failure — don't record (#584).
+            logger.debug("async_reindex_cancelled")
         except BaseException as exc:
+            # Record, never propagate — escaping strands the writer (#584).
             self._last_reindex_error = exc
             logger.exception("Async reindex job failed")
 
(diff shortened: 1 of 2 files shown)


pvliesdonk:
Both review findings are addressed in the force-pushed revision:

- **Finding 1 (Medium — `_on_done` has the same bug):** fixed. The `_on_done` closure is extracted into a `_on_build_index_done` method with the same `except CancelledError` carve-out, so a drain-cancelled `BuildIndex` future no longer flips the readiness state to `"failed"`. The `finally: mark_done()` still fires on the cancellation path; `test_on_build_index_done_cancellation_unblocks_waiters` pins that liveness invariant (mutation-verified: removing `mark_done` makes only that test fail). This higher-impact case is now folded into the PR, which also closes #590.
- **Finding 2 (Low — missing symmetric test):** added `test_on_build_embeddings_done_records_genuine_failure`, plus full cancellation/genuine-failure/BaseException symmetry across all three callbacks.

The `except BaseException` breadth on the build callback is deliberately pinned by `test_on_build_index_done_records_baseexception` so a future narrowing to `except Exception` (which would re-open the #585 strand) fails fast.

PR #592, opened 2026-06-03
fix(indexing): distinguish failing list_notes from empty index + clear stale sync-build error

### LINKED ISSUE #583: get_index_status: failing list_notes() is reported as documents_indexed=0, indistinguishable from an empty index
## Summary

`get_index_status()` reports `documents_indexed: 0` both when the index is legitimately empty and when the underlying FTS query fails, with no error visible at the default log level.

## Observed

In `src/markdown_vault_mcp/indexing/coordinator.py`, `get_index_status()` wraps the document count in a broad handler:

```python
try:
    documents_indexed = len(self._fts.list_notes())
except Exception:
    logger.debug("get_index_status: list_notes failed; reporting 0", exc_info=True)
    documents_indexed = 0
```

When `list_notes()` raises (e.g. a `sqlite3.OperationalError` from a locked or corrupt DB), the status payload returns `documents_indexed: 0` and the only trace is a `DEBUG`-level log line — invisible at the default (`INFO`) level. An operator calling the status tool to diagnose trouble sees a clean-looking `{"documents_indexed": 0, "status": ...}` and cannot distinguish "empty vault" from "FTS query is throwing".

Cause: the broad `except Exception` + `DEBUG`-only log is `[verified: read at coordinator.py get_index_status, PR #582 commit b4ca995]`.

## Expected

A status tool — whose purpose is operator visibility — should let the caller distinguish an empty index from a failing index query.

## Context

- Verified by reading `src/markdown_vault_mcp/indexing/coordinator.py` at PR #582 (commit `b4ca995`).
- Pre-existing behaviour: moved verbatim from `collection.py` (pre-#576 location ~`774-781`) during the #576 extraction; **not introduced by #582**. Surfaced by the silent-failure review lens while reviewing #582 and deferred as a behaviour-changing follow-up.

## Scope boundary

Not about the async done-callback error capture — that is a separate observation (filed separately). Not part of #576, which was a behaviour-preserving extraction.

## Open questions

- `[unverified]` which exception types `list_notes()` realistically raises (is `sqlite3.Error` the right narrowing target, or can other types occur?).
- `[unverified]` whether the surfaced status should carry a distinct field for "count unavailable" vs. report `0`.

### LINKED ISSUE #587: ReadinessState.begin_sync_build leaves stale _error — get_index_status reports 'failed' during an active sync rebuild
## Summary

`begin_sync_build()` clears only `_index_built`; a stale `_error` from a prior failed async build then makes `get_index_status()` report `status="failed"` during the window of an actively-running sync rebuild.

## Observed

In `src/markdown_vault_mcp/indexing/readiness.py`, `begin_sync_build()` sets `_index_built=False` only — it leaves `_error` and the done-event untouched (by design, mirroring the pre-#576 sync `build_index` begin). Sequence:

1. An async build fails → `fail_build(exc)` sets `_error` and the done-event.
2. A subsequent sync `build_index()` calls `begin_sync_build()` (clears only `_index_built`) and starts the build.
3. During that window (before `mark_built()`), a concurrent `get_index_status()` sees `is_queryable()=False`, `_done.is_set()=True`, `_error is not None` → `status_fields()` returns `status="failed"` with the stale prior error — even though a build is actively in progress.

Cause: `[verified: read at readiness.py begin_sync_build + status_fields; traced against origin/main (pre-#576) collection.py sync build_index, which likewise set only _index_built=False]`.

## Expected

Status during an active sync rebuild should not report `"failed"` from a stale prior error. The module's stated invariant is "a captured build error is diagnostic state, never a control-flow gate" — surfacing it as `"failed"` mid-build is misleading.

## Context

Pre-existing: relocated verbatim from `collection.py` during #576 (PR #582). **Not introduced by #582** — #576 deliberately preserved the sync/async `begin` divergence; this status-contamination side-effect is the observation. Surfaced by `claude-review` on PR #582 (finding #3).

## Scope boundary

Not about the build-exception recording or the `reason` conflation (separate issues). Not about #582.

## Open questions

- `[unverified]` whether clearing `_error` in `begin_sync_build()` would have any unintended effect on the diagnostic-error contract elsewhere.
- `[unverified]` how often a sync `build_index()` follows a failed async build in practice (the deprecated/legacy paths vs. the lifespan path).

### LINKED ISSUE #591: sync build_index does not record a BaseException as failed — stranded 'building' (asymmetric with async)
## Summary

A `BaseException` raised by a sync `build_index()` (from the build job or
`set_build_completed()`) is not recorded via `fail_build`, so the readiness
state machine is left reporting `status="building"` with `error=null`
indefinitely — asymmetric with the async path, which records a build-job
`BaseException` as `"failed"`.

## Observed

`IndexWriteCoordinator.build_index()` (`coordinator.py:208-215`) wraps both
`self._writer.submit(BuildIndex(...)).result()` and
`self._fts.set_build_completed()` in a single `except Exception` handler:

```python
self._readiness.begin_sync_build()      # clears _index_built + _error (#587)
self._fts.clear_build_completed()
try:
    result = self._writer.submit(BuildIndex(force=force)).result()
    self._fts.set_build_completed()
except Exception as exc:                  # <-- not BaseException
    self._readiness.fail_build(exc)
    raise
self._readiness.mark_built()
```

A `BaseException` (e.g. `KeyboardInterrupt`, `SystemExit`,
`asyncio.CancelledError`) from the build job is captured onto the Future by
the writer and re-raised by `.result()` on the calling thread; it bypasses
`except Exception`, so `fail_build` is never called and the exception
propagates. The resulting readiness state is `_index_built=False`,
`_error=None`, `_done` set, which `ReadinessState.status_fields()` reports as
`{"status": "building", "error": None}` — permanently, since nothing is
building.

This is now more visible after #587: `begin_sync_build()` clears a stale
`_error`, so even a prior failure's diagnostic no longer surfaces on this
path (before #587 a stale prior error would have shown a misleading
`"failed"`).

The **async** path differs: `_on_build_index_done` records a build-job
`BaseException` via `fail_build` -> `"failed"` (#589,
`test_on_build_index_done_records_baseexception`), while a `set_build_completed`
`BaseException` is deliberately *not* recorded and propagates (#585,
`test_async_set_completed_base_exception_does_not_strand`).

Cause: [verified: read `coordinator.py:208-215`, `readiness.py`
`status_fields`; surfaced by the silent-failure review of the #583/#587 PR
(#589 follow-up); see also the long-standing "sync BaseException path
untested and unhandled" note].

## Expected

A `BaseException` terminating a sync `build_index()` should leave the
readiness state diagnosable (status `"failed"` with the exception surfaced),
symmetric with the async path — not stranded at `"building"` with no error.

## Context

- Reachability is low in production: the build runner executes on the
  writer thread, which does not receive process signals
  (`KeyboardInterrupt`/`SystemExit` go to the main thread). The path is
  reachable mainly by a runner that itself raises a `BaseException`.
- Pre-existing: the `except Exception` predates #587 (the sync handler was
  narrowed from `BaseException` to `Exception` during #585). Not introduced
  by #583/#587.

## Scope boundary

Not about the `get_index_status` `list_notes` narrowing (#583) or the
`begin_sync_build` stale-error clear (#587). Not about the async done-callbacks
(#584/#585/#589/#590).

## Open questions

- [unverified] whether the fix is a simple widen to `except BaseException`
  (record job + set_completed `BaseException` as `"failed"`), or whether it
  must restructure into two try-blocks to preserve the async distinction
  where a `set_build_completed` `BaseException` is deliberately *not* recorded
  (#585). The single sync `try` cannot express that distinction as-is.
- [unverified] whether recording a propagating signal (e.g. `KeyboardInterrupt`)
  as a build `"failed"` is desirable, or whether the sync path should only
  guarantee a non-`"building"` terminal state.

pvliesdonk:
Closes #583, closes #587, closes #591.

Three status-surface bugs in the indexing readiness path (all pre-existing — moved verbatim from `collection.py` during the #576 extraction).

## #583 — failing `list_notes()` masked as an empty index

`get_index_status()` wrapped the document count in a broad `except Exception` that reported `documents_indexed=0` and logged at **DEBUG** whenever `list_notes()` raised — masking a locked/corrupt/closed FTS DB (`sqlite3.Error`) as a clean, empty index. Narrowed to `except sqlite3.Error`, log **WARNING**, keep `documents_indexed: 0`, and add a **`documents_indexed_error: str | None`** field so callers can tell an empty index from a failed read. Non-sqlite exceptions now propagate.

## #587 — stale `"failed"` + premature `never_built` during a sync rebuild

`begin_sync_build()` cleared only `_index_built`, so a stale `_error` from a prior failed async build made `get_index_status()` report `status="failed"` during an active sync rebuild. It now also clears `_error` **and the done-event**.

Clearing the done-event addresses @gemini-code-assist's HIGH finding: with `_done` left set, a concurrent `wait_until_queryable()` returned `never_built` immediately instead of blocking for the active build (inconsistent with the `"building"` the same state reports). Now a concurrent reader blocks correctly.

## #591 — sync `BaseException` no longer strands status

With the done-event cleared at begin, a `BaseException` escaping the sync build would strand it. Restructured `build_index`'s cold path to **fully mirror the async `_on_build_index_done`**:
- drain `CancelledError` → **not** recorded as a failure (skip `fail_build`, re-raise) — mirrors #590;
- build-job `BaseException` → recorded `"failed"`, re-raised;
- `set_build_completed` `BaseException` → propagates without being recorded (mirrors #585's `except Exception`);
- `finally: mark_done()` re-sets the done-event on every exit so waiters never hang.

## Tests

Symmetric, mutation-verified coverage: #583 (sqlite-error / success / non-sqlite-propagate); #587 (readiness unit clears built+error+done; a threaded integration test asserting a concurrent reader sees `"building"` AND blocks — `wait_until_queryable` times out, not `never_built`); #591 (job `BaseException` → "failed"; `set_completed` `BaseException` → `never_built`, no strand; drain `CancelledError` → not failed). All four load-bearing pieces (done-clear, CancelledError carve-out, `except BaseException`, `finally`) are individually mutation-pinned. Full suite: 1893 passed.

## Docs

`get_index_status` tool docstring, the `Collection.get_index_status` facade docstring (now "eleven-key"), `docs/tools/index.md`, and the `docs/design.md` status-surface paragraph document the new field.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -548,7 +548,12 @@ file-mutation lock.
 **Status surface.** `Collection.get_index_status()` merges the writer's
 non-blocking snapshot (`queue_depth`, `in_flight`, `dirty_paths`,
 `dirty_embeddings`) into its response shape, so operators can observe
-exactly what the writer is doing without taking any lock.
+exactly what the writer is doing without taking any lock. The document
+count is read from `FTSIndex.list_notes()`; a `sqlite3.Error` there
+(locked / corrupt / closed DB) is not an empty index, so `get_index_status`
+keeps `documents_indexed` at `0` and surfaces the reason in a separate
+`documents_indexed_error` field (logged at `WARNING`) rather than masking
+the failure as a clean zero (#583).
 
 The contract is:
 
--- a/docs/tools/index.md
+++ b/docs/tools/index.md
@@ -213,7 +213,12 @@ build attempt.
 **Returns:**
 - `status`: `"queryable"`, `"building"`, or `"failed"`.
 - `documents_indexed`: count of documents committed to the FTS index
-  right now (rises during `"building"`).
+  right now (rises during `"building"`). `0` both for an empty index
+  and when the count could not be read — check `documents_indexed_error`
+  to tell them apart.
+- `documents_indexed_error`: `null` on a normal read; the SQLite error
+  message when the document count could not be read (e.g. a locked or
+  closed database), in which case `documents_indexed` is `0`.
 - `error`: `null` unless the background build raised; otherwise the
   exception message.
 
--- a/src/markdown_vault_mcp/_server_tools.py
+++ b/src/markdown_vault_mcp/_server_tools.py
@@ -654,6 +654,12 @@ def register_tools(mcp: FastMCP, *, transport: str = "stdio") -> None:
               ``"failed"``.
             - documents_indexed (int): Count of documents committed to
               the FTS index right now (rises during ``"building"``).
+              ``0`` both for an empty index and when the count could not
+              be read — see ``documents_indexed_error`` to tell them apart.
+            - documents_indexed_error (str | None): ``None`` on a normal
+              read; the SQLite error message when the document count
+              could not be read (e.g. a locked or closed database), in
+              which case ``documents_indexed`` is ``0``.
             - error (str | None): ``None`` unless the background build
               raised.
         """
--- a/src/markdown_vault_mcp/collection.py
+++ b/src/markdown_vault_mcp/collection.py
@@ -511,14 +511,16 @@ class Collection:
         return self._coordinator.wait_for_drain(timeout)
 
     def get_index_status(self) -> dict[str, Any]:
-        """Return a non-blocking ten-key snapshot of build + writer state.
+        """Return a non-blocking eleven-key snapshot of build + writer state.
 
         Keys: ``status`` (``"queryable"`` | ``"building"`` | ``"failed"``),
-        ``documents_indexed``, ``error``, ``last_reindex_error``,
-        ``last_build_embeddings_error``, plus ``queue_depth``, ``in_flight``,
-        ``dirty_paths``, ``dirty_embeddings``, ``write_generation`` merged from
-        the writer. A captured build error appears in ``error`` as diagnostic
-        context without demoting a ``queryable`` status.
+        ``documents_indexed``, ``documents_indexed_error``, ``error``,
+        ``last_reindex_error``, ``last_build_embeddings_error``, plus
+        ``queue_depth``, ``in_flight``, ``dirty_paths``, ``dirty_embeddings``,
+        ``write_generation`` merged from the writer. A captured build error
+        appears in ``error`` as diagnostic context without demoting a
+        ``queryable`` status; ``documents_indexed_error`` carries a SQLite
+        read failure (``documents_indexed`` stays ``0``) (#583).
         """
         return self._coordinator.get_index_status()
 
--- a/src/markdown_vault_mcp/indexing/coordinator.py
+++ b/src/markdown_vault_mcp/indexing/coordinator.py
@@ -12,6 +12,7 @@ coordinator constructs, starts, and closes the single-owner
 from __future__ import annotations
 
 import logging
+import sqlite3
 import threading
 import time
 from concurrent.futures import CancelledError, Future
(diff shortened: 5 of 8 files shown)


pvliesdonk:
Thanks — this HIGH finding is correct and is now fixed. `begin_sync_build` clears the done-event (`self._done.clear()`), so a concurrent `wait_until_queryable()` blocks for the active build instead of prematurely raising `never_built`. The readiness unit test now asserts `r.wait(timeout=0) is False`, and a coordinator integration test pins the concurrent-reader behaviour end-to-end (`wait_until_queryable(timeout=0.1)` raises `reason="timeout"`, not `never_built`, mid-rebuild).

One addition beyond the suggested one-liner: clearing the done-event meant a `BaseException` escaping the sync build could *strand* it cleared (a waiter would then hang to its timeout). So `build_index`'s cold path was restructured to fully mirror the async `_on_build_index_done` — a `finally: mark_done()` guarantees the done-event is re-set on every exit, plus a `CancelledError` carve-out (drain cancellation isn't a failure) and `BaseException` recording on the job. That closes #591 in the same change.

PR #582, opened 2026-06-02
refactor: extract index-write state machine into indexing/ package (closes #576)

pvliesdonk:
## Summary

PR1 of the `collection.py` facade-decomposition effort (issue #576). Extracts the index-write state machine out of the 1881-LOC `collection.py` hotspot into a focused `indexing/` package, **behavior-preserving**, with `Collection` reduced to thin delegators.

`collection.py`: **1881 → 1346 LOC** (−535); it now owns no writer/build state.

## What moved

New `src/markdown_vault_mcp/indexing/` package:

- `index_writer.py` — the single-owner `IndexWriter` thread + job dataclasses/runners (git-renamed from `writer.py`, content unchanged).
- `readiness.py` — **`ReadinessState`**: the build-readiness state machine (the `_index_built` + done-event + error triple). Encapsulates the load-bearing invariant from #531: *a captured build error is diagnostic state, never a control-flow gate.* Unit-tested in isolation.
- `coordinator.py` — **`IndexWriteCoordinator`**: owns the `IndexWriter` + a `ReadinessState`; holds sync/async build orchestration, the per-variant async error capture (#561), status/drain observation, dirty routing, and lifecycle. `Collection` delegates to it.

## Behavior preservation

- Public `Collection` API unchanged — every moved method remains a thin delegator; external callers (`_server_*`, CLI, webhook) and their tests are untouched except internal-access repoints.
- Each `ReadinessState` transition mirrors an **exact per-callsite mutation** from the former `collection.py` (the sync-vs-async `begin` divergence — sync clears only `_index_built`, async also clears error + the done-event — is preserved deliberately, not unified).
- `close()` ordering preserved: the coordinator (joins the legacy bg-build thread → closes the writer) runs **before** `fts.close()`, since the writer's drain touches FTS.
- **1867 tests pass** (baseline 1850 + 17 new unit tests for `ReadinessState` / `IndexWriteCoordinator`). Patch coverage **95%**; `mypy --strict` clean on the new `indexing/` package.

## Epic context

PR1 of 4 toward a `Vault` composition root exposing `reader` / `writer` / `graph` / `index` facets (the `index` facet *is* this coordinator). PR2 extracts the write-callback dispatcher; PR3 introduces the facet objects; PR4 flips the public API and removes the flat methods on the in-flight major bump. Design: `docs/superpowers/specs/2026-06-02-collection-facade-decomposition-design.md`.

## Review

[...]

PR #602, opened 2026-06-03
refactor(collection): extract WriteCallbackDispatcher into write_callback.py

### LINKED ISSUE #599: refactor: extract WriteCallbackDispatcher from collection.py into write_callback.py (facade-decomposition PR2)
Part of the `collection.py` facade-decomposition epic (PR2 of 4). PR1 shipped as #582 (extracted `indexing/`). Design: `docs/superpowers/specs/2026-06-02-collection-facade-decomposition-design.md` §"PR2 — callback dispatcher". Refs #576.

## Goal

Move the deferred git-commit write-callback worker (originally #175) out of the `Collection` god-facade into a focused, separately-testable `WriteCallbackDispatcher`. **Behavior-preserving, non-breaking** (no public surface change, `Collection` keeps delegating).

## Deliverables

- New `src/markdown_vault_mcp/write_callback.py` with `WriteCallbackDispatcher` (a *separate concern* — git-commit dispatch, not index mutation — so NOT in the `indexing/` package).
  - Owns the background worker thread, its queue, and its start-lock, plus the `on_write` reference.
  - `.fire(abs_path, content, operation)` — replaces `Collection._fire_write_callback` + `_ensure_callback_worker`.
  - `.close(timeout=30.0)` — the `close()` step-2 drain logic (sentinel → join → warn-on-timeout).
- `Collection`:
  - Constructs `WriteCallbackDispatcher(on_write)`, passes `.fire` as `DocumentManager`'s `on_write_callback`, calls `.close()` in the same teardown slot.
  - Removes `_fire_write_callback` / `_ensure_callback_worker` and the three `_callback_queue` / `_callback_worker` / `_callback_worker_lock` attributes; drops the now-unused `import queue`.
  - **Keeps** `self._on_write` — it has a second, separate use: the `_on_write.close()` resource-teardown at `close()` (closing a non-git-strategy closeable callback), which is unrelated to draining the queue.
- New `tests/test_write_callback.py` unit-pins the dispatcher contract directly (mirrors how PR1 added `test_index_coordinator.py`).
- Docs: the CLAUDE.md module map and `docs/design.md` Project Structure gain `write_callback.py`.

## Invariants to preserve (behavior-preserving extraction)

1. `on_write=None` → `fire()` is a no-op (no enqueue, no worker started).
2. Lazy + idempotent worker start — concurrent `fire()` calls start exactly one worker (guarded by the start-lock).
3. FIFO ordering — single queue + single worker; git commits fire in submission order.
4. A callback raising `Exception` is caught + logged (`error`, `exc_info=True`); the worker continues processing subsequent items (does not die).
5. `close()` drains pending items then joins (bounded ≤ 30 s), logs a WARNING if the worker doesn't finish; is safe when the worker was never started; is safe on a second call.
6. Worker is a `daemon` thread named `"write-callback"`.

## Non-goals

- No behavior change — in particular, the current fire-after-close semantics are preserved as-is (no new `_closed` guard); any such hardening is a separate, justified change.
- No `Collection`→`Vault` rename (PR4), no facet objects (PR3).

## Documentation

- CLAUDE.md Project Structure module map — add `write_callback.py`.
- `docs/design.md` Project Structure — add `write_callback.py`.
- No README / tools / config / resource changes (no user-facing surface change).

pvliesdonk:
Closes #599. Refs #576.

PR2 of the `collection.py` facade-decomposition epic (PR1 was #582, which extracted the `indexing/` package).

## What

Moves the deferred git-commit write-callback worker (originally #175) out of the `Collection` god-facade into a focused, separately-testable `WriteCallbackDispatcher` in `src/markdown_vault_mcp/write_callback.py`. **Behaviour-preserving** — no public surface change.

- `WriteCallbackDispatcher(on_write)` owns the daemon worker thread, the FIFO queue, and the start-lock.
- `.fire(abs_path, content, operation)` replaces `_fire_write_callback` + `_ensure_callback_worker`.
- `.close(timeout=30.0)` is the former `close()` step-2 drain (sentinel → join → warn-on-timeout).
- `Collection` constructs the dispatcher **before** `DocumentManager` (so `on_write_callback=self._write_callback.fire` captures a live reference), and calls `.close(timeout=30.0)` in the same teardown slot. It removes the three `_callback_*` attributes, the two private methods, and the now-unused `import queue`; it keeps `self._on_write` for the separate `_on_write.close()` resource teardown.

## Invariants preserved (verbatim move)

`on_write=None` → `fire` no-op; lazy + idempotent single worker (start-race lock); FIFO; a raising callback is logged (`exc_info=True`) and the worker continues; `close()` drains then joins (≤30 s), warns on timeout, is safe when never started and on a second call; daemon thread named `write-callback`.

## Tests

New `tests/test_write_callback.py` unit-pins all six invariants directly (deterministic — `threading.Event` + `close()`-join, no sleeps). The pre-existing Collection-facade callback tests (`test_collection.py`, `test_thread_safety.py`, `test_managers_document.py`, `test_git.py`) remain green as integration coverage. Full suite: 1903 passed, 1 skipped, 1 xpassed. `write_callback.py` patch coverage 95% (the one uncovered branch is a provably-dead defensive guard, carried over verbatim).

## Deferred

Latent thread-contract/lifecycle hardening for the dispatcher (close/fire lock discipline, use-after-close guard, removing the dead guard, quantifying the timeout warning) is **pre-existing** (carried verbatim from `Collection`) and tracked in #601 — out of scope for this behaviour-preserving move, per the #582 verbatim-move contract.

## Docs

`CLAUDE.md` and `docs/design.md` module maps gain `write_callback.py`. Internal-only change — no README / tools / config / resource surface.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -29,6 +29,7 @@ src/markdown_vault_mcp/
   providers.py         -- embedding provider ABC + implementations
   tracker.py           -- hash-based change detection
   collection.py        -- thin facade: lifecycle, wiring, delegation (index-write → indexing/coordinator.py)
+  write_callback.py    -- WriteCallbackDispatcher: deferred git-commit callback worker (#599)
   config.py            -- configuration loading
   server.py            -- generic FastMCP server factory (make_server) with tool annotations
   cli.py               -- CLI entry point
--- a/docs/design.md
+++ b/docs/design.md
@@ -65,6 +65,7 @@ markdown-vault-mcp (new package)
 +-- providers.py      -- Ollama / OpenAI / SentenceTransformers
 +-- tracker.py        -- hash-based change detection
 +-- collection.py     -- thin facade: lifecycle, wiring, delegation (index-write → indexing/coordinator.py)
++-- write_callback.py -- WriteCallbackDispatcher: deferred git-commit callback worker (#599)
 +-- config.py         -- configuration loading
 +-- server.py         -- generic FastMCP server
 +-- cli.py            -- CLI entry point
--- a/src/markdown_vault_mcp/collection.py
+++ b/src/markdown_vault_mcp/collection.py
@@ -8,7 +8,6 @@ from __future__ import annotations
 
 import contextlib
 import logging
-import queue
 import re
 import threading
 from typing import TYPE_CHECKING, Any, Literal
@@ -41,10 +40,10 @@ from markdown_vault_mcp.types import (
     ReindexResult,
     RenameResult,
     WriteCallback,
-    WriteOperation,
     WriteResult,
 )
 from markdown_vault_mcp.utils import effective_attachment_extensions
+from markdown_vault_mcp.write_callback import WriteCallbackDispatcher
 
 if TYPE_CHECKING:
     from collections.abc import Iterator
@@ -312,6 +311,12 @@ class Collection:
             snippet_words=snippet_words,
             length_downweight_alpha=length_downweight_alpha,
         )
+        # Deferred write callback (issue #175): the git-commit on_write
+        # callback runs on a background worker so write methods return after
+        # the FTS update.  Constructed before DocumentManager, whose
+        # ``on_write_callback`` is wired to ``fire`` (#599).
+        self._write_callback = WriteCallbackDispatcher(self._on_write)
+
         # 4. DocumentManager (mark_paths_dirty routes through the writer)
         self._doc_mgr = DocumentManager(
             fts=self._fts,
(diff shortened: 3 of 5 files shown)


pvliesdonk:
Thanks both. Consolidated response to the review.

**Gemini's 4 HIGH findings** (`_closed` flag, lock `fire()`, de-lock `_ensure_worker`, join-outside-lock in `close()`) — **declining for this PR, tracked in #601.** I verified each against the pre-extraction `Collection` worker: the concurrency shape is moved **verbatim** — the original had no `_closed` flag, no lock around `_fire_write_callback`, `_ensure_callback_worker` acquiring the lock internally, and `close()` reading `_callback_worker` unlocked. So these are **pre-existing latent hazards, not introduced or worsened by this extraction** (reachability is unchanged — same callers, same threading model). This PR is a behaviour-preserving move under the #582 verbatim-move contract, so hardening them here would be an out-of-scope behaviour change.

They are real and worth doing — which is why they're filed as **#601**. They belong in a dedicated pass with proper failure-mode enumeration and TDD: adding a lock to `fire()` puts a lock acquisition on the hot write path (contention tradeoff), a `_closed` guard interacts with the existing `is_alive()` checks, and join-outside-lock has its own ordering considerations against `Collection`'s other locks. Those decisions deserve their own reviewed change, not a reactive bolt-on to an extraction.

**claude-review** — thanks; agreed on all three deferred items (already in #601). The test-helper type nit (#4: `object` → `WriteCallback`, `operation: str` → `WriteOperation`) is real but cosmetic — `tests/` is currently in mypy `ignore_errors` (open issue #425), so it isn't a gate finding. I'll fold it into #601, which already touches this module's tests.

Net: PR #602 stays a clean verbatim extraction; the dispatcher's thread-contract hardening (the substance of Gemini's findings) is tracked and scoped in #601.

PR #613, opened 2026-06-03
refactor(config): group CollectionConfig fields into config_sections sub-configs

pvliesdonk:
Refs #579 (PR B of 3). PR A (#611) removed the dead auth duplicates; this groups the remaining domain fields into composed sub-configs. PR C (from_env + frozen, now unblocked by pvl-core v3.2.0) follows.

## What
`CollectionConfig`'s ~37 flat domain fields move into six composed sub-config dataclasses in a new `config_sections/` package: `GitConfig`, `IndexingConfig`, `EmbeddingsConfig`, `SearchConfig`, `SyncConfig`, `ContentConfig`. `source_dir`, `read_only`, `server_name`, `instructions`, and `server: ServerConfig` stay top-level.

`load_config()` stays the entrypoint and builds the sub-configs from the same env reads; `to_collection_kwargs()` + the src consumers (`providers.py`, `_server_deps.py`, `_server_resources.py`, `server.py`) + tests read `config.<group>.<field>`.

## Naming / shape decisions
- **`source_dir`/`read_only` stay top-level** (vault identity/mode) so the common `CollectionConfig(source_dir=...)` construction stays clean. Refines spec §6, which had grouped them into a `VaultConfig`.
- **Dropped redundant whole-group prefixes** (`git_token`→`config.git.token`, `embedding_provider`→`config.embeddings.provider`); **kept meaningful sub-prefixes** (`config.embeddings.ollama_host`, `config.sync.file_watcher_enabled`).
- The `ollama_host` non-empty/no-trailing-slash normalization moved from `CollectionConfig.__post_init__` to `EmbeddingsConfig.__post_init__`.

## No behaviour change
No env-var renames; every `MARKDOWN_VAULT_MCP_*` var is read exactly as before, just landing on a sub-config. Defaults preserved verbatim.

## Scope decision (pre-declared)
Field **validation** (e.g. `SearchConfig`'s `chunks_per_file >= 1`) stays in `load_config()` for now — that's the production path and behaviour is unchanged. Moving validation into the sub-configs' `__post_init__`/`from_env` is deferred to **PR C** (the `from_env`+frozen convergence), where env-reading itself moves into the sub-configs. Doing it here would be a behaviour change (direct construction would start raising); PR B is a pure no-behaviour-change relocation.

## Verification
1829 passed / 1 skipped / 1 xpassed; mypy + ruff clean (106 files). Full five-lens preflight + type-design/pr-test/code/comment analyzers run before this first push — clean at the ≥80 bar; sub-80 items (a stale `server.py` comment, `EmbeddingsConfig.__post_init__` direct-construction tests, `docs/design.md` structure entry) closed pre-push. config.py 808→712 LOC (the <400 target lands with PR C retiring the imperative `load_config`).

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -67,6 +67,7 @@ markdown-vault-mcp (new package)
 +-- collection.py     -- thin facade: lifecycle, wiring, delegation (index-write → indexing/coordinator.py)
 +-- write_callback.py -- WriteCallbackDispatcher: deferred git-commit callback worker (#599)
 +-- config.py         -- configuration loading
++-- config_sections/  -- domain-grouped sub-configs (git/indexing/embeddings/search/sync/content)
 +-- server.py         -- generic FastMCP server
 +-- cli.py            -- CLI entry point
 +-- utils/
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -124,6 +124,9 @@ ignore = ["E501"]
 "src/markdown_vault_mcp/_server_resources.py" = ["B008", "TC001", "TC002"]
 "src/markdown_vault_mcp/_server_prompts.py" = ["B008", "TC002"]
 "src/markdown_vault_mcp/_server_apps.py" = ["B008", "TC001", "TC002"]
+# IndexingConfig needs pathlib.Path at runtime so typing.get_type_hints() works
+# on the public dataclass; keep the import top-level, not behind TYPE_CHECKING.
+"src/markdown_vault_mcp/config_sections/indexing.py" = ["TC003"]
 # Test fixtures import collection/protocol types only used in annotations.
 "tests/conftest.py" = ["TC002", "TC003"]
 "tests/test_smoke.py" = ["TC002", "TC003"]
--- a/src/markdown_vault_mcp/_server_deps.py
+++ b/src/markdown_vault_mcp/_server_deps.py
@@ -126,16 +126,16 @@ def make_collection_lifespan(config: CollectionConfig) -> Any:
         )
         from markdown_vault_mcp.exceptions import IndexUnavailableError
 
-        # Use the *resolved* pull interval from kwargs, not config.git_pull_interval_s:
+        # Use the *resolved* pull interval from kwargs, not config.git.pull_interval_s:
         # the config default is 600 even on non-git vaults, but to_collection_kwargs()
         # only passes a non-zero interval through when a git strategy is configured.
         git_pull_active = kwargs.get("git_pull_interval_s", 0) > 0
 
         file_watcher = None
         if should_start_file_watcher(
-            config.file_watcher_enabled,
+            config.sync.file_watcher_enabled,
             git_pull_active,
-            config.github_webhook_secret,
+            config.sync.github_webhook_secret,
         ):
 
             def _on_file_change() -> None:
@@ -152,10 +152,10 @@ def make_collection_lifespan(config: CollectionConfig) -> Any:
             file_watcher = VaultFileWatcher(
                 config.source_dir,
                 _on_file_change,
-                debounce_s=config.file_watcher_debounce_s,
+                debounce_s=config.sync.file_watcher_debounce_s,
             )
             file_watcher.start()
-        elif not config.file_watcher_enabled:
+        elif not config.sync.file_watcher_enabled:
             logger.debug("file_watcher: disabled via FILE_WATCHER=false")
         else:
             logger.info(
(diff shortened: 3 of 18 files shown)


pvliesdonk:
Dispositions on the review findings (commit ed0231a):

**Fixed (in this diff's scope):**
- **#1 (`IndexingConfig` Path behind `TYPE_CHECKING`):** real latent fragility — imported `Path` at runtime so `typing.get_type_hints(IndexingConfig)` works on the public dataclass, with a per-file `TC003` ignore matching the project's conftest/test_smoke pattern. Verified `get_type_hints(IndexingConfig)` now resolves.
- **#4 (stale `to_collection_kwargs` docstring):** `embeddings_path` → `indexing.embeddings_path`.
- **#5 (cross-section coupling):** added a comment on the gate noting semantic search is enabled by `config.indexing.embeddings_path` while the provider lives in `config.embeddings`.

**Pre-existing — not introduced by this PR (a no-behaviour-change field relocation), so out of scope here:**
- **#2 (redundant `.rstrip('/')` in `providers.py`):** already redundant before this PR — the old `CollectionConfig.__post_init__` stripped `ollama_host` too. Harmless dead code; leaving it to keep this PR a pure relocation.
- **#3 (three near-identical `GitWriteStrategy` branches):** the duplication predates this PR (it only changed the field paths within the branches). Filed as its own behaviour-neutral cleanup: see the new `refactor: dedup the three GitWriteStrategy construction branches` issue.

1829 tests green; mypy/ruff clean.

pvliesdonk:
Reconsidered the round-2 findings — I was wrong to defer #2/#3 as "pre-existing"; they're behaviour-neutral cleanups in code this PR already touches, so they belong here. Both addressed in commit acaabab:

- **#3 (three `GitWriteStrategy` blocks):** extracted a typed `_build_git_strategy()` helper applying the 8 shared kwargs; each branch passes only its distinct `token`/`repo_url`/`managed`/`enable_*`. (Used a typed helper rather than the literal `**shared` dict — that would fail mypy on the mixed value types.) Behaviour-identical; `Closes #616`.
- **#2 (`providers.py` ollama_host):** dropped the redundant `.rstrip('/')` and use the single `host` variable for both the probe and `OllamaProvider`, eliminating the split reference.

- **#5 (fail-fast validation):** the *comment* mitigation stays; the suggested validation that **raises** when exactly one of `indexing.embeddings_path` / `embeddings.provider` is set is a behaviour change (it would turn a silently-degrading misconfig into a startup crash for existing deployments) — deferred to PR C, where env-reading + validation move into the sub-configs.

1829 tests green (unchanged → behaviour-neutral); mypy/ruff clean.

pvliesdonk:
Round-3 disposition + full local pre-flight review (10 lenses) on `587a662`.

**Finding #1 (CONFIRMED) — the dedup helper re-introduced the `get_type_hints` landmine.** Fixed: `GitWriteStrategy` is now imported at module level (it is *constructed* at runtime in `_build_git_strategy`, and `git.py` has no back-import of `config.py`, so there's no circularity and no cycle). Verified the whole class of defect is gone, not just this instance — a `get_type_hints` sweep over `CollectionConfig` + all six sub-configs + every method now resolves cleanly. My first pass at this attached a code comment citing `get_type_hints` as the rationale; a local review caught that as **inaccurate** (`get_type_hints` is never actually called in the tree — the operative reason is the runtime construction, and the comment could mislead someone into moving the import back under `TYPE_CHECKING` and breaking the call). The comment is removed; the import is self-explanatory.

**Finding #2 (default-literal drift) — deferred to PR C.** Pre-existing; the clean fix (env-reading referencing the dataclass defaults) lands when `from_env()` removes the `load_config()` literals entirely.

**Finding #3 (`EmbeddingsConfig` normalizes `ollama_host` but not `openai_base_url`) — defended; revisited in PR C.** The asymmetry is *functional*, not an oversight: `ollama_host` is interpolated into the auto-detect probe URL built directly at `providers.py:462` (`f\"{host}/api/tags\"` — a trailing slash yields `//api/tags`), so it must be stripped; `openai_base_url` is handed to the OpenAI SDK client, which normalizes it. When PR C moves normalization into the sub-configs it'll be made symmetric.

Other items the local review surfaced and their disposition:
- **`templates_folder` invariant enforced only in `load_config`, not the type** — same category as #3, deferred to PR C (normalization-in-subconfig); existing `test_templates_folder_*` pin the invariant meanwhile.
- **`to_collection_kwargs` re-flattens into `Collection(**kwargs)`** — by design; `Collection.__init__`'s flat signature is out of scope for a config-only restructure.
- **`test_includes_all_collection_params` keeps its `xfail`** — this is the conscious deferral from #596 (tracked by #595 for the HuggingFace download flakiness); PR B only mechanically migrates the field access, so re-working the test's network strategy is out of scope here.

1829 tests pass, mypy/ruff clean, behaviour-neutral throughout.

PR #615, opened 2026-06-03
refactor(collection): extract reader/writer/graph/index facets (#604)

pvliesdonk:
## Summary

PR3a of the `collection.py` facade-decomposition epic (#576). Introduces a `facets/` package that groups the formerly-flat `Collection` surface into four cohesive, thin delegating views over the existing managers + coordinator:

| Facet | Surface | Collaborators |
|-------|---------|--------------|
| `ReaderFacet` | search, read, list, folders, tags, toc, recent, similar, context, stats, embeddings_status, history, diff, read_attachment | `SearchManager`, `DocumentManager`, `IndexManager`, `GitQueryManager`, `require_built` |
| `WriterFacet` | write, edit, delete, rename, write_attachment | `DocumentManager` |
| `GraphFacet` | backlinks, outlinks, broken_links, orphans, most_linked, connection_path | `LinkManager`, `require_built` |
| `IndexFacet` | build/reindex/embeddings (sync + async), readiness, writer status | `IndexWriteCoordinator` (public subset only) |

`Collection` constructs the facets once in `__init__` and exposes them via `reader` / `writer` / `graph` / `index` properties; the flat methods now delegate to them (**addition before removal**). The bucket-3 readiness gate (`require_built`) moved inside the facets, so it is expressed in exactly one place per method.

This is a deliberately **behaviour-preserving** extraction — every flat method routes 1:1 to the matching facet, which delegates 1:1 to the same manager/coordinator as before. Signatures, kwarg forwarding, and the readiness-gate placement are unchanged.

## Why ReaderFacet is thin

The discarded first attempt (#608) produced a 12-parameter `ReaderFacet` because `stats()` computed FTS aggregates inline and `get_history`/`get_diff` held git logic directly. PR-A (#612) first moved `stats()` into `SearchManager` and extracted `GitQueryManager`, so `ReaderFacet` now lands at **5 collaborators** with no duplicated config state. (Manager-boundary refactor before facet extraction — the sequencing lesson from #608.)

## Note on docstrings (deviation from #604's text)

[...]

PR #619, opened 2026-06-04
feat(graph): expose limit on get_backlinks / get_outlinks (closes #617)

### LINKED ISSUE #617: feat: expose LinkManager's `limit` on get_backlinks / get_outlinks (currently unreachable)
## Summary

`LinkManager.get_backlinks` and `get_outlinks` both accept a `limit: int | None = None` keyword-only cap, but the public surface (`Collection.get_backlinks` / `get_outlinks`, and now `GraphFacet`) takes only `path`, so the cap is unreachable from any public caller. This is inconsistent with the sibling `get_most_linked`, which *does* surface `limit` on all three layers.

## Observed

`[verified: read managers/link.py + collection.py + facets/graph.py at 8d36b4e]`

- `managers/link.py`: `get_backlinks(self, path, *, limit: int | None = None)`, `get_outlinks(self, path, *, limit: int | None = None)`, `get_most_linked(self, *, limit: int = 10)`.
- `Collection.get_backlinks(self, path)` / `get_outlinks(self, path)` — no `limit`; delegate to `GraphFacet`, which also omits it.
- `Collection.get_most_linked` / `GraphFacet.get_most_linked` — *do* forward `limit`.

So `limit` is supported by the manager and exposed for `get_most_linked`, but silently dropped for `get_backlinks` / `get_outlinks` — an inconsistency, most likely an oversight rather than deliberate curation.

## Expected

`limit` reachable on `get_backlinks` / `get_outlinks` through the public surface, consistent with `get_most_linked`.

## Scope boundary

Deliberately **not** folded into PR #615 (#604): that PR is a behavior-preserving facet extraction, and adding a parameter *widens* the public API (an addition, not preservation). The facets there faithfully mirror Collection's current surface. This is the separate "expose the param" feature.

## Deliverable

Add `limit: int | None = None` (keyword-only) to `get_backlinks` / `get_outlinks` on `GraphFacet` and `Collection`, forward to `LinkManager`, document the `Args: limit:` entry, and add a test exercising the cap. Mirror the `get_most_linked` pattern.

Surfaced by claude-review on #615. Refs #576.

pvliesdonk:
## Summary

`LinkManager.get_backlinks` / `get_outlinks` have accepted a keyword-only `limit: int | None` cap since #210 (applied as a SQL `LIMIT` in `FTSIndex`), but it was **unreachable from any public surface** — `Collection`, `GraphFacet`, and the `get_backlinks` / `get_outlinks` MCP tools all dropped it. This was inconsistent with `get_most_linked`, which exposes `limit` end-to-end, and was filed as #617 during the #615 review.

This threads `limit` through the whole stack:

- **`GraphFacet`** and **`Collection`** — keyword-only `limit: int | None = None`, forwarded to `LinkManager`.
- **MCP `get_backlinks` / `get_outlinks` tools** — `limit` exposed (the primary interface), forwarded via `asyncio.to_thread(..., limit=limit)`.

**Non-breaking:** `limit` is keyword-only at the library layers and defaults to `None` (unlimited) everywhere, so existing callers are unaffected. Adding an optional param to the MCP tools is a non-breaking wire-shape change.

## Contract

`None` (the default) returns all results. A positive `N` caps to `N`. The value forwards verbatim to SQLite's `LIMIT`, so `limit=0` returns none and a negative `limit` returns all (SQLite treats a negative `LIMIT` as unbounded) — this matches every other unvalidated `limit` param in the codebase (`search`, `get_most_linked`, `get_recent`). The `limit=0` / negative boundary is pinned by a test so the contract is explicit.

## Tests

- `tests/test_facet_graph.py::TestGraphFacetLimit` — caps 2→1 at the facet, `limit=None` returns all.
- `tests/test_links.py` — Collection-level cap tests + `test_get_backlinks_limit_boundary_values` (`limit=0` → `[]`, `limit=-1` → all).
- `tests/test_server.py::TestLinkTools::test_link_tools_accept_limit` — MCP `get_outlinks` caps 2→1; `get_backlinks` accepts + forwards `limit`.

## Docs

`docs/tools/index.md` (both tool parameter tables) and `docs/design.md` (the `LibraryFacade` Protocol signatures) updated for the new param.

## Verification

Full suite **1883 passing**, mypy / ruff clean, diff-cover **100%** on changed source. Pre-flight review (CLAUDE.md, diff, history, prior-PR, comments, code-review, test-coverage, comment-accuracy lenses) clean.

Closes #617. Refs #576.


pvliesdonk:
Thanks for the review. On the three observations:

**#2 (no outlinks boundary test) — fixed in `ca7fbf2`.** You're right; the #210 reviewer convention asks for symmetric backlinks/outlinks coverage. Added `test_get_outlinks_limit_boundary_values` mirroring the backlinks one (`limit=0` → `[]`, `limit=-1` → all).

**#3 (`docs/design.md` LibraryFacade Protocol) — already done.** The Protocol section (`docs/design.md:1256-1257`) is the single `Collection`/library signature list, and both `get_backlinks`/`get_outlinks` lines there now carry `*, limit: int | None = None`. (The diff you saw was likely partial.)

**#1 (MCP `get_backlinks` test is acceptance-only) — left as-is, intentionally.** The `_mcp_env_linked` vault has no 2-backlink document, and adding one would break the sibling `test_get_backlinks` assertion (it pins `notes/topic.md` at exactly 1 backlink). Capping for backlinks is already pinned at the Collection level (`test_get_backlinks_limit_caps_results` + the boundary test), the two tool bodies forward identically, and the `get_outlinks` tool path exercises real capping (2→1) at the same layer — so a dropped-`limit` regression on the backlinks tool would still be caught. Not worth a separate single-purpose server fixture for a transitively-covered path.

Full suite 1884 passing; local review pass clean.

PR #621, opened 2026-06-04
docs: correct inherited Raises/Returns docstring gaps (closes #614)

### LINKED ISSUE #614: docs: write/read docstrings omit some Raises entries and overstate section-match strictness (Collection + facets)
## Summary

Several `Collection` write/read method docstrings (and their verbatim copies in the new `facets/` package, #604) under-document the `Raises:` contract of the `DocumentManager` methods they delegate to, and one overstates section-match strictness. The docstrings are not *wrong* (the code does raise the listed exceptions), but they omit catchable failure modes a caller would want to handle.

## Observed

Surfaced by the PR3a (#604) pre-flight `comment-analyzer` lens, which cross-checked each `Raises:` block against the delegated manager implementation:

- `edit` — docstring lists `EditConflictError` / `ReadOnlyError` / `ConcurrentModificationError` / `ValueError` but omits **`DocumentNotFoundError`**, which `DocumentManager.edit` raises (and documents) when the target file does not exist. `[verified: DocumentManager.edit raises DocumentNotFoundError]`
- `rename` — docstring omits **`DocumentExistsError`**, which `DocumentManager.rename` raises (and documents) when `new_path` already exists. `[verified: DocumentManager.rename raises DocumentExistsError]`
- `write` / `write_attachment` — the `ConcurrentModificationError` clause says "if *if_match* is provided and does not match the current file hash" but the method also raises it when `if_match` is supplied for a file that **does not yet exist**; `DocumentManager.write`'s own docstring captures this ("or the file does not exist") while the copy drops it. `[verified: DocumentManager.write/write_attachment raise ConcurrentModificationError for if_match on a missing file]`
- `read` — the `section` arg says the heading "matches *section* exactly (case-sensitive)", but `FTSIndex.get_section` normalises (collapses internal whitespace runs, strips) before comparing, so `"1.3.  Reducing"` matches a stored `"1.3. Reducing"`. Case-sensitivity itself is correct; "exactly" overstates. `[verified: FTSIndex section lookup normalises whitespace before comparison]`

All four are **pre-existing** in the flat `Collection` docstrings and were faithfully copied verbatim into the facet methods by the behaviour-and-documentation-preserving extraction in #604.

## Expected

Each `Raises:` block lists every exception the delegated method can raise; the `read` `section` description reflects the actual (whitespace-normalised) matching.

## Context

- Discovered during the #604 pre-flight review (facet extraction). The facets are slated to become the docstrings' permanent home once the flat methods are removed (PR4), so the canonical copy should be corrected.
- Deliverable: fix the `Raises:`/section text in **both** `Collection` and the `facets/` copies in one pass so they stay in sync (the facet docstrings currently mirror `Collection` verbatim). Could fold into PR3c/PR4 of #576.

## Scope boundary

Documentation only — no behaviour change. Not about adding new exceptions or changing what the methods raise; only about documenting the existing contract accurately.

Refs #576, #604.

pvliesdonk:
## Summary

The facet docstrings introduced in #604 were verbatim copies of the flat `Collection` docstrings, which carried several small pre-existing completeness gaps. These surfaced one-at-a-time across the #604 / #615 / #619 review rounds, so #614 was filed to own the **whole class** and fix it in a single cross-layer pass. This is that pass.

Each gap is corrected at the **source** (manager / coordinator) *and* every verbatim copy (`Collection`, the relevant facet, and — where applicable — the MCP `read` tool) so the layers stay in sync:

| Method | Correction |
|--------|-----------|
| `edit` | document `DocumentNotFoundError` (target file missing) |
| `rename` | document `DocumentExistsError` (`new_path` already exists) |
| `write` / `write_attachment` | `ConcurrentModificationError` also fires when `if_match` is supplied for a not-yet-existing file (and `Collection.write_attachment` gains the `Returns`/`Raises` it was missing entirely) |
| `read` | section match is "case-sensitive; internal whitespace is collapsed before comparison" — not "exactly" (fixed at manager, `Collection`, `ReaderFacet`, **and the MCP `read` tool**) |
| `get_diff` | `Returns` empty when the vault is not a git repo (mirrors `get_history`) |
| `wait_for_drain` | one-liner now documents the `bool` return (`True` = drained, `False` = timeout) |

## Behaviour

**None changed.** Every documented exception/return was traced to its implementation and verified accurate; the `"exactly"` wording was stale ever since #497 (May 15) added whitespace-tolerant section matching without updating the upstream docstrings.

## Tests

Documentation-only, with one exception: the `write_attachment` `if_match`-on-missing-file branch was the one documented behaviour with **no test** (the `write` `.md` equivalent was covered, but `write_attachment` routes through its own branch). Added `test_write_attachment_with_if_match_on_nonexistent_raises` to pin it (asserts `ConcurrentModificationError` with `actual == "(file does not exist)"`).

## Verification

Full suite **1887 passing**, mypy + ruff clean. Pre-flight review (CLAUDE.md / diff / history / prior-PR / comments / code-review / comment-accuracy / test-coverage lenses) clean over two rounds — round 1 caught a missed lockstep copy (the MCP `read` tool still said "exactly") and the untested `write_attachment` branch, both fixed here.

Closes #614. Refs #576, #604.

### DIFF
--- a/src/markdown_vault_mcp/_server_tools.py
+++ b/src/markdown_vault_mcp/_server_tools.py
@@ -424,8 +424,9 @@ def register_tools(mcp: FastMCP, *, transport: str = "stdio") -> None:
                 (e.g. "Journal/note.md" or "assets/diagram.pdf").
                 Case-sensitive.
             section: When provided, return only the section whose heading
-                matches *section* exactly (case-sensitive). Pass the ``heading``
-                value from a ``search`` result unchanged for guaranteed match.
+                matches *section* (case-sensitive; internal whitespace is
+                collapsed before comparison). Pass the ``heading`` value
+                from a ``search`` result unchanged for guaranteed match.
                 ``None`` (the default) returns the whole document.
                 Ignored for non-.md paths.
 
@@ -1545,7 +1546,8 @@ def register_tools(mcp: FastMCP, *, transport: str = "stdio") -> None:
             ValueError: If content_base64 is missing/invalid for
                 attachments, or the content exceeds the size limit.
             McpError: If if_match is provided and the file has been
-                modified (ConcurrentModificationError).
+                modified, or if_match is supplied for a file that does not
+                yet exist (ConcurrentModificationError).
         """
         if not path.endswith(".md"):
             if not content_base64:
@@ -1627,6 +1629,9 @@ def register_tools(mcp: FastMCP, *, transport: str = "stdio") -> None:
                 numbers are out of range.
             EditConflictError: If old_text is not found or appears more
                 than once.
+            DocumentNotFoundError: If no file exists at the given path.
+            McpError: If if_match is provided and the file has been modified
+                (ConcurrentModificationError).
         """
         try:
             result = await asyncio.to_thread(
--- a/src/markdown_vault_mcp/collection.py
+++ b/src/markdown_vault_mcp/collection.py
@@ -544,7 +544,7 @@ class Collection:
         return self._index_facet.write_generation()
 
     def wait_for_drain(self, timeout: float | None = None) -> bool:
-        """Block until :meth:`is_drained`, or until *timeout* (best-effort)."""
+        """Block until :meth:`is_drained`; ``True`` if drained, ``False`` on timeout."""
         return self._index_facet.wait_for_drain(timeout)
 
     def get_index_status(self) -> dict[str, Any]:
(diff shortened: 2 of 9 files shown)


pvliesdonk:
Thanks — all three addressed in `7de2f0a`:

**Finding 1 (edit MCP tool Raises) — fixed.** You're right that this is the same cross-layer gap #614 set out to close; I'd scoped the MCP layer too narrowly. The `edit` tool now lists `DocumentNotFoundError` and the `McpError`/`ConcurrentModificationError` case, matching the already-complete `rename`/`delete` templates. One nuance: I deliberately scoped the edit tool's CME line to "the file has been modified" only (no missing-file sub-case) — in `edit`, the `DocumentNotFoundError` check (`document.py:708-709`) runs *before* the `if_match` check (`:711`), so CME-on-missing is unreachable there; `DocumentNotFoundError` covers it.

**Finding 2 (write MCP tool) — fixed.** The `write` tool's CME line now includes "or if_match is supplied for a file that does not yet exist" (where it *is* reachable, on both the `.md` and attachment paths).

**Finding 3 (wait_for_drain "Polls every 50ms") — keeping it dropped.** The poll interval is an implementation detail (still `poll_interval = 0.05` in the code), not part of the public contract — and re-adding it only to the coordinator would re-diverge the three layers I just synced, while adding it everywhere would push the one-liner past the 88-char limit. The docstring now documents the actual *contract* (the `bool` return), which was the real gap. Happy to revisit if you feel the timeout-precision hint is worth a multi-line docstring.

Re-ran the full local review pass (7 lenses) after the change; clean. Full suite 1887 passing.

PR #631, opened 2026-06-05
test: close Vault-owning fixtures via yield/finally

### LINKED ISSUE #625: test: several Collection fixtures return without closing (teardown-hygiene gap)
## Summary

Several test fixtures/helpers construct a `Collection` and `return` it without ever calling `col.close()`, leaking the SQLite handle (and any background threads) past the test, instead of the `yield col` / `finally: col.close()` teardown the suite's other Collection fixtures use.

## Observed

In `tests/conftest.py`, the `built` fixture (added in #608) closes correctly:

```python
@pytest.fixture
def built(vault_path: Path):
    ...
    try:
        yield col
    finally:
        col.close()
```

but the sibling `populated_collection` fixture (used by ~12 tests) returns without closing:

```python
@pytest.fixture
def populated_collection(tmp_path: Path):
    ...
    col.index.build_index()
    col.index.build_embeddings()
    return col            # <- no yield / no col.close()
```

`git blame` dates `populated_collection`'s `return col` to `ef019d6` (`feat(search)!: field-collapse search results into GroupedResult`, 2026-05-12) — it predates the facade-decomposition epic and is unrelated to it.

A grep shows it is not isolated; multiple fixtures/helpers return a `Collection` (or a tuple containing one) without closing:

```
tests/conftest.py:103
tests/test_github_webhook.py:78
tests/test_background_fts.py:1029
tests/test_git.py:2939, 2981, 3077, 3175
tests/test_collection.py:107, 767, 785, 2473, 3095, 3525
```

## Expected

Each fixture that owns a `Collection`'s lifetime should release it via `yield col` + `finally: col.close()`, matching the convention prior reviewers explicitly required on the sibling fixtures during #608/#615. (Some of the sites above may be helper functions where the *caller* owns cleanup rather than pytest fixtures — each needs checking individually. `[unverified]` per site.)

## Context

Surfaced by the prior-PR-review lens during the PR3c (#606) preflight circus. The leak is **pre-existing** and out of PR3c's "pure call-site rename, no behaviour change" scope, so it was deliberately not folded in. Low severity (tests pass; pytest `tmp_path` cleanup + GC mask most of it), but the inconsistency is a latent cross-test-contamination / locked-DB flake risk. I have not verified any concrete flake stems from it.

## Scope boundary

Not about the facade decomposition (#576) — purely test-fixture teardown hygiene.

## Verification

```bash
# After the fix, no fixture should hand out a Collection it doesn't close:
grep -rnE "return (col|collection|coll|writable)\b" tests/   # audit each remaining hit
```

## Deliverable

- Convert the genuine Collection-owning fixtures to `yield` + `finally: col.close()`.
- Leave helper functions whose caller owns cleanup as-is (document which, and why, in the PR).

pvliesdonk:
Closes #625. Refs #576.

## Problem

Eight `Vault`-owning pytest fixtures built a `Vault` (opening a SQLite handle + spawning the writer thread) and `return`ed it without ever closing it. Each leaked its handle and background thread for the lifetime of the test session — inconsistent with the sibling `built` fixture, which already used the `yield`/`finally: close()` teardown pattern.

## Fix

Convert the eight leaking fixtures from `return X` to `try: yield X / finally: X.close()`, so the SQLite handle and writer thread are released at fixture teardown even when a test raises:

- `populated_vault` (`tests/conftest.py`)
- `corpus_vault` (`tests/test_api_validation.py`) — yields the `(vault, stats)` tuple, closes `vault` in `finally`
- `vault`, `writable`, `writable_with_embeddings`, `semantic_vault`, `writable_semantic`, `vault_with_long_doc` (`tests/test_vault.py`)

Each `-> Vault` annotation widens to `-> Iterator[Vault]` (generator fixtures); `Iterator` is imported under `TYPE_CHECKING`. `corpus_vault`'s docstring switches `Returns:` → `Yields:` to match.

## Scope

Per the issue, the helper functions (`_make_vault*`, `_ready_vault`) are intentionally left as-is — their callers own the returned `Vault`'s lifetime. Tightening those callers is a lower-severity follow-up.

## Verification

- `uv run ruff check .` / `ruff format --check` — clean
- `uv run pytest` — 1852 passed
- `populated_vault` consumers pass under `-W error::ResourceWarning`, confirming the close releases the handle rather than masking it

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -98,9 +98,12 @@ def populated_vault(tmp_path: Path):
         embedding_provider=MockEmbeddingProvider(),
         embeddings_path=tmp_path / "vectors",
     )
-    col.index.build_index()
-    col.index.build_embeddings()
-    return col
+    try:
+        col.index.build_index()
+        col.index.build_embeddings()
+        yield col
+    finally:
+        col.close()
 
 
 @pytest.fixture
--- a/tests/test_api_validation.py
+++ b/tests/test_api_validation.py
@@ -13,6 +13,7 @@ import pytest
 from markdown_vault_mcp.vault import Vault
 
 if TYPE_CHECKING:
+    from collections.abc import Iterator
     from pathlib import Path
 
     from markdown_vault_mcp.types import IndexStats
@@ -118,10 +119,10 @@ def corpus_path(tmp_path: Path) -> Path:
 
 
 @pytest.fixture
-def corpus_vault(corpus_path: Path) -> tuple[Vault, IndexStats]:
-    """Return a built Vault configured with ifcraftcorpus settings.
+def corpus_vault(corpus_path: Path) -> Iterator[tuple[Vault, IndexStats]]:
+    """A built Vault configured with ifcraftcorpus settings.
 
-    Returns:
+    Yields:
         Tuple of (vault, index_stats) so tests can inspect both.
     """
     vault = Vault(
@@ -130,8 +131,11 @@ def corpus_vault(corpus_path: Path) -> tuple[Vault, IndexStats]:
         indexed_frontmatter_fields=["cluster", "topics"],
         read_only=True,
     )
-    stats = vault.index.build_index()
-    return vault, stats
+    try:
+        stats = vault.index.build_index()
+        yield vault, stats
+    finally:
+        vault.close()
 
 
 # ---------------------------------------------------------------------------
--- a/tests/test_vault.py
+++ b/tests/test_vault.py
@@ -43,6 +43,8 @@ from markdown_vault_mcp.vault import Vault
 from tests.conftest import wait_for_writer_drain
 
 if TYPE_CHECKING:
+    from collections.abc import Iterator
+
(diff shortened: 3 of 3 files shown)


pvliesdonk:
Addressed the Gemini setup-leak finding: moved the fixture setup calls (`build_index()` / `build_embeddings()` / `_load_vectors()`) **inside** the `try` block across all 8 converted fixtures, so a failure during setup still hits `finally: vault.close()` rather than leaking the SQLite handle + writer thread. The constructor stays outside the `try` (it can't leak before the object exists, and `Vault.close()` is partial-init-safe via `hasattr` guards). For `corpus_vault`, `stats = vault.index.build_index()` now runs inside the `try` and is still yielded as part of the `(vault, stats)` tuple. Full suite green (1852 passed). The pre-existing `built` fixture is intentionally left untouched here — it's outside #625's return→yield scope and is hardened by #618 (PR #632) to avoid a merge conflict on the same lines.

PR #656, opened 2026-06-10
feat(config): explicit embedding provider fails fast instead of silent degrade (#638 PR2)

pvliesdonk:
## Summary

PR2 of #638. Removes a **silent failure**: previously, when an embedding provider couldn't be loaded, `to_vault_kwargs` always logged a warning and disabled semantic search — even when the operator had *explicitly* chosen a backend via `MARKDOWN_VAULT_MCP_EMBEDDING_PROVIDER`. A misconfigured deployment silently became keyword-only search with no signal.

**Refs #638** — the issue closes with PR3 (`list`→`tuple`, + #639).

### Behaviour
- **Explicit provider** (`EMBEDDING_PROVIDER` set) that can't be constructed at startup — missing dependency, missing/empty credentials, or an unrecognised value → **fail fast** with `ConfigurationError`.
- **Auto-detect** (variable unset) with no backend available → unchanged: warn + continue with semantic search disabled (the zero-config FTS-only path stays supported).
- `get_embedding_provider` now raises `ConfigurationError` (was `ValueError`) for an unrecognised provider name, so it propagates as the canonical config error (and isn't caught by the `except (ImportError, RuntimeError)` gate).
- An unreachable Ollama/OpenAI *service* is **not** probed at startup (constructors don't connect) — it surfaces later as an embedding error during index build. Documented as such.

### Tests
- New `TestToVaultKwargsProvider` covers the full matrix: explicit ImportError/RuntimeError → raise; unrecognised name → raise; auto-detect ImportError/RuntimeError → degrade (no `embedding_provider` kwarg); no `embeddings_path` → resolver never called; happy path → provider threaded through.
- Asserts the provider→`max_chunk_chars` derivation integration (#649 coupling): `512` context → `round(512*2.8)`; no provider → `6000` fallback.
- Rewrote the previously-flaky `test_includes_all_vault_params` to monkeypatch the resolver (deterministic) instead of relying on a live FastEmbed download + `xfail`.
- 2067 tests pass; ruff + mypy + mkdocs `--strict` clean.

### Review
9-lens preflight circus run; addressed a doc inaccuracy (an unreachable service does not fail fast at startup) and two test improvements (the `max_chunk_chars` integration assertion + the xfail rewrite) before opening.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -96,7 +96,7 @@ The first three knobs adjust *ranking and rendering* — they take effect immedi
     2. **Ollama** — if `OLLAMA_HOST` is reachable
     3. **FastEmbed** — if the `fastembed` package is installed
 
-    **Explicit vs. auto-detect failure handling:** when you set `MARKDOWN_VAULT_MCP_EMBEDDING_PROVIDER` to a specific backend and it cannot be constructed at startup — a missing dependency, missing/empty credentials, or an unrecognised value — the server **fails fast** with a `ConfigurationError` rather than silently falling back to keyword-only search. (An unreachable Ollama/OpenAI *service* is not probed at startup; that surfaces later as an embedding error during index build.) When the variable is *unset* (auto-detect) and no backend is available, the server logs a warning and continues with semantic search disabled. Set the variable explicitly if you want a missing provider to be a hard startup error.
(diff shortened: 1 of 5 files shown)


pvliesdonk:
Thanks — addressed the docs wording note in 50ca5d5: you're right that for Ollama `to_vault_kwargs` reads `provider.context_length`, which probes `/api/show` (failing gracefully), so "not probed at startup" was inaccurate. Reworded to "an unreachable service does not *prevent* startup — the provider still loads; the failure surfaces later as an embedding error during index build."

The `_FakeProvider` duplication note: left as-is for now (two trivial 2-line inner classes); happy to DRY into a fixture in a follow-up if you'd prefer. The `ConfigurationError`-not-`RuntimeError` invariant is already documented in the `to_vault_kwargs` comment ("raises ConfigurationError from the resolver and propagates here unchanged").

PR #663, opened 2026-06-10
fix(write-callback): enforce close/fire thread contract (closes #601)

### LINKED ISSUE #601: harden WriteCallbackDispatcher: close/fire thread-contract + use-after-close guard (deferred from #599)
Surfaced while extracting `WriteCallbackDispatcher` in #599 (PR2 of the #576 facade-decomposition epic). These are **pre-existing** hazards carried verbatim from the former `Collection` callback worker (#175) — not regressions introduced by #599, which faithfully preserved them per the #582 verbatim-move contract. Filed for a separate hardening pass. Refs #599, #576.

## Observed

`src/markdown_vault_mcp/write_callback.py`:

1. **`close()` reads `self._worker` outside `_worker_lock`** (the lock `_ensure_worker` holds when it mutates `self._worker`). A `close()` racing a concurrent `fire()` is an unsynchronised read of a lock-published field. `[verified: read both methods]` Benign today because `Collection.close()` runs from single-threaded teardown after writes quiesce, so the race is not currently exercised — but the type makes no such guarantee.

2. **No use-after-close / closed-state guard.** After `close()` drains and joins, a subsequent `fire()` re-enters `_ensure_worker`, sees a dead thread, and spawns a new worker — enqueuing an item that will never be drained (no further `close()` is expected). `[verified: traced _ensure_worker restart path]` Use-after-close is silently accepted; double-close is safe only by the `is_alive()` accident, not by an explicit flag.

3. **Dead defensive branch in `_run`** (`if on_write is None: ... continue`). Unreachable: `_on_write` is set once in `__init__`, never reassigned, and `fire()` already short-circuits when it is `None`, so no real item is ever enqueued with a `None` callback. `[verified: only assignment is in __init__; fire guards]` Carried over verbatim; keeps `_run` at <100% coverage.

4. **`close()` timeout warning is not quantified.** On join timeout it logs "pending git commits may be lost" without `self._queue.qsize()`, so an operator can't tell whether one commit or fifty were lost. `[verified: read close()]`

## Expected

The dispatcher's lifecycle/thread contract should be enforced by the type, not by caller convention: `close()`/`fire()` reads of `_worker` consistently synchronised; use-after-close rejected (or made an explicit, logged no-op); the unreachable `_run` guard removed once the contract is enforced; the timeout warning quantified.

## Scope boundary

This is the `ReadinessState`-style "unenforced cross-thread contract" tech-debt, here for the write-callback dispatcher. Not about changing the deferred-commit design (#175) — git commits stay off the write path. Behaviour on the happy path must not change.

## Open questions

- `[unverified]` Whether to add a `threading.Lock`-guarded `_closed` flag (consulted by both `_ensure_worker` and `close`) vs. a different lifecycle shape; lock ordering against any caller lock must be checked.
- `[unverified]` Whether the nullable `on_write: WriteCallback | None` should instead be a disabled-case construction (no thread, unconditional `fire`), which would also delete the dead `_run` branch.

pvliesdonk:
First PR of M2 (the git-robustness milestone) of the #577 epic. Hardens `WriteCallbackDispatcher`'s lifecycle — enforcing the close/fire contract **by the type**, not by caller convention. These were pre-existing hazards carried from #175/#599 (filed as #601).

## Changes (`src/markdown_vault_mcp/write_callback.py`)
- **`close()` now reads `_worker` under `_worker_lock`** (was an unsynchronised read of a lock-published field) and sets an explicit **`_closed`** flag, so a second `close()` is an explicit idempotent no-op rather than relying on the `is_alive()` accident.
- **`fire()` after `close()` is a logged no-op** — it no longer resurrects a worker or enqueues an item that would never be drained (use-after-close guard). The closed-check, worker-ensure, and `queue.put` of a real item now all run **under `_worker_lock`**, so `fire()` is atomic w.r.t. `close()`: an item can never land after the sentinel. (`close()` releases the lock before the blocking `join()` — no deadlock; `_run` never takes the lock.)
- **Removed the unreachable per-item `if on_write is None` branch** in `_run`; the invariant (worker only starts via `fire()` when a callback is configured, and `_on_write` is set once) is enforced by a single capture + `assert` at worker start.
- **Join-timeout warning now quantifies the loss** — `%d pending git commit(s)`. (`qsize()` counts queued items + sentinel but not the in-flight item; those offsets cancel, so it equals the true at-risk count — commented so it isn't naively "corrected" to `qsize()-1`.)

## Tests
New `TestThreadContract`: fire-after-close is dropped (no resurrected worker, callback not run, logged), double-close is an explicit no-op; plus a quantified-timeout test asserting the exact pending count. 2069 → 2076.

## Verification
- `uv run pytest`: **2076 passed, 1 skipped**
- `uv run mypy src/ tests/`: Success
- `uv run ruff check . && ruff format --check .`: clean
- Local preflight-circus (8 lenses incl. silent-failure-hunter) clean at ≥80; concurrency reasoning (no deadlock / no item-after-sentinel / no lost item) independently verified; caller (`vault.py`) closes the coordinator before the dispatcher, so the fire-after-close drop is a safety net, not a normal path.

Closes #601. Refs #577

### DIFF
--- a/src/markdown_vault_mcp/write_callback.py
+++ b/src/markdown_vault_mcp/write_callback.py
@@ -45,12 +45,19 @@ class WriteCallbackDispatcher:
             queue.Queue()
         )
         self._worker: threading.Thread | None = None
+        # Guards every read/write of ``_worker`` and ``_closed`` AND the
+        # ``_queue.put`` of a real item, so ``fire`` is atomic with respect to
+        # ``close``: an item is enqueued only while not closed, and never after
+        # ``close`` has queued the sentinel.
         self._worker_lock = threading.Lock()
+        self._closed = False
 
     def fire(self, abs_path: Path, content: str, operation: WriteOperation) -> None:
         """Queue a callback invocation, starting the worker if needed.
 
-        No-op when no callback is configured.
+        No-op when no callback is configured. After :meth:`close`, this is a
+        logged no-op — it does not resurrect a worker or enqueue an item that
+        would never be drained.
 
         Args:
             abs_path: Absolute path of the written file.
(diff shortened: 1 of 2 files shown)


pvliesdonk:
Thanks — all three minor notes are intentional no-ops, confirming the disposition:

1. **Dead-worker restart in `_ensure_worker_locked()`** — agreed it's unreachable post-hardening (a dead worker implies `_closed`, so `fire()` returns early). Keeping the `is_alive()` guard as a cheap defensive net rather than asserting the invariant a second time; it's inert as you note.
2. **`assert` under `-O`** — intentional: the invariant is independently guaranteed (`_on_write` set once; worker only started from `fire()` which early-returns on `None`), and the `except Exception` backstop turns even the impossible case into a logged failure, not corruption.
3. **`"2 pending"` string coupling** — deliberate: that exact assertion is what pins the `qsize()` = at-risk arithmetic (the thing the comment warns not to "correct" to `qsize()-1`); a looser `\d+ pending` wouldn't catch a regression of that math.

No code changes. LGTM stands.

PR #671, opened 2026-06-11
test(git): de-flake test_multiple_writes_single_push deterministically (closes #430)

### LINKED ISSUE #430: Flaky test: test_multiple_writes_single_push races on Python 3.14
## Summary

`tests/test_git.py::TestGitWriteStrategyClass::test_multiple_writes_single_push` is flaky on Python 3.14 (3.11-3.13 are stable so far).

First observed: PR #428, run [24957717115](https://github.com/pvliesdonk/markdown-vault-mcp/actions/runs/24957717115/job/73078777668) on commit `68a4043`. The failing assertion was that ``note_4.md`` should NOT yet be in the bare-repo log immediately after the 5th write — but it was, because the deferred push had already fired in the ~300 ms gap.

Marked as ``xfail(strict=False)`` on Python 3.14 only in PR #428 to unblock CI; this issue tracks the actual fix.

## Root cause hypothesis

```python
strategy = GitWriteStrategy(token=None, push_delay_s=0.3)
for i in range(5):
    strategy(md_file, ..., "write")  # 5 writes, each schedules a push
# Race window: assertion below fires immediately
result = subprocess.run(["git", "log", ...])
assert "note_4.md" not in result.stdout
```

The test relies on the 300 ms `push_delay_s` window NOT closing between the 5th `strategy(...)` call and the `subprocess.run` that follows. On Python 3.14 (faster startup? different GC? subinterpreter changes?) the test process slips past 300 ms before the assertion runs, so the push has already happened and `note_4.md` is in the log.

## Proposed fixes (pick one)

1. **Use a longer delay + tighter assertion window**: bump `push_delay_s` to 2 s, then assert immediately. Cheap but still timing-dependent.
2. **Replace timer-based assertion with a control hook**: expose a way to inspect the strategy's pending-push state (e.g. `strategy._timer is not None`) and assert that instead of inspecting the bare repo log. Deterministic.
3. **Use a `threading.Event` from inside the strategy** so the test can `wait` on a signal rather than racing the timer.

Option 2 is preferred — it removes the timing dependency entirely and exercises the actual contract being tested ("are writes batched into a single push?").

## Verification

- [ ] Reproduce the failure deterministically (run on 3.14 enough times to confirm flake rate ≥ 10 %).
- [ ] Implement chosen fix.
- [ ] Run the test 100× on each of 3.11/3.12/3.13/3.14 — must pass every time.
- [ ] Remove the `xfail` marker.

## Out of scope

- Investigating other flakes in `tests/test_git.py` — this one is the only one that's surfaced so far. Address others separately if/when they appear.

pvliesdonk:
M2 (git-robustness) of the #577 epic. De-flakes `test_multiple_writes_single_push`, which raced the push-debounce timer.

## The flake
The test asserted "`note_4.md` not yet in the bare-repo log" immediately after 5 rapid writes — relying on the 0.3 s debounce timer NOT firing in the microseconds before the assertion. On Python 3.14 the process slips past 0.3 s first, the deferred push has already fired, and the assertion fails. It was `xfail`'d on 3.14 in #428; that marker was dropped during the git decomposition, leaving it unguarded and latently flaky.

## The fix — remove the timer race entirely
- `push_delay_s=30.0` so the idle timer **cannot** fire during the synchronous body.
- Assert the coalesced pending state directly: `_push_pending is True`, a single `_timer`, and zero pushes so far (`push_calls == []`).
- Force the single deferred push via `flush()`, then assert `len(push_calls) == 1` (coalesced — one push, not five) and all 5 commits in the bare repo (a counting `_push` wrapper calls through to the real push, so delivery is genuinely exercised).

The contract — *5 rapid writes coalesce into one push delivering all 5 commits* — is now verified by direct state + a push-call count, with no wall-clock dependence. (The issue's literal "check `_timer` right after the writes" suggestion is itself still racy with a short delay; long-delay + `flush()` + push-count is the deterministic realization.) The 3 s poll loop is gone too, so the test is faster, and `strategy.close()` moved into a `finally` so the timer can't leak on assertion failure.

## Verification
- New test runs **10/10 deterministically** (~0.2 s each); full suite **2076 passed**; mypy + ruff clean.
- Local preflight-circus (7 lenses incl. pr-test-analyzer) clean at ≥80 — determinism, preserved/strengthened coverage, and coalescing soundness independently confirmed.

Closes #430. Refs #577

### DIFF
--- a/tests/test_git.py
+++ b/tests/test_git.py
@@ -406,45 +406,62 @@ class TestGitWriteStrategyClass:
         strategy.close()
 
     def test_multiple_writes_single_push(
-        self, git_repo_with_remote: tuple[Path, Path]
+        self,
+        git_repo_with_remote: tuple[Path, Path],
+        monkeypatch: pytest.MonkeyPatch,
     ) -> None:
-        """Multiple rapid writes result in a single deferred push."""
+        """Multiple rapid writes coalesce into a single deferred push (#430).
+
+        Verified deterministically rather than by racing the debounce timer: a
+        long ``push_delay_s`` guarantees the timer never fires during the
+        synchronous body, the pending state is asserted directly, and the push
+        is then forced via ``flush()`` while counting the underlying ``_push``
+        calls. (The earlier "assert not-yet-in-the-bare-log right after the
+        writes" approach raced the 0.3 s timer and was flaky on Python 3.14.)
+        """
+        import markdown_vault_mcp.git.strategy as git_strategy
 
         work, bare = git_repo_with_remote
 
-        strategy = GitWriteStrategy(token=None, push_delay_s=0.3)
+        push_calls: list[tuple[object, ...]] = []
+        real_push = git_strategy._push
 
-        for i in range(5):
-            md_file = work / f"note_{i}.md"
-            md_file.write_text(f"# Note {i}\n")
-            strategy(md_file, f"# Note {i}\n", "write")
+        def counting_push(*args: object, **kwargs: object) -> None:
+            push_calls.append(args)
+            real_push(*args, **kwargs)  # type: ignore[arg-type]
 
-        # Not pushed yet.
-        result = subprocess.run(
-            ["git", "-C", str(bare), "log", "--oneline"],
-            capture_output=True,
-            text=True,
-        )
-        assert "note_4.md" not in result.stdout
+        monkeypatch.setattr(git_strategy, "_push", counting_push)
 
-        # Poll until push lands (max 3s).
-        for _ in range(30):
-            time.sleep(0.1)
+        # 30 s delay: the idle timer cannot fire during the writes below, so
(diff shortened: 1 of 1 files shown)


pvliesdonk:
Thanks — both nits are intentional, keeping as-is:

1. **Late module import** — matches this file's established inline-import convention (e.g. lines 25, 62, the `_run` imports, etc.); keeping it local is consistent with the surrounding tests.
2. **Explicit `push_delay_s=30.0`** — deliberately explicit rather than relying on the `__init__` default. The test's determinism *must not* silently depend on that default: if it were ever lowered (e.g. for faster commits), an implicit-default test would quietly become flaky again — the exact failure this PR fixes. Pinning it here decouples the test from the production default, and the adjacent comment already documents the "timer can't fire during the synchronous body" rationale.

Appreciate the thorough read.

PR #676, opened 2026-06-11
feat(write-callback): drain() primitive + fire write callbacks inside the file-write lock (refs #571)

### LINKED ISSUE #571: Write-callback / git-pull window can cause aborted pulls & push-rejection churn (not silent data loss)
## Summary

When an MCP write lands on disk in the sub-second window before its deferred git commit, concurrent with a git pull (periodic `sync_once`, webhook, or interactive `git_sync`), the pull can abort on a dirty working tree. The MCP write is **not lost** — it is preserved on disk and committed by the callback worker — but the interleaving produces operational churn: an aborted pull, a later non-fast-forward push rejection, and possibly a spurious `.conflict-mcp-*` sibling on the eventual reconcile.

> **Correction (2026-06-02):** the original version of this issue claimed silent data loss via the merge overwriting a dirty file. That mechanism is **not reachable** — see "Why the original framing was wrong" below. Severity downgraded from bug to low-priority robustness; `_write_lock` references corrected to `_file_write_lock`.

## Verified on `main` (commit 10b739c)

The write path is split across two phases by the callback queue:

1. **Disk write** — `DocumentManager.write()` (`src/markdown_vault_mcp/managers/document.py:487-533`) writes the file atomically under `_file_write_lock`, then fires the callback at line 533 **outside** the lock.
2. **Git commit** — the callback worker (`src/markdown_vault_mcp/collection.py:1645-1666`) dequeues and calls `GitWriteStrategy.__call__`, which does **not** hold `_file_write_lock`.

`__call__` (`src/markdown_vault_mcp/git.py:615`) holds `GitWriteStrategy._lock` around `_stage_and_commit`; `force_pull` (`git.py:1363`) and `sync_once` (`git.py:1796`) hold the **same** lock for their fetch + merge/rebase. So the git operations are mutually exclusive — they never physically interleave. `_stage_and_commit` (`git.py:2596`) stages with `git add -- <path>` (whatever is on disk) and commits.

## Why the original framing was wrong

The code uses only `git merge --ff-only` and `git rebase @{upstream}` — no `git stash`, `reset --hard`, `clean`, or `-f`. Git refuses to overwrite a dirty working-tree file: both merge and rebase **abort** when the uncommitted file would be touched. The `git checkout --ours` / `git checkout @{upstream} -- <path>` calls exist only inside rebase-conflict handling, which requires the rebase to have started — impossible on a dirty tree. So "the merge overwrites the file and `git add` stages the post-merge content" cannot happen.

## Actual behavior in the race window

A write lands (file on disk, dirty relative to HEAD, callback queued) just before a pull acquires `GitWriteStrategy._lock`:

- **Upstream did not touch that file** → `merge --ff-only` fast-forwards, leaves the dirty file untouched, the worker commits it next. No loss, no churn.
- **Upstream touched that file** → `merge --ff-only` aborts (dirty tree) → rebase fallback can't start (dirty tree → empty conflict set) → `force_pull` returns `conflict_resolution_failed`, HEAD unchanged, **the MCP write preserved on disk**. The worker commits it; the next push is non-fast-forward → rejected with the existing `git_sync(direction='pull')` hint; the next reconcile produces a `.conflict-mcp-*` sibling (#232). Data preserved throughout.

**Edge:** if an operator's git config sets `rebase.autostash=true`, `git rebase` would autostash the dirty change and could leave a conflicted stash on reapply — messier, but driven by external config the server does not set.

## On `pause_writes()`

`Collection.force_pull` wraps the pull in `pause_writes()` (#570, `collection.py:447`); the periodic `sync_once` does not (only its `on_pull=reindex` is wrapped). But `pause_writes()` holds `_file_write_lock`, which the merge does not need — so during a pull it only blocks *new* writes from landing. It neither un-dirties an already-landed write nor drains the callback queue. For this specific race the #570 wrapping is close to cosmetic.

## Correct fix shape (if pursued)

Before the merge in **both** `sync_once` and `force_pull`: block new writes **and drain the callback queue** so all pending writes are committed first. The merge then runs on a clean tree, converting "dirty tree at merge time" into "clean tree + local commit", which the existing rebase + Syncthing-sibling machinery (#232) already handles correctly.

- Prerequisite: the callback worker should hold `_file_write_lock` for the duration of each `_stage_and_commit`, so `pause_writes()` blocks until in-flight commits finish (not just new writes).
- Then add an explicit queue-drain (wait for the worker to process all queued items) before the fetch/merge.

## Risk note

This touches the concurrency-critical write path. The repo has a history of abandoned concurrency PRs (#513 / #519 / #520 / #528 / #531). Given the low severity (self-healing churn, no data loss), this is not worth a rushed change. If pursued, it must be TDD'd with explicit interleaving tests (write-lands-before-merge, both upstream-touches and upstream-untouches the file).

## References

- PR #570 (webhook) — Gemini review surfaced the underlying concern; partial `pause_writes()` mitigation deferred here.
- #232 — Syncthing-style conflict siblings (the reconcile path that preserves the data).

pvliesdonk:
**PR1 of #571** (drain-before-pull). Two safe, additive prerequisites for the M2 fix that stops a git pull aborting on a dirty tree when an MCP write lands just before its deferred commit. **This PR does not change pull behavior yet** — PR2 wires the quiescer and closes #571.

## 1. `WriteCallbackDispatcher.drain(timeout=30.0)`
Block until all *currently-queued* write callbacks have been committed, **without closing** the dispatcher (it keeps running afterward). Mechanism: a `_DrainMarker` (a `threading.Event`-carrying sentinel) enqueued under `_worker_lock` exactly like `fire()`; the worker sets the event when it reaches the marker (FIFO ⇒ all prior commits done) and *continues* (doesn't exit). No-op when no callback is configured / closed / worker never started; best-effort with a WARNING (and pending count) on timeout rather than hanging. Faithfully reuses the #601 close/fire lock contract.

## 2. Fire the write callback inside `_file_write_lock`
All five write ops (`write`, `write_attachment`, `edit`, `delete`, `rename`) now fire `_on_write_callback(...)` as the last statement *inside* the `with self._file_write_lock:` block instead of just after it. This makes "file landed on disk" ⟺ "callback queued" **atomic**, so once a puller holds the lock (PR2), no landed write can still be un-queued. `fire()` is a non-blocking enqueue and the dispatcher worker never takes `_file_write_lock`, so it's deadlock-free and behavior-preserving (each op still fires exactly once).

## Tests
- `TestDrain` (6): waits-for-all, keeps-worker-alive, the three no-op guards, timeout-warns (worker blocked via an `Event`, deterministic).
- Fire-under-lock atomicity probes for **all five** ops (a separate probe thread must fail to acquire the `RLock` during the callback — a genuine cross-thread discriminator, not a tautology).
- Restored an assertion (`"second occurrence body" not in nc.content`) accidentally dropped from an unrelated read-section test.

## Verification
2079 → **2087 passed**, mypy + ruff clean. Two-round preflight-circus (9 lenses incl. silent-failure-hunter, concurrency-reviewed) clean at ≥80. Sub-threshold circus carryovers (worker BaseException-death observability, `drain`→`bool` return, a qsize count assertion) are recorded against PR2, where `drain` gains a caller and they become load-bearing.

Refs #571

### DIFF
--- a/src/markdown_vault_mcp/managers/document.py
+++ b/src/markdown_vault_mcp/managers/document.py
@@ -536,8 +536,8 @@ class DocumentManager:
                 self._mark_paths_dirty([path])
 
             result = WriteResult(path=path, created=created)
+            self._on_write_callback(abs_path, file_content, "write")
 
-        self._on_write_callback(abs_path, file_content, "write")
         return result
 
     def write_attachment(
@@ -731,7 +731,8 @@ class DocumentManager:
             if self._mark_paths_dirty is not None:
                 self._mark_paths_dirty([path])
 
-        self._on_write_callback(abs_path, new_content, "edit")
+            self._on_write_callback(abs_path, new_content, "edit")
+
         return EditResult(path=path, replacements=1, match_type=match_type)
 
     def _edit_with_lines(
@@ -895,7 +896,8 @@ class DocumentManager:
                         )
                 abs_path.unlink()
 
-        self._on_write_callback(abs_path, "", "delete")
+            self._on_write_callback(abs_path, "", "delete")
+
         return DeleteResult(path=path)
 
     def rename(
@@ -1004,9 +1006,9 @@ class DocumentManager:
 
                 callback_content = ""
 
-        self._on_write_callback(new_abs, callback_content, "rename")
-        for src_abs, src_content in backlink_callbacks:
-            self._on_write_callback(src_abs, src_content, "edit")
+            self._on_write_callback(new_abs, callback_content, "rename")
+            for src_abs, src_content in backlink_callbacks:
+                self._on_write_callback(src_abs, src_content, "edit")
 
         return RenameResult(
             old_path=old_path,
--- a/src/markdown_vault_mcp/managers/document.py
+++ b/src/markdown_vault_mcp/managers/document.py
@@ -601,8 +601,8 @@ class DocumentManager:
                 Path(tmp_name).unlink(missing_ok=True)
                 raise
             result = WriteResult(path=path, created=created)
+            self._on_write_callback(abs_path, "", "write")
 
-        self._on_write_callback(abs_path, "", "write")
         return result
 
     def edit(
--- a/src/markdown_vault_mcp/write_callback.py
+++ b/src/markdown_vault_mcp/write_callback.py
@@ -159,6 +159,12 @@ class WriteCallbackDispatcher:
             # its sentinel ahead of this marker.
             self._queue.put(marker)
         if not marker.event.wait(timeout):
+            # On timeout the worker is blocked on an in-flight item (else it
+            # would have reached the marker), so qsize() counts the queued real
+            # items plus our still-queued marker but NOT that in-flight commit.
+            # The marker (+1) and the uncounted in-flight commit (-1) cancel, so
+            # qsize() equals the number of commits genuinely at risk — same
+            # accounting as close(). Do NOT "correct" this to qsize()-1.
             logger.warning(
                 "Write-callback drain did not finish within %s s; "
                 "%d pending git commit(s) not yet committed before pull.",
--- a/src/markdown_vault_mcp/write_callback.py
+++ b/src/markdown_vault_mcp/write_callback.py
@@ -24,6 +24,20 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
+class _DrainMarker:
+    """Queue sentinel meaning 'all items enqueued before me are processed'.
+
+    Unlike the ``None`` close-sentinel, the worker does NOT exit on this: it
+    sets ``event`` and continues. FIFO ordering guarantees every real item
+    enqueued before the marker has been processed when the worker reaches it.
+    """
+
+    __slots__ = ("event",)
+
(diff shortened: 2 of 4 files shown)


pvliesdonk:
Thanks — agreed on all counts, nothing to change for PR1:

- **Carryovers (worker `BaseException`-death observability, `drain()` → `bool` return, the `qsize` count assertion):** recorded against PR2 in the plan, where `drain()` gains its pull-path caller and they become load-bearing. And yes — the timeout count is **`"2 pending"`** (`[second, marker]` in the queue, the in-flight `first` uncounted; marker +1 / in-flight −1 cancel, same as `close()`), exactly as you reasoned; PR2 adds that assertion to mirror `test_close_timeout_warning_quantifies_pending`.
- **`write_attachment` passing `"write"` as the operation:** pre-existing and deliberately left as-is. It only affects commit-message granularity (`"write: assets/x.png"`), not behavior, and `WriteOperation` has no richer discriminant planned in PR2 (PR2 is the pull-quiescer, not the type) — so this is outside the #571/#577 robustness scope rather than a deferred fix. Happy to file it as a standalone cosmetic issue if you'd like attachment commits distinguishable, but I won't bundle it into the epic.

PR #677, opened 2026-06-11
feat(git): quiesce writes (pause + drain) before a pull so the merge runs on a clean tree (closes #571)

### LINKED ISSUE #571: Write-callback / git-pull window can cause aborted pulls & push-rejection churn (not silent data loss)

pvliesdonk:
**PR2 of #571** (drain-before-pull) — the integration that **closes #571**, built on PR1 (#676, merged). Eliminates the churn where an MCP write that lands on disk just before its deferred commit, concurrent with a git pull, aborts the pull on a dirty tree (→ non-fast-forward push rejection → spurious `.conflict-mcp-*` sibling on the reconcile).

## What it does
Before the fetch/merge in **both** pull paths (`force_pull` and the periodic `sync_once`), `GitWriteStrategy` now **pauses new writes and drains the deferred-commit queue** so the merge always runs on a clean tree — "dirty-at-merge" becomes "clean + already-committed-locally," which the existing rebase + #232 sibling machinery handles.

- **`WriteCallbackDispatcher.drain(timeout) → bool`** — block until the queue is empty *without closing*; `True` = drained/nothing-to-drain, `False` = timeout or dead worker. `_run` now logs + re-raises a `BaseException` worker death (previously silent), and the dead-worker drain returns `False` + logs the `~qsize()+1` stranded count.
- **`GitWriteStrategy.set_write_quiescer()` + `_quiesce_writes(skip)`** — `with self._quiesce_writes(skip=dry_run), self._lock:` enters pause → drain *before* `_lock`, so the dispatcher worker (which takes `_lock` to commit) drains freely while the puller holds only the file-write lock. The deadlock invariant is documented at the call site. Best-effort: a stalled/dead drain logs a WARNING and proceeds (never blocks the pull).
- **Wiring:** `Vault.__init__` calls `set_write_quiescer(pause_writes=…, drain_writes=self._write_callback.drain)` (so `force_pull` is covered even when the periodic loop is disabled); `Vault.force_pull`'s now-redundant pause-wrap is removed; `start()` drops its `pause_writes` param.

## Why this is the right prerequisite shape
The atomicity comes from PR1 (fire the write callback *inside* `_file_write_lock`): once the puller holds that lock, no landed write can still be un-queued, so the drain provably commits the whole backlog. (We deliberately did **not** take the issue's "worker holds `_file_write_lock` during commit" route — it would deadlock the drain.)

## Tests
Headline interleaving tests (the issue's TDD mandate): write-lands-uncommitted then a pull where upstream touched a **different** file (clean apply, no sibling) and the **same** file (the pre-fix `conflict_resolution_failed` discriminator → now rebased with the MCP write preserved as a `.conflict-mcp` sibling), for **both** `force_pull` and `sync_once`; a real-dispatcher cross-thread **no-deadlock** test (worker commits via the strategy taking `_lock` while the pull drains before acquiring `_lock`); dry-run-skips-quiesce; `drain` bool/timeout/dead-worker contract.

## Verification
2087 → **2098 passed**, mypy + ruff clean, docs updated (`design.md` + git-integration guide). Executed subagent-driven (implementer → spec-review → quality-review per task); **five full preflight-circus rounds**, final round clean at ≥80 across all lenses (the rounds caught a missed 5th write op, a stale docstring, and a self-introduced dead-worker count off-by-one — all fixed and re-verified).

Closes #571

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -598,7 +598,11 @@ The contract is:
   `.conflict-mcp-*` sibling on the eventual reconcile). The write callback now
   fires *inside* `_file_write_lock`, so once the puller holds that lock no
   landed write can still be unqueued; `WriteCallbackDispatcher.drain()` then
-  blocks until the queue is empty before the merge takes the git lock.
+  blocks until the queue is empty before the merge takes the git lock. The
+  drain is **best-effort and bounded**: if it does not finish within the
+  timeout (or the dispatcher worker has died), the pull logs a WARNING and
+  proceeds anyway, accepting the pre-fix dirty-tree behavior for the
+  still-pending commit rather than blocking the pull indefinitely.
 - The `on_write` callback fires in a **background thread** — it must not
   itself call write methods on the same Vault instance (deadlock).
 - Callbacks must not raise; exceptions are logged and swallowed.
--- a/docs/design.md
+++ b/docs/design.md
@@ -590,6 +590,15 @@ The contract is:
 - Concurrent file-write tools on the same path are serialised by
   `_file_write_lock`; the IndexWriter's FIFO queue serialises everything
   else.
+- Before a git pull (periodic `sync_once` or interactive `force_pull`), the
+  strategy **pauses new writes and drains the deferred-commit queue** so the
+  merge runs on a clean working tree (#571). A write that landed on disk just
+  before the pull is committed first, rather than aborting the merge on a dirty
+  tree (which previously caused a non-fast-forward push rejection and a spurious
+  `.conflict-mcp-*` sibling on the eventual reconcile). The write callback now
+  fires *inside* `_file_write_lock`, so once the puller holds that lock no
+  landed write can still be unqueued; `WriteCallbackDispatcher.drain()` then
+  blocks until the queue is empty before the merge takes the git lock.
 - The `on_write` callback fires in a **background thread** — it must not
   itself call write methods on the same Vault instance (deadlock).
 - Callbacks must not raise; exceptions are logged and swallowed.
--- a/docs/guides/git-integration.md
+++ b/docs/guides/git-integration.md
@@ -80,8 +80,9 @@ dry-run push always returns `applied=false` with
 
 !!! note "Writes landing during a pull"
     A write whose deferred git commit has not yet run when a pull starts is
-    never lost. Before every pull (periodic or `git_sync`), the server pauses
-    new writes and drains the deferred-commit queue, so in the normal case the
+    never lost. Before every real (non-dry-run) pull (periodic or `git_sync`),
+    the server pauses new writes and drains the deferred-commit queue (a
+    `dry_run` preview only fetches and never quiesces), so in the normal case the
     just-written file is committed first and the merge runs on a clean tree
     ([#571](https://github.com/pvliesdonk/markdown-vault-mcp/issues/571)). If
     that write and the remote touched the same file, it flows through the
--- a/docs/guides/git-integration.md
+++ b/docs/guides/git-integration.md
@@ -78,6 +78,15 @@ dry-run push always returns `applied=false` with
     sibling(s), reconcile the local content against the remote, and
     `delete` the sibling once merged.
 
+!!! note "Writes landing during a pull"
+    A write whose deferred git commit has not yet run when a pull starts is
+    **not** lost and does not abort the pull. Before every pull (periodic or
+    `git_sync`), the server pauses new writes and drains the deferred-commit
+    queue, so the just-written file is committed first and the merge runs on a
+    clean tree ([#571](https://github.com/pvliesdonk/markdown-vault-mcp/issues/571)).
+    If that write and the remote touched the same file, it flows through the
+    Syncthing-style sibling resolution above rather than failing.
+
 The full enumeration of `pull.reason` and `push.reason` values lives in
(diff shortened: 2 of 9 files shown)


pvliesdonk:
Thanks — addressed the one documentation note. `sync_once`'s docstring now states (in `3445402`) that the write-pause is held across the whole fetch + merge (including the network round-trip), so MCP writes block for the pull's duration — mirroring `force_pull`'s existing note, so the shared tradeoff is now documented symmetrically on both sides. It's an accepted tradeoff (a periodic background pull, default every 600 s, with a normally-fast fetch), not a regression. The nits in your review were all sub-blocking and left as-is.

The deeper "narrow the write-pause to just the merge by fetching outside the quiesce" is a possible future optimization, not pursued here — it would restructure the concurrency-critical locked region for marginal benefit (the fetch is read-only and fast), and the current behavior is correct + now documented.

PR #684, opened 2026-06-11
feat(git-tools): get_diff/get_history on attachments (closes #342)

### LINKED ISSUE #342: feat(git-tools): allow get_diff on attachments (non-.md files)
## Problem

`get_diff` and `get_history` currently reject non-`.md` paths. The `.md` constraint is enforced via the tool-level docstring contract and `Collection._validate_path()` conventions (notes are `.md`).

Attachments (images, PDFs, other whitelisted extensions) are also committed to git by the write pipeline, so their commit history exists and could be useful:

- Audit **when** a diagram was replaced, by whom, and with what commit message
- Surface `get_history` on a PNG to show its rename/replacement lineage
- For PDFs, see metadata-level changes (even if the binary diff itself is not meaningful)

LLM-client feedback labeled this **low priority**, but flagged it as a natural extension of the tool pair.

## Open Questions

Before implementing, a few design calls need to be made:

1. **Diff payload for binary files.** Unified diffs of binary files are meaningless. Options:
   - Return `git log`-style metadata only (commit list + messages + timestamps) — effectively "just call `get_history`"
   - Return `git diff --stat` (rename/copy detection, size change) instead of the full patch body
   - Return `"Binary files differ"` when git reports binary
   - Reject with a clear error: "diff not supported for binary attachments, use get_history"

2. **Extension allowlist coupling.** Should this reuse `ATTACHMENT_EXTENSIONS` from the write pipeline, or accept any file under the vault?

3. **Tool surface.** Keep overloading `get_diff` (dispatch on extension) or introduce `get_attachment_history` / `get_attachment_diff`?

4. **`_validate_path()` scope.** Does relaxing validation here risk loosening it elsewhere? The path validator is shared; any relaxation needs to be scoped to history tools only (or a new validator variant).

## Proposed Direction (to discuss)

- `get_history(path=...)` relaxes the `.md` check when the path exists inside the vault and matches `ATTACHMENT_EXTENSIONS` (or is any tracked file — open question)
- `get_diff(path=...)` on a binary attachment returns `{diff: "<binary file>", binary: True, stat: "<--stat output>"}` — never a raw patch
- `per_commit=True` on a binary file returns the per-commit metadata list but with `diff` replaced by the `--stat` line

## Acceptance Criteria

(pending design decision above)

## Documentation

- `docs/tools/index.md` — update `get_history`/`get_diff` to list supported path types
- `docs/design.md` — update git-tools section with attachment semantics
- `docs/guides/` — add an "auditing attachment history" example if a guide exists

## Priority

**Low.** File for tracking; pick up after the `since` on `get_history` and `limit` on `get_diff per_commit` land (issues #339, #340).

---

Follow-up from LLM client feedback on PR #337 (git history tools).

pvliesdonk:
**M3 (features) of the #577 git epic — the final epic issue.**

## What
`get_history` and `get_diff` now accept attachment paths (configured non-`.md` extensions), not just notes:
- **`get_history`** returns an attachment's commit history (replacement/rename lineage).
- **`get_diff`** lets **git classify each file** (`git diff --numstat` → `-\t-`): a **binary** attachment returns a `git diff --stat` summary (`assets/x.png | Bin 1234 -> 5678 bytes`), a **text** attachment (`.svg`, `.csv`, …) returns a real unified diff, and **`.md` notes are completely unchanged** (`summarize_binary=False` default → byte-identical code path).

## Design
- A new scoped `validate_history_path` (accepts `.md` + configured attachment extensions, incl. the `"*"` wildcard) — the strict-`.md` `validate_path` used by write/edit/read is **untouched** (issue Q4).
- `GitQueryManager` gains `attachment_extensions` (wired from `Vault`); passes `summarize_binary=not path.endswith(".md")`.
- `get_file_diff` resolves the rename target once (`resolve_path_at_ref`) and reuses it for binary detection + `--stat` + the full diff, so the **non-per-commit binary diff is rename-aware** (a renamed binary still gets a `--stat`, not the meaningless `Binary files … differ` patch). New `--stat` subprocess wraps `CalledProcessError → ValueError` (the #337 contract).

## Tests
Real-git fixtures (binary png, text svg, note across commits): binary→`--stat` (markers `" -> "` present / `"Binary files"` absent — verified discriminators), text→full diff, note unchanged, per-commit variants, no-change→`""`, a **rename discriminator** (NUL-byte fixture, mutation-verified it reproduces the garbage patch when reverted), a bad-ref+`summarize_binary` error test, and the full `validate_history_path` suite. 2198 passed; mypy + ruff clean.

## Known limitation (documented + tracked: #683)
`per_commit=True` diff of a **renamed binary** currently renders a text-style stat for the rename commit (the single-path pathspec defeats git's rename pairing in `git show`). It's documented in the code/docstrings/design.md and **pinned by a test** that flips when #683 is fixed — narrow (per-commit + renamed + binary), graceful (not a crash), and the non-per-commit path handles it correctly.

## Verification
Preflight-circus (5 core + code-reviewer + silent-failure-hunter + pr-test-analyzer + comment-analyzer) clean at ≥80 over four rounds — it drove the wildcard fix, the binary-test discriminators, the rename-awareness, and the #683 binding. Docs updated (README, `docs/tools/index.md`, `docs/design.md`, all docstring layers). Designed via brainstorm→spec→plan.

Closes #342. Refs #577

### DIFF
--- a/README.md
+++ b/README.md
@@ -396,8 +396,8 @@ markdown-vault-mcp reindex [--source-dir PATH] [--index-path PATH]
 | `get_orphan_notes` | Find all notes with no inbound or outbound links |
 | `get_most_linked` | Find the most-linked-to notes ranked by backlink count |
 | `get_connection_path` | Find the shortest path between two notes via BFS on the undirected link graph (max 10 hops) |
-| `get_history` | List commits that touched a note or the whole vault (git-backed vaults only) |
-| `get_diff` | Return a unified diff of a note between a reference commit/timestamp and HEAD (git-backed vaults only) |
+| `get_history` | List commits that touched a note, attachment, or the whole vault (git-backed vaults only) |
+| `get_diff` | Return a diff of a note or attachment between a reference commit/timestamp and HEAD; binary attachments return a `--stat` size summary instead of a unified patch (git-backed vaults only) |
 | `git_sync` | Force an immediate git pull / push / both, bypassing the periodic loops. Returns structured state (SHAs, commit counts, Syncthing-style conflict file paths if any). Hidden when `MARKDOWN_VAULT_MCP_GIT_REPO_URL` isn't set or `READ_ONLY=true`. |
 | `fetch` | Download a file from a URL and save it to the vault as a note or attachment (MCP-to-MCP transfer) |
 | `create_download_link` | Mint a one-time capability URL to download a vault note or attachment (HTTP/SSE only; `BASE_URL` required) |
--- a/docs/design.md
+++ b/docs/design.md
@@ -2237,6 +2237,14 @@ Both methods use the existing `_git_env()` / `_cleanup_git_env()` pattern for cr
 
 **MCP response envelope**: the MCP wrappers for `get_history` and `get_diff(per_commit=True)` return a `{"commits": [...], "total": N}` envelope rather than a bare list, so the structured payload is self-describing on the wire. FastMCP otherwise auto-wraps list-typed tool returns under a synthetic `"result"` key (`x-fastmcp-wrap-result: true` in the output schema), which forces clients re-reading persisted MCP content to know FastMCP's wrapping convention to find the data. The Python facade (`ReaderFacet.get_history`, `ReaderFacet.get_diff`) stays list-returning — only the MCP-tool wrapper transforms to the envelope. `get_diff(per_commit=False)` keeps its existing `{"diff": str}` shape since it is already self-describing.
 
+**Attachment history & diff (#342).** `get_history` and `get_diff` accept a
+`.md` note OR a configured attachment, validated by `validate_history_path`
+(distinct from the strict-`.md` `validate_path` that the write/edit/read paths
+use). `get_diff` lets git classify each attachment: a binary file (git
+`--numstat` reports `-\t-`) returns a `git diff --stat` size/rename summary,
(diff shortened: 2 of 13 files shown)


pvliesdonk:
Thanks — dispositioning the five (all rated minor; CI green, no Gemini findings, local preflight-circus clean at ≥80 over four rounds):

- **#1 (per-commit binary flag is range-level)** and **#5 (`resolve_path_at_ref` skips copy records)** — both are narrow `get_file_diff` edges in the same area as the already-tracked per-commit renamed-binary `--stat` limitation, so I've **folded them into #683** (a single rename-aware/copy-aware per-commit path-resolution fix covers all three). Not blocking this PR; the common path and the non-per-commit binary diff are correct.
- **#2 (no extension-less-path validator test)** and **#4 (`_diff_is_binary` docstring should note `check=False` errors surface downstream)** — minor test/docstring polish; added to #683's deliverables. The behaviors are already correct and covered by logic/the `test_get_file_diff_summarize_binary_invalid_ref_raises` test respectively.
- **#3 (SHA regex rejects uppercase hex)** — pre-existing and intentional (`[0-9a-f]{4,40}`); git itself is case-insensitive on SHAs, so the strictness is conservative. Accepted as-is; not introduced by this PR.

None of the five is a correctness defect in the shipped paths, so none blocks merge. The per-commit renamed-binary case remains the one documented + test-pinned limitation (#683).

PR #721, opened 2026-06-22
fix: break file-watcher reindex self-feedback loop

### LINKED ISSUE #720: File watcher self-feedback loop: reindex walk re-triggers itself, pinning CPU
## Summary

On a non-git vault with the file watcher active (`MARKDOWN_VAULT_MCP_FILE_WATCHER=true`, no git pull, no webhook), the server reindexes in an endless loop and burns ~20–30 % CPU at idle, with **no real file changes**. `docker logs` shows a continuous stream of:

```
reindex: 0 added, 0 modified, 0 deleted, N unchanged, 0 skipped
```

There are no embedding calls — it is pure scan/hash/save churn.

## Root cause (verified with py-spy + inotifywait)

A self-feedback loop between the reindex walk and the watcher:

1. `IndexManager.reindex()` → `tracker.detect_changes()` walks the whole vault (read-only: `glob`/`walk`/`stat`/`compute_file_hash`).
2. On Linux that walk makes the kernel emit inotify **read** events — `OPEN`, `CLOSE_NOWRITE` (also on directories / `ISDIR`) — for every entry touched.
3. `_VaultEventHandler.on_any_event` in `src/markdown_vault_mcp/_file_watcher.py` filters only **hidden paths** (`.git/`, `.markdown_vault_mcp/`), **not the event type**. So those read events on non-hidden dirs are forwarded to `_schedule()` (2 s debounce) → `on_change` → reindex → walk → more read events → loop.

py-spy hot path: `index_writer._run → run_reindex_all → managers/index.py:reindex → tracker.detect_changes` and the unconditional `vectors.save()` near the end of `reindex()`.

## Two defects

1. **`_file_watcher._VaultEventHandler.on_any_event`** forwards *every* event type, including read-only ones (`opened` / `closed_no_write` / `closed`). It should only react to mutating events (`created` / `modified` / `moved` / `deleted`). This is what breaks the loop.
2. **`managers/index.py` `reindex()`** calls `vectors.save(...)` unconditionally, even on an empty-diff pass (0 added/modified/deleted). Re-serialising the whole vector index to disk on every (looping) run is needless churn; it should only save when the vector index actually changed.

## Expected behaviour

- Idle vault → ~0 % CPU, no reindex spam in the logs.
- A real file change in scope → exactly one reindex, then quiet again.
- `vectors.save()` only runs when the reindex actually mutated the vector index.

## Environment

- File watcher mode (no git pull loop, no webhook), Linux/inotify backend, watchdog >= 4.0.

Finomosec:
Closes #720

## Problem

On a non-git vault with the file watcher active, the server reindexes in an endless loop at idle (~20–30 % CPU), with no real file changes:

```
reindex: 0 added, 0 modified, 0 deleted, N unchanged, 0 skipped
```

### Root cause (verified with py-spy + inotifywait)

1. `reindex()` → `tracker.detect_changes()` walks the whole vault read-only.
2. On Linux that walk emits inotify **read** events (`opened` / `closed_no_write`, also on directories).
3. `_VaultEventHandler.on_any_event` filtered only hidden paths, **not the event type**, so those reads were forwarded to the debounce scheduler → reindex → walk → more reads → loop.

## Changes

1. **`_file_watcher.py`** — `on_any_event` now forwards only an allowlist of mutating event types (`created` / `modified` / `moved` / `deleted`); read-only events (`opened` / `closed` / `closed_no_write`) are dropped before the hidden-path check. This breaks the loop. The hidden-path filter is unchanged.
2. **`managers/index.py`** — `reindex()` no longer calls `vectors.save()` on an empty-diff pass; it saves only when the vector index was actually mutated (added/modified/deleted > 0), removing needless disk churn on every watcher wake-up.

## Tests

- `test_file_watcher.py`: `opened`/`closed_no_write`/`closed` events do **not** schedule a reindex; `created`/`modified`/`moved`/`deleted` (+ `DirModified`) do; a mutating event inside a hidden dir is still ignored.
- `test_managers_index.py`: an empty-diff reindex does **not** call `vectors.save()`; a real change calls it exactly once.

All gates pass locally: `pytest`, `ruff check`/`format`, `mypy --strict`, patch coverage 100 % on both changed files.

## Documentation

Behavioural bug fix in internal watcher/indexing logic — no user-facing config, tool, resource, or env-var surface changes; no docs pages require updates.

### DIFF
--- a/src/markdown_vault_mcp/_file_watcher.py
+++ b/src/markdown_vault_mcp/_file_watcher.py
@@ -24,10 +24,34 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 try:
-    from watchdog.events import FileSystemEvent, FileSystemEventHandler
+    from watchdog.events import (
+        EVENT_TYPE_CREATED,
+        EVENT_TYPE_DELETED,
+        EVENT_TYPE_MODIFIED,
+        EVENT_TYPE_MOVED,
+        FileSystemEvent,
+        FileSystemEventHandler,
+    )
     from watchdog.observers import Observer
 
     _WATCHDOG_AVAILABLE = True
+
+    # Only these event types represent an actual mutation of the vault.
+    # Read-only events (``opened``, ``closed``, ``closed_no_write``, access)
+    # must be ignored: ``reindex()`` walks the whole vault read-only, and on
+    # Linux that walk emits inotify read events (OPEN / CLOSE_NOWRITE, also on
+    # directories) for every entry it touches. Forwarding those to the
+    # debounce scheduler would make each reindex re-arm the watcher, which
+    # fires another reindex — an endless self-feedback loop that pins the CPU
+    # with no real file changes (#720).
+    _MUTATING_EVENT_TYPES = frozenset(
+        {
+            EVENT_TYPE_CREATED,
+            EVENT_TYPE_MODIFIED,
+            EVENT_TYPE_MOVED,
+            EVENT_TYPE_DELETED,
+        }
+    )
 except ImportError:
     _WATCHDOG_AVAILABLE = False
 
@@ -197,7 +221,12 @@ class VaultFileWatcher:
 if _WATCHDOG_AVAILABLE:
 
     class _VaultEventHandler(FileSystemEventHandler):
-        """Forward non-hidden filesystem events to the debounce scheduler."""
+        """Forward mutating, non-hidden filesystem events to the scheduler.
+
+        Read-only events (``opened`` / ``closed`` / ``closed_no_write``) are
+        dropped before the hidden-path check so the read-only vault walk that
+        ``reindex()`` performs cannot re-trigger itself (#720).
+        """
 
         def __init__(self, schedule: Callable[[], None], source_dir: Path) -> None:
             super().__init__()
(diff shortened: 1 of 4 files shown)


pvliesdonk:
Looks good to me.
@claude review

pvliesdonk:
Thanks @claude — the `stale_excluded` gap is real and now fixed in `c26b0a9`.

**Fix:** `vector_index_changed` now includes `stale_excluded`:

```python
vector_index_changed = bool(
    indexed_added or indexed_modified or changes.deleted or stale_excluded
)
```

When `exclude_patterns` purge an already-indexed doc, the stale-excluded loop calls `vectors.delete_by_path()` (a real in-memory mutation) while the file diff stays empty — so without `stale_excluded` the save was skipped and the stale vectors reloaded on next startup, exactly as you described.

**Tests added** (both non-vacuous — each fails if its operand is dropped from the guard):
- `test_stale_excluded_reindex_saves_vectors` — sets `_exclude_patterns` after build; asserts `(added, modified, deleted) == (0, 0, 0)` yet `vectors.save` called once.
- `test_deletion_only_reindex_saves_vectors` — closes the parallel gap for the `changes.deleted` operand (a pure delete drives the save with all other operands at 0), which the pre-existing embeddings-less delete test never exercised.

Ran the full local pre-flight review (five context lenses + code-reviewer + silent-failure-hunter + pr-test-analyzer + comment-analyzer): no findings at the ≥80 confidence bar; the deletion-only test above was the one sub-threshold suggestion, added anyway. `pytest` / `ruff` / `mypy` all green.

Two unrelated red signals to disregard for this PR:
- **Dependency Audit** (CI) fails on `pydantic-settings 2.13.1` (GHSA-4xgf-cpjx-pc3j, fixed in 2.14.2) — pre-existing transitive-dep drift, untouched by this PR. Worth a separate bump.
- The config-wizard Playwright smoke tests false-fail locally against a stale built `site/` (they pass after `mkdocs build`, and skip in CI) — tracked separately in #722; not a regression from this change.

PR #732, opened 2026-06-23
fix: atomic VectorIndex sidecar writes + self-heal corrupt sidecars on load

### LINKED ISSUE #733: Interrupted VectorIndex.save() corrupts embeddings sidecar and wedges boot until manual rebuild
## Summary
An interrupted `VectorIndex.save()` can leave a truncated or zero-byte embeddings sidecar (`embeddings.npy` / `embeddings.json`); the corrupt file is then re-read on every boot and keeps semantic search down until a manual force rebuild.

## Observed
A process quit mid-save left a truncated `embeddings.json` in production. On a synchronous-load path this crashed the server with a `-32000` connection failure (JSON parse error). On current `main` the async embeddings build no longer crashes for this, but it does not self-heal: `_load_vectors` (`src/markdown_vault_mcp/managers/index.py`) only rebuilds on `VectorIndexCompatibilityError`, so a corrupt sidecar is re-read every boot and semantic search stays down until a manual force rebuild.

## Expected
- A mid-write interruption must never corrupt a previously persisted vector index.
- A corrupt/incomplete sidecar found at load time should self-heal (rebuild) rather than wedge every boot.

## Context
- `VectorIndex.save()` (`src/markdown_vault_mcp/vector_index.py`) writes both sidecars directly to their final paths — non-atomic. The atomic-write pass in #322 covered the document writes but left the vector sidecars non-atomic.
- Reproduced in production via a quit mid-save.

## Scope boundary
This tracks the prevention (atomic sidecar writes) + recovery (self-heal on load) work delivered in PR #732. It does not cover the count-mismatch parity gap that remains after that PR (tracked separately).

## Resolution
Fixed by PR #732 — atomic `mkstemp` + `os.replace()` sidecar writes (mirroring `tracker._save_state`) and a broadened `_load_vectors` catch that routes corrupt-sidecar errors to a `force=True` rebuild.

mikebronner:
## Summary
`VectorIndex.save()` writes the embeddings sidecars (`embeddings.npy`, `embeddings.json`) directly to their final paths. An interrupted write (process kill / sleep mid-save) can leave a truncated or zero-byte sidecar. We hit this in production: a quit mid-write left a truncated `embeddings.json`, which on a synchronous-load path crashed the server with a -32000 connection failure (JSON parse error). On current `main` the async embeddings build no longer crashes for this — but it does not self-heal: `_load_vectors` only rebuilds on `VectorIndexCompatibilityError`, so the corrupt sidecar is re-read every boot and semantic search stays down until a manual force rebuild.

This PR addresses both halves — prevention and recovery.

## Prevention — atomic sidecar writes
`save()` writes each sidecar to a temp file and `os.replace()`s it into place, mirroring the existing `tracker._save_state` / `managers/document.py` idiom (the atomic-write pass in #322 covered the document writes but left the vector sidecars non-atomic). A failed write unlinks its temp and re-raises, so a previously persisted index is left intact.

## Recovery — self-heal on load
`_load_vectors` now treats a corrupt/incomplete sidecar like a compatibility miss: it catches `json.JSONDecodeError`/`ValueError` (truncated/garbage `.json`, corrupt or bad-version `.npy`), `EOFError` (a zero-byte `.npy` — the canonical interrupted-save residue), and `FileNotFoundError` (a missing `.json` while `.npy` exists), routing to the same `force=True` rebuild. Broad `OSError` is deliberately not caught — `PermissionError`/disk-IO faults are environmental and continue to propagate rather than triggering a futile destructive rebuild. A fully-missing index still cold-builds.

## Notes
- Sidecars are made atomic independently, not as a unit; this closes the truncation window (a crash between the two replaces can still leave a count-mismatched but individually valid pair, which still loads).
- No `fsync` before replace — consistent with the existing `tracker`/`document` pattern; guards against process-kill, not power-loss.

## Tests
- Atomic save: prior sidecars survive a `json.dump` failure, survive an `np.save` failure, and a failed save leaves no temp files behind.
- Self-heal: corrupt `.json`, zero-byte `.npy`, garbage `.npy`, and missing `.json` each trigger a rebuild; a `PermissionError` on load propagates (no rebuild); a fully-missing index cold-builds.
- Full suite green; `ruff` + `mypy` clean.

---
Closes #733. Residual count-mismatch parity gap (acknowledged in Notes above) tracked as follow-up in #734.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -1707,6 +1707,10 @@ Key methods:
   embeddings have been computed outside a lock section (see Thread Safety).
 - `delete_by_path(path)`: removes all rows for a document.
 - `save(path)` / `load(path, provider)`: persist/restore sidecar files.
+  `save()` writes each sidecar (`.npy`, `.json`) to a temp file in the same
+  directory and atomically `replace()`s the final path (mirrors
+  `tracker._save_state`), so an interrupted write can never corrupt a
+  previously persisted index.
 
 ### `providers.py`: Embedding Providers
 
--- a/src/markdown_vault_mcp/managers/index.py
+++ b/src/markdown_vault_mcp/managers/index.py
@@ -164,6 +164,36 @@ class IndexManager:
                     raise ValueError(
                         "Failed to rebuild vector index after a compatibility error."
                     ) from exc
+            except (
+                json.JSONDecodeError,
+                ValueError,
+                EOFError,
+                FileNotFoundError,
+            ) as exc:
+                # A corrupt/incomplete sidecar wedges every boot until a manual
+                # rebuild (#720 follow-up): an interrupted save can leave a
+                # truncated or zero-byte file. Self-heal by routing to the same
+                # force-rebuild path as a compatibility mismatch.
+                #   ValueError — truncated/garbage .json (JSONDecodeError is a
+                #     ValueError subclass) and corrupt/bad-version .npy.
+                #   EOFError — a zero-byte .npy (the canonical interrupted-save
+                #     residue; numpy raises EOFError, not ValueError/OSError).
+                #   FileNotFoundError — a missing .json while the .npy exists
+                #     (an incomplete pair); the .npy-exists guard above means a
+                #     FileNotFoundError here can only be the gone .json sidecar.
+                # Deliberately NOT broad OSError: PermissionError / disk-IO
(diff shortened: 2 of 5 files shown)


pvliesdonk:
Thank you very much!

PR #750, opened 2026-06-27
fix: resolve git sync ref as origin/<branch> not @{upstream}

### LINKED ISSUE #749: Managed-git sync depends on @{upstream} tracking; breaks on non-tracking / detached clones
## Problem

`GitWriteStrategy`'s sync paths resolve the remote-tracking branch through `@{upstream}` (the branch's configured upstream), with an `origin/HEAD` fallback in a couple of places. This couples managed-git sync to branch *tracking* being configured on the local clone.

A managed clone is created and maintained by the server. Depending on how it was produced, it may have **no upstream tracking** configured:

- `git clone --branch <b> <url>` without later `push -u`
- a detached `HEAD` checkout (e.g. pinned to a SHA)
- a clone whose `branch.<name>.merge` / `.remote` config was never set

When `@{upstream}` does not resolve, the affected operations silently degrade:

- **Periodic pull** (`sync_once`) and the **pull loop guard** (`start`) log "no upstream configured" and skip fetching entirely — the vault never pulls.
- **Startup unpushed check** (`_push_if_unpushed`) skips.
- **`force_pull` / `force_push`** return `no_remote`.

The `origin/HEAD` fallback that exists in `force_pull` / `force_push` is also subtly wrong on a feature branch: `origin/HEAD` points at the remote's *default* branch, not the branch the clone is actually on.

## Affected sites

`src/markdown_vault_mcp/git/strategy.py`: `_push_if_unpushed`, `force_pull`, `_force_pull_rebase_fallback`, `force_push`, `sync_once`, `start`; and `src/markdown_vault_mcp/git/conflict.py`: `restore_upstream_paths` (the rebase-abort `checkout @{upstream} -- <path>`).

## Proposed fix

Resolve the remote-tracking ref **tracking-independently** as `origin/<current-branch>`, derived from `git symbolic-ref --short HEAD`, with a fallback to `origin/HEAD` only when HEAD is detached (no symbolic branch) or the branch has no counterpart on `origin`. This stays entirely within the existing "everything is `origin`" invariant — no new config, no behavior change for normally-tracking clones — it just removes the dependency on tracking being set.

A single helper (`resolve_tracking_ref(git_root, env)`) centralizes the resolution; all sites use it in place of `@{upstream}`.

## Documentation

- `docs/design.md` — note the tracking-independent ref resolution in the git-sync section.
- `docs/tools/index.md` — update the `no_remote` / `rebased` reason descriptions.

No user-facing config or env var changes.

Finomosec:
## Summary

Managed-git sync resolved the remote-tracking branch through `@{upstream}`, coupling it to branch *tracking* being configured on the local clone. A managed clone may be a detached or `git clone --branch` checkout with no upstream set — in which case `@{upstream}` does not resolve and fetch/merge/rebase/push silently degrade (the vault never syncs).

This PR makes ref resolution **tracking-independent**: a single `resolve_tracking_ref()` helper derives `origin/<current-branch>` from `git symbolic-ref --short HEAD`, falling back to `origin/HEAD` only when HEAD is detached (or the branch has no counterpart on `origin`). It stays entirely within the existing "everything is `origin`" invariant — **no new config, no env var, no behavior change for normally-tracking clones**.

## Changes

- `git/_run.py`: new `resolve_tracking_ref(git_root, env)` helper.
- `git/strategy.py`: route `_push_if_unpushed`, `force_pull`, `_force_pull_rebase_fallback`, `force_push`, `sync_once`, and `start` through it; replace the `@{upstream}` / `origin/HEAD` two-step lookups.
- `git/conflict.py`: `restore_upstream_paths` now takes the resolved `ref` for the rebase-abort `checkout <ref> -- <path>`.
- `git/types.py`: `no_remote` / `rebased` reason docstrings updated.
- Docs: `docs/design.md` (git-sync section) + `docs/tools/index.md` (reason tables).
- Tests: new `tests/test_git_tracking_ref.py` (normal / non-tracking / detached / no-remote resolution, no-remote-ref guards, and an end-to-end `force_pull` on a clone with tracking unset); updated the checkout-failure simulation in `test_git_force_methods.py` to match the new ref.

## Gates

- `uv run pytest -x -q` — green
- `uv run ruff check` + `ruff format --check` — clean
- `uv run mypy src/ tests/` — no issues
- Patch coverage 89% (diff-cover vs `main`, gate ≥80%)

Closes #749

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -2192,7 +2192,7 @@ local commits are pushed in a single `git push`. On shutdown,
 `Vault.close()` flushes any pending push.
 
 Startup recovery: `GitWriteStrategy` checks for unpushed local commits
-(`git log @{upstream}..HEAD`) on first invocation and pushes them before
+(`git log origin/<branch>..HEAD`) on first invocation and pushes them before
 accepting new writes.
 
 **Deferred push mechanics**: `GitWriteStrategy` uses a `threading.Timer` that
@@ -2243,6 +2243,17 @@ Set `MARKDOWN_VAULT_MCP_GIT_LFS=false` for repos that do not use LFS, or when
   enables concurrent readers during index writes).
 - If `MARKDOWN_VAULT_MCP_GIT_LFS=true`, each successful pull tick ends with
   `git lfs pull` so LFS pointer files are resolved before reads and indexing.
+
+**Tracking-independent remote ref**: all sync operations (startup unpushed
+check, periodic `sync_once`, interactive `force_pull`/`force_push`, and the
+rebase-abort upstream restore) target `origin/<current-branch>` rather than the
+branch's configured upstream (`@{upstream}`). A managed clone is created and
+maintained by the server and may be a detached or `git clone --branch` checkout
+with no upstream tracking configured; deriving the ref from the current branch
+name (`git symbolic-ref --short HEAD`) keeps sync working regardless. When HEAD
+is detached, or the current branch has no counterpart on `origin`, resolution
+falls back to `origin/HEAD` (the remote's default branch); when neither
+resolves, the operation reports `no_remote` and is skipped.
 
 Safety branch mode for push failures is tracked separately (see #119).
 
--- a/docs/tools/index.md
+++ b/docs/tools/index.md
@@ -488,8 +488,8 @@ failures; `null` for clean fast-forwards and dry-runs):
 | Reason | Meaning | `applied` |
 |--------|---------|-----------|
 | `"fetch_failed"` | `git fetch origin` exited non-zero (network / auth / proxy). HEAD did not move. | `false` |
-| `"no_remote"` | Neither `@{upstream}` nor `origin/HEAD` could be resolved on the local clone. | `false` |
-| `"rebased"` | Local and remote diverged but `git rebase @{upstream}` replayed local commits cleanly. `conflict_files` empty. | `true` |
+| `"no_remote"` | No remote-tracking ref (`origin/<branch>`, or `origin/HEAD` for a detached checkout) could be resolved on the local clone. | `false` |
+| `"rebased"` | Local and remote diverged but `git rebase origin/<branch>` replayed local commits cleanly. `conflict_files` empty. | `true` |
 | `"conflicts_resolved_with_siblings"` | Rebase hit real conflicts; resolved by accepting upstream and writing local versions as `.conflict-mcp-*` siblings (#232). `conflict_files` populated. | `true` |
 | `"conflict_resolution_failed"` | The conflict-resolution loop could not produce a recoverable working tree; rebase was aborted. HEAD did not move. | `false` |
 | `"non_fast_forward_with_conflicts"` | Rare catastrophic fallback when even the conflict-resolution path could not stabilise the working tree. HEAD did not move. | `false` |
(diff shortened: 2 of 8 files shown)


pvliesdonk:
@claude review

pvliesdonk:
Pushed a follow-up commit (`a217b09`) addressing the `claude-review` finding plus a local review pass.

**Addressed the medium finding** — `_force_pull_rebase_fallback`'s `ref = self._tracking_ref(...) or "@{upstream}"`. Rather than swapping the `or "@{upstream}"` for an early-exit guard, the resolved `ref` is now passed **down as a parameter** from `force_pull` (which already resolves and `None`-guards it before delegating). This removes the re-resolution entirely, drops the `@{upstream}` string and the unreachable `None` branch (the two lines Codecov flagged uncovered), and makes the `_force_pull_rebase_fallback` docstring's "force_pull already verified a ref resolves" claim structurally true rather than defended by a comment.

**Also in the same commit:**
- Fixed three stale `@{upstream}` references the ref change made inaccurate: `conflict.py` `resolve_rebase_conflicts` docstring and the pathspec-less-commit `NOTE`, and the `strategy.py` `_resolve_rebase_conflicts` wrapper docstring (all now name `git rebase <ref>` / `checkout <ref>`).
- Removed a redundant inline comment that duplicated the new `ref` `Args:` entry; hoisted an inline `import subprocess` to the module top.
- **Added two regression tests** in `TestForcePullWithoutTracking` driving the rebase fallback through `force_pull` on a `--unset-upstream` clone — a clean divergent rebase (`PULL_REASON_REBASED`) and a same-file sibling-conflict resolution. Both **fail** if the rebase target reverts to `@{upstream}` (verified), closing the gap where the suite would otherwise stay green on a tracking-dependent regression — since the previous non-tracking test only exercised a fast-forward.

Local gates green: `pytest` (2367 passed), `ruff check`/`format`, `mypy` clean. A five-lens local review pass (CLAUDE.md adherence, diff bugs, git history, prior-PR concerns, comment accuracy) is clean.

**Two observations from local review, below the merge-blocking confidence bar — flagging for your call, not blocking:**

1. **`origin/HEAD` fallback widens the destructive rebase target (design question).** `resolve_tracking_ref` falls back to `origin/HEAD` not only for a detached HEAD but also when an attached branch has no `origin/<branch>` counterpart. For the documented single-branch managed-vault clone (on `main` tracking `origin/main`) this never fires — `origin/main` resolves via the primary path. But a clone checked out on a never-pushed named branch with divergent history + pull enabled would now silently rebase onto the remote default branch. Pre-`a217b09` the rebase fallback used literal `git rebase @{upstream}`, which simply errors and bails when upstream is unset — so this PR does widen the mutating path. It's outside the documented single-branch contract and the `origin/HEAD` fallback is documented as intentional, so I left it as-is. If multi-branch managed clones are ever in scope, consider restricting the fallback to genuinely-detached HEAD (or logging a `WARNING` when it fires).

2. **One test-coverage edge remains.** The new tests cover the rebase *target* on non-tracking clones, but `restore_upstream_paths(ref)` runs only on the rebase-**abort** path, which the same-file-conflict test sidesteps via `rebase --continue`. A successful `git checkout <ref> -- <path>` restore on a non-tracking clone is therefore still unexercised (its fail-case is covered by `test_force_pull_checkout_failure_drops_path_from_siblings`). A monkeypatch-forced-abort test would close it if you want belt-and-suspenders on that sub-path.

---

PR #756, opened 2026-06-27
feat: move_folder folder-level move with vault-wide link rewrite

### LINKED ISSUE #511: feat(rename): folder-level move/rename that updates links across the vault
## Summary

Add the ability to move/rename an **entire folder** (subtree) in one MCP call, rewriting all affected links — the directory-level analogue of the existing single-file `rename(old, new, update_links=True)` tool.

## Motivation

Vault reorganization (e.g. `drafts/` → `archive/2026/`, or renaming a project folder) is a common, high-value operation. Today an LLM agent can only do it by enumerating every file under the folder and issuing one `rename` per file. That is:

- **Tedious and error-prone** — the agent must list, loop, and reconstruct each child path.
- **Link-incorrect under the naive loop** — moving files one at a time corrupts links *between* documents in the same subtree: when file A moves, its backlink in still-unmoved sibling B is rewritten to A's new home, then B itself moves, and the intermediate rewrites interleave badly. A correct folder move must compute the full old→new path mapping for every moved file and rewrite links in a single pass *after* the subtree has moved.
- **Not atomic** — a failure partway through leaves the vault in a half-moved state with no clear recovery.

## Current behavior (the gap)

`rename` operates on a single file and explicitly rejects directories:

- `DocumentManager.rename` (`src/markdown_vault_mcp/managers/document.py:924`) guards with `old_abs.is_file()` and raises `DocumentNotFoundError` for a directory.
- Link rewriting is handled by `_update_backlinks` (`managers/document.py:1042`), which rewrites every source doc that links to the moved file via `_compute_new_raw_target` + `_apply_link_replacement`, followed by a vault-wide `resolve_vault_wikilinks()`.
- MCP surface: `rename` tool in `src/markdown_vault_mcp/_server_tools.py:1459`; facade in `Collection.rename` (`collection.py:1124`); result type `RenameResult`.

There is **no** way to move a folder and fix links in one operation.

## Proposed feature

Move every document and attachment under a source folder prefix to a target folder prefix, preserving the relative structure, and rewrite all links across the vault that point into the moved subtree.

### Tool surface — two options (decide during planning)

**Option A — new dedicated tool** (e.g. `move_folder(old_dir, new_dir)`):
- Cleaner contract: no per-file `if_match` etag (meaningless for a directory), `update_links` defaults on, and a folder-level result shape (counts of files moved, links updated, per-file failures).
- Existing single-file `rename` stays untouched.

**Option B — extend `rename`** to detect when `old_path` is a directory and move the whole subtree:
- Smaller tool surface, but overloads `if_match` / `update_links` semantics and the `RenameResult` shape across two structurally different operations.

Both are recorded here intentionally; pick one when the implementation plan is written.

### Failure semantics: atomic move, best-effort links

- **Directory move is all-or-nothing.** Validate first (source exists and is a non-empty dir within the vault; target prefix does not collide with existing files), then perform the move. If the target collides, fail *before* moving anything.
- **Link rewrites are best-effort**, mirroring today's `_update_backlinks`: a source file that can't be rewritten is logged and skipped, and the operation reports how many links were updated vs. failed. This matches existing behavior rather than introducing transaction/rollback machinery the codebase doesn't have.

## Key design considerations (for the implementation plan)

- **Single-pass link rewrite over an old→new path map.** Move the subtree first (ideally one `shutil.move` of the directory, or per-file moves under a single lock), build a mapping of every moved `old_path → new_path`, then rewrite links across the whole vault in one pass. This correctly handles links *between* documents inside the moved subtree as well as backlinks from outside.
- **Attachments included.** The existing `rename` moves both `.md` notes and allowlisted attachments; a folder move should carry attachments along (notes get index + link updates; attachments are moved without index changes, as today).
- **Index consistency.** Delete old FTS/vector entries and re-parse moved notes at their new paths (the per-file rename path already does this); call `resolve_vault_wikilinks()` once at the end, not per file.
- **Path validation.** Reuse `_validate_path` / `_validate_attachment_path` traversal guards for both source and target; reject moves that escape the vault.
- **Read-only + git callbacks.** Respect `_check_writable()`; fire write/edit callbacks (git commit) for moved files and rewritten sources, consistent with the single-file path.
- **Empty-directory cleanup.** Remove the now-empty source directory tree after a successful move.
- **Result shape.** Report `old_dir`, `new_dir`, `files_moved`, `updated_links`, and a list of per-file link-rewrite failures.

## Acceptance criteria

- [ ] An MCP tool moves a folder subtree (notes + attachments) to a new prefix in one call.
- [ ] Links across the vault pointing into the moved subtree are rewritten (markdown links, wikilinks), including links *between* documents inside the subtree.
- [ ] The move is atomic: a target collision (or validation failure) aborts before any file is moved; link rewrites are best-effort with a reported success/failure count.
- [ ] FTS + vector indexes are consistent after the move (old entries purged, moved notes re-indexed, `resolve_vault_wikilinks()` run once).
- [ ] Read-only mode rejects the operation; git write callbacks fire for moved + rewritten files.
- [ ] Source directory is removed when empty after the move.

## Documentation

- [ ] `docs/design.md` — document the folder-move semantics and the chosen tool surface.
- [ ] `docs/tools/index.md` — new/changed tool entry with the agent-facing contract.
- [ ] `README.md` — tool list / feature mention.
- [ ] MCP server `instructions` string in `src/markdown_vault_mcp/server.py` (currently documents single-file `rename`).
- [ ] Inline docstrings on the new/changed manager, facade, and tool methods.

## Testing

- [ ] Folder move with inter-subtree links, external backlinks, wikilinks, and attachments.
- [ ] Target-collision abort (nothing moved).
- [ ] Best-effort link rewrite: one unwritable source is skipped and reported, the rest succeed.
- [ ] Read-only mode rejection.
- [ ] Index consistency (search finds moved notes at new paths, not old).

pvliesdonk:
Closes #511.

## Summary

Adds a `move_folder(old_dir, new_dir)` MCP tool — the folder-level analogue of single-file `rename`. It moves an entire folder subtree (`.md` notes, allowlisted attachments, and any other files) to a new prefix in one call and rewrites every link across the vault that points into the moved subtree (markdown links **and** wikilinks), including links *between* documents inside the subtree as well as external backlinks.

## Design decisions (locked in the spec)

- **Dedicated tool**, not a `rename` overload — a subtree has no meaningful `if_match` etag or single-file result shape.
- **Atomic at the gate**: a pre-existing destination-file clash aborts before anything moves. Merging into an existing target dir is allowed; only per-file name clashes abort.
- **Move all files, index/link only `.md` + allowlisted attachments** — a faithful folder move so the source dir ends up empty and removable.
- **`update_links` always on** — a folder move that left links dangling would be useless.
- **Single-pass link rewrite** over an old→new path map; a source inside the subtree is read at its new location so relative targets compute correctly (prefix-shift-idempotent for intra-subtree links).
- **Best-effort link rewrites** reported in `failed_links`; no transaction/rollback machinery (matches existing `_update_backlinks`).
- Single batched `mark_paths_dirty` drives one `resolve_vault_wikilinks()`; empty source dir removed after a successful move.

## Result shape

`MoveFolderResult(old_dir, new_dir, files_moved, updated_links, failed_links)`.

## Implementation (5 commits, TDD)

1. `MoveFolderResult` dataclass.
2. Behavior-preserving refactor: extract `_rewrite_one_source` from `_update_backlinks` (shared by single-file rename and folder move; 91/91 rename regression parity).
3. `DocumentManager.move_folder` + `WriterFacet.move_folder` + 18 tests.
4. `move_folder` MCP tool (`title: "Move Folder"`; exact-manifest + title-enforcement tests updated).
5. Docs: `docs/design.md`, `docs/tools/index.md`, `README.md`, server `instructions`.

A final whole-branch review surfaced one honest gap — a mid-*move* OS error (permissions/disk-full/concurrent deletion) can leave the subtree partially moved with the index unchanged. Per the locked "no rollback machinery" decision, this is now **documented across all surfaces and pinned by a test** (`test_move_folder_mid_move_oserror_leaves_subtree_partial`); a `reindex` reconciles.

## Gates

- `uv run pytest -x -q` → **2392 passed, 2 skipped**
- `uv run ruff check` / `ruff format --check` → clean
- `uv run mypy src/ tests/` → clean
- New-code patch coverage ~96.6%.

### DIFF
--- a/.vale/styles/config/vocabularies/Base/accept.txt
+++ b/.vale/styles/config/vocabularies/Base/accept.txt
@@ -47,6 +47,7 @@ traceback
 accessors
 affordances
 allowlist
+allowlisted
 ANDed
 APIs
 async
@@ -167,6 +168,7 @@ stderr
 streamable
 subprocess
 subqueries
+subtree
 svg
 syscalls
 teardown
--- a/README.md
+++ b/README.md
@@ -23,11 +23,11 @@ Point it at a directory of Markdown files (an Obsidian vault, a docs folder, a Z
 > **Upgrading.** As of this release, `search` returns query-relevant snippets in the `content` field by default (approximately 200 words). Pass `snippet_words=0` to recover the prior full-chunk behaviour, or use `read(path, section=heading)` to fetch the full section after seeing a snippet. Documents are also re-chunked on next `reindex` to honour the adaptive `MARKDOWN_VAULT_MCP_MAX_CHUNK_WORDS` threshold (default 400).
 - **Frontmatter-aware** — indexes YAML frontmatter fields, supports required field enforcement
 - **Incremental reindexing** — hash-based change detection, only re-processes modified files; an automatic boot-time reconciliation pass picks up changes made while no server was running, and the vector index converges to the reconciled chunk set (embedding exactly the delta)
-- **Write operations** — create, edit, delete, rename documents with automatic index updates
+- **Write operations** — create, edit, delete, rename documents, and move entire folder subtrees with automatic index updates
 - **Attachment support** — read, write, delete, and list non-markdown files (PDFs, images, etc.)
 - **Git integration** — optional auto-commit and push on every write via `GIT_ASKPASS`
 - **OIDC authentication** — optional token-based auth for HTTP deployments (Authelia, Keycloak, etc.)
-- **MCP tools** — 31 LLM-visible tools including search, read, write, edit, delete, rename, git history, manual git sync, one-time transfer links, and admin operations; plus 6 app-only tools for MCP Apps clients
+- **MCP tools** — 32 LLM-visible tools including search, read, write, edit, delete, rename, move_folder, git history, manual git sync, one-time transfer links, and admin operations; plus 6 app-only tools for MCP Apps clients
 - **MCP resources** — 9 resources exposing vault configuration, statistics, tags, folders, document outlines, similar notes, recent notes, and an interactive SPA
 - **MCP prompts** — 7 prompt templates including template-driven note creation
 <!-- DOMAIN-END -->
@@ -383,6 +383,7 @@ markdown-vault-mcp reindex [--source-dir PATH] [--index-path PATH]
 | `edit` | Replace text in a document — exact match, line-range, or scoped match with normalized fallback |
 | `delete` | Delete a document or attachment and its index entries |
 | `rename` | Rename/move a document or attachment, updating all index entries; pass `update_links=true` to also rewrite backlinks in other notes |
+| `move_folder` | Move an entire folder subtree to a new prefix, rewriting all vault links that point into the moved subtree in one call |
 | `list_documents` | List indexed documents; pass `include_attachments=true` to also list non-markdown files |
 | `list_folders` | List all folder paths in the vault |
 | `list_tags` | List all unique frontmatter tag values |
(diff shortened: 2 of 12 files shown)


pvliesdonk:
Thanks — all three findings addressed in `ddc08d6`.

**Finding 1 (Medium — double write-callback):** Confirmed. An intra-subtree note that was also link-rewritten landed in both `md_map.values()` (→ `rename` callback) and `pending_callbacks` (→ `edit` callback), emitting a redundant git event for the same path. Step 5 rewrites the file *before* the step-7 callbacks, and the `rename` callback re-reads the rewritten content from disk, so the `edit` callback was pure redundancy. Now suppressed via `moved_note_paths = set(md_map.values())`, with a regression test (`test_move_folder_no_double_callback_for_intra_subtree_source`) asserting each moved note fires exactly one callback.

**Finding 2 (Low — frozen dataclass with a list):** Agreed and fixed. `MoveFolderResult` was the *only* frozen result dataclass; `RenameResult`, `IndexStats`, etc. are plain `@dataclass`. Dropped `frozen=True` (your option B) to match convention — a frozen dataclass holding a list is neither immutable in practice nor hashable.

**Finding 3 (Low — destructiveHint):** Agreed. `move_folder` `shutil.rmtree`s the source and carries the documented mid-move partial-failure window, a materially larger blast radius than single-file `rename`. Set `destructiveHint=True`; `test_server.py` annotation grouping updated to assert it alongside `delete`.

**Minor (rmtree silent swallow):** Also addressed — replaced `ignore_errors=True` with a `try/except OSError` that logs a warning, per the no-silent-failure logging standard.

Gates after the change: 2393 passed / 2 skipped, ruff + mypy clean.

PR #772, opened 2026-06-29
feat: converge config with template drift gate (ProjectConfig + pvl-core 4.3.0)

### LINKED ISSUE #767: config: v2.7.0 drift gate test_config_wizard_drift.py fails to import against this repo
Part of #766. **Sequence: 1 of 5 (blocker — do first).**

## Summary

The v2.7.0 template gate `tests/test_config_wizard_drift.py` cannot import this repo's config module, so the whole file fails to collect on `copier update`.

## Observed `[verified: ran it]`

- `uv run python -c "from markdown_vault_mcp.config import ProjectConfig"` →
  `ImportError: cannot import name 'ProjectConfig' from 'markdown_vault_mcp.config'`.
- This repo's config class is `VaultConfig` (`src/markdown_vault_mcp/config.py:72`); the template gate hard-imports `ProjectConfig` (pristine `tests/test_config_wizard_drift.py:26`). Because the import is at module top level, every test in the file fails at collection — not a single assertion.
- The gate is **new in v2.7.0** (absent in v2.5.3). The template `config.py` body is byte-identical between v2.5.3 and v2.7.0; what changed is the addition of the gate that enforces the `ProjectConfig.from_env` convention. (`image-generation-mcp`#274 was the same class of blocker — there the cause was a missing `from_env`; here `from_env` exists but the class name differs.)

## Second observation — the gate is also blind to this repo's config `[verified: read the gate source]`

Independent of the class name: the gate's coverage check derives the domain env surface from `inspect.getsource(ProjectConfig.from_env)` and only matches calls literally named `env` / `env_int` / `env_float` (pristine `test_config_wizard_drift.py:72,87-94`). This repo:

- delegates domain env reads into `config_sections/` sub-configs — `VaultConfig.from_env` calls `GitConfig.from_env(prefix)`, `IndexingConfig.from_env(prefix)`, … (`config.py:419-426`); the literal reads live in e.g. `config_sections/git.py`, and
- imports the helper aliased as `_env` (`config.py:27`: `from …config_sections._helpers import env as _env`).

Both make the domain reads invisible to the gate's AST scan, so its coverage half would treat the domain surface as empty — passing while not actually checking coverage of the ~40 domain env vars.

## Context `[verified]`

- The `config_sections/` decomposition is the sanctioned outcome of #579 (owner-closed). #579's own open questions flagged the template-conformance of the sentinel / env-read regions as `[unverified]` and deferred it. This gate is the template now asserting that convention.
- `paperless-mcp` (same template) uses class `ProjectConfig` with literal `env(_ENV_PREFIX, "SUFFIX")` reads directly inside `ProjectConfig.from_env`.

## Scope boundary

Records the observed gate failure and the observed AST-blindness only. It does **not** prescribe a resolution — the approach (template-side gate change, repo-side change, or an accepted documented divergence) is to be decided before implementation. The `config_sections/` split itself is sanctioned and is **not** up for reversal in this issue.

## Open questions

- `[unverified]` Whether the resolution is template-side (teach the gate to recurse into delegated sub-config `from_env` sources — upstream-first), repo-side (rename + surface the reads), or an accepted divergence relying on this repo's own `test_every_inventory_var_is_wired_or_allowlisted`. To be discussed.
- `[unverified]` Whether the `VaultConfig` → `ProjectConfig` rename has callers beyond `vault.py` / `server.py` / `cli.py` / `tests/` / `docs/`.

I have not verified the resolution — only the failure and its mechanism.

pvliesdonk:
Closes #767. Final phase of the epic #766 config-convergence (after pvl-core #211 → v4.3.0 and template #224 → v2.8.0).

## What

Converges MVM's config with the v2.8.0 config-wizard drift gate, which hard-imports `ProjectConfig` and measures the domain env surface via `fastmcp_pvl_core.domain_env_suffixes`:

- **Rename `VaultConfig` → `ProjectConfig`** (direct, no alias — the template/sibling class name the gate imports). ~290 occurrences across src/tests/docs.
- **Un-alias `env` in `config.py`** (was `env as _env`) so the five top-level scalar reads (`SOURCE_DIR`/`READ_ONLY`/`DISABLE_APPS_UI`/`SERVER_NAME`/`INSTRUCTIONS`) are visible to the gate's AST scan.
- **Bump `fastmcp-pvl-core` floor to `>=4.3.0`** (ships `domain_env_suffixes`, which recurses into the `config_sections/` sub-configs).
- **Add `test_domain_env_suffixes_sees_decomposed_reads`** — asserts the scanner sees a top-level read plus reads from four representative sub-configs, with a count lower-bound guarding against a sub-config silently dropping out of recursion.

The #579 `config_sections/` decomposition is preserved — the gate now measures it correctly via the upstream recursive scanner, instead of the old single-function scan that was blind to it.

## Verification

`domain_env_suffixes(ProjectConfig)` returns 40 suffixes including `SOURCE_DIR` (top-level, proving the un-alias) and reads from git/embeddings/content/transfer/indexing/search/sync (proving sub-config recursion). Full suite: **2381 passed**, `mypy src/ tests/` clean, ruff clean.

## Follow-up (noted, not in this PR)

The wizard-spec inventory extractor's import-alias resolution (`_helper_alias_map`, added in #734986d for config.py's old `_env` alias) is now production-unused — worth a separate decision on keep-as-defensive vs remove.

### DIFF
--- a/docs/api/config.md
+++ b/docs/api/config.md
@@ -6,15 +6,15 @@ The `config` module loads configuration from environment variables and provides 
 
 ```python
 import os
-from markdown_vault_mcp import Vault, VaultConfig
+from markdown_vault_mcp import Vault, ProjectConfig
 
 os.environ["MARKDOWN_VAULT_MCP_SOURCE_DIR"] = "/path/to/vault"
-config = VaultConfig.from_env()
+config = ProjectConfig.from_env()
 vault = Vault(**config.to_vault_kwargs())
 ```
 
 ## API Reference
 
 <!-- vale off -->
-::: markdown_vault_mcp.config.VaultConfig
+::: markdown_vault_mcp.config.ProjectConfig
 <!-- vale on -->
--- a/docs/design.md
+++ b/docs/design.md
@@ -1867,7 +1867,7 @@ template functions with no vault dependency.
 
 | URI | Source | Description |
 |-|-|-|
-| ``config://vault`` | ``VaultConfig`` | Source dir, read-only flag, indexed fields, extensions |
+| ``config://vault`` | ``ProjectConfig`` | Source dir, read-only flag, indexed fields, extensions |
 | ``stats://vault`` | ``ReaderFacet.stats()`` | Document/chunk/folder counts, capabilities |
 | ``tags://vault`` | ``ReaderFacet.list_tags()`` | All tags grouped by indexed field |
 | ``tags://vault/{field}`` | ``ReaderFacet.list_tags(field)`` | Flat list for one field (template) |
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ classifiers = [
     "Topic :: Scientific/Engineering :: Artificial Intelligence",
 ]
 dependencies = [
-    "fastmcp-pvl-core>=4.1.0,<5",
+    "fastmcp-pvl-core>=4.3.0,<5",
     "typer>=0.12",
 
     # PROJECT-DEPS-START — add domain dependencies below; kept across copier update
--- a/src/markdown_vault_mcp/__init__.py
+++ b/src/markdown_vault_mcp/__init__.py
@@ -20,7 +20,7 @@ import importlib
 from typing import TYPE_CHECKING, Any
 
 if TYPE_CHECKING:
-    from markdown_vault_mcp.config import VaultConfig
+    from markdown_vault_mcp.config import ProjectConfig
     from markdown_vault_mcp.exceptions import (
         ConcurrentModificationError,
         ConfigurationError,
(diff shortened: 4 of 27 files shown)


pvliesdonk:
Thanks — addressed in `5ee3a1c`:

- **Stale comment (`test_config_wizard_spec.py:42`)** — fixed. Updated to note the extractor resolves aliases per-file but `config.py` now imports `env` without an alias (#767). (The pre-flight review caught the line-262 comment but missed this one — good catch.)
- **Loose count bound** — tightened `>= 35` to `>= 38` (full surface ~40), shrinking the silent-loss window.
- **`_helper_alias_map`** — agreed, deferred to a follow-up decision; nothing changed here.

PR #780, opened 2026-06-29
feat: folder/subtree table-of-contents + get_toc tool (#773)

### LINKED ISSUE #773: Folder/subtree table-of-contents (and: per-note TOC is resource-only, no tool)
## Summary

There is no way to get a heading-level table of contents for a **folder/subtree** of the vault — only for a single note. Navigating a large vault area's structure currently requires reading or searching note-by-note.

## What exists today `[verified]`

- `ReaderFacet.get_toc(path)` returns a single note's H1–H6 outline — a list of `{heading, level}` (the document title prepended as a synthetic H1). Backed by the FTS `sections` table (`fts_index.get_toc` → `managers/document.get_toc`).
- It is exposed **only as an MCP resource**: `toc://vault/{path}` (`_server_resources.py:vault_toc`, returning `{level, text, anchor}` JSON). There is **no TOC tool** — confirmed: `grep -rn "toc" src/markdown_vault_mcp/_server_tools/` returns nothing.
- Related read/navigation surfaces that do *not* fill this gap: `read(path)` / `read(path, section=heading)` (full content, one note), `search` (returns matching section headings, not a full ordered outline), `list_documents` / `list_folders` (paths only, no headings).

## The gap

1. **Subtree/folder TOC** — no aggregation of headings across the notes under a folder. A caller wanting "the structure of `Projects/` " has no single call.
2. **Per-note TOC is resource-only** — clients/agents that surface MCP *tools* but not *resources* cannot retrieve even a single note's outline as a tool call.

## Scope boundary

This is about a read-only outline/navigation surface (headings + levels/anchors + their note path). NOT about full-content retrieval (that's `read`), search ranking, or the graph/link tools.

## Open questions `[unverified]`

- Is a per-note TOC **tool** worth adding given the existing `toc://vault/{path}` resource, or is the resource sufficient? Depends on how clients consume resources vs tools.
- For a subtree TOC: what bounds the output? (max depth of subtree, max notes, max heading level, whether to include the per-note path per heading). A large folder could produce a very large response — does it need pagination / a heading-level cap / a "headings ≤ Hn" filter?
- Should a subtree TOC reuse the FTS `sections` table (cheap, index-dependent) the way `get_toc` does, and therefore require a built index (bucket 3)?
- Output shape: flat list of `{path, heading, level, anchor}` vs nested-per-note vs nested-by-heading-hierarchy?

## Documentation impact (if pursued)

New/changed tool or resource → `docs/tools/index.md` and/or `docs/resources.md`, `README.md`, `docs/design.md`. If a per-note TOC tool is added, cross-reference the existing `toc://vault/{path}` resource.

pvliesdonk:
Closes #773.

## What this adds

A heading-level **table of contents for a folder/subtree**, and exposes the per-note TOC as a **tool** (it was resource-only). Both go through one path-dispatched library method.

`get_toc(path, *, max_level=None, max_notes=200)` dispatches on the `.md` suffix:

- **Note path** (`a/b.md`) → flat `list[{heading, level}]` with the title as a synthetic H1 (unchanged shape; now also honors `max_level`).
- **Folder path** (`a/b`) → nested-per-note `{path, notes: [{path, title, headings}], truncated}`, recursive over the subtree, ordered by path, capped at `max_notes` (default 200) with a `truncated` flag.

Exposed two ways with identical behavior:

- **Extended resource** `toc://vault/{path}` — now dispatches to folders too (defaults only). The pre-existing inaccurate `{level, text, anchor}` docstring is corrected to the real `{heading, level}` shape.
- **New tool** `get_toc` — same dispatch plus the `max_level` / `max_notes` / `wait_for_pending_writes` controls; `needs_queryable` (bucket 3), `_meta.index_stale` reported out-of-band.

## Design notes

- **Boundary match.** Folder matching reuses the precomputed `documents.folder` column with `folder = ? OR folder LIKE ? ESCAPE '\'` (via the existing `_escape_like`), so `Project` never matches `Projects/…` or a sibling `Projectile/`.
- **Bounds.** Truncation is detected with `LIMIT max_notes + 1`; `max_level` is pushed into SQL; the synthetic H1 is prepended *after* the filter, so it always survives.
- **Union return + staleness.** The tool's `list | dict` return makes FastMCP's output schema require a `{"result": …}` envelope on every branch. This is handled by the shared `_staleness_result` helper via a new `force_result_wrap=True` flag — keeping the out-of-band `index_stale` contract in one place (no inlined duplicate, no `# type: ignore`). `result.data` is still the bare list/dict for clients.
- **No schema migration; backward compatible.** Note-mode output is byte-identical to before; the resource extension is purely additive.

## Out of scope

Anchors, pagination, heading-hierarchy nesting. Tightening the TOC payload types (`dict[str, Any]` → `TypedDict`/dataclass) is tracked separately in #779 — deliberately deferred since `get_toc` already returned dicts pre-PR and the JSON wire shape is unchanged.

## Testing

- Full suite green (2404 passed); `ruff`, `ruff format --check`, and `mypy src/ tests/` all clean.
- **100% patch coverage** on all changed source files (diff-cover).
- New tests cover: path dispatch; both shapes; `max_notes` at-cap vs over-cap; boundary prefix (`Project`/`Projects`/`Projectile/`); `_escape_like` wildcard folder names; trailing-slash normalization; empty/nonexistent folder → empty (no raise); missing note → `ValueError`; folder traversal/root → `ValueError` (manager) and `McpError` through the MCP stack (resource); `max_level` in both modes with the synthetic H1 surviving; heading order within a note; `_prepend_title_h1` dedup; the tool's `_meta.index_stale` signal (added to the uniform-stale parametrized test) and the `{"result": …}` envelope on both branches; the `readOnly`/`destructive` annotations; and full-registry title/manifest enforcement.

## Docs

`docs/design.md`, `docs/resources.md`, `docs/tools/index.md`, and `README.md` updated in-PR for the new tool and extended resource (both shapes, the bounds, and the resource↔tool cross-reference).

### DIFF
--- a/README.md
+++ b/README.md
@@ -396,6 +396,7 @@ markdown-vault-mcp reindex [--source-dir PATH] [--index-path PATH]
 | `get_outlinks` | Find all links from a document, with existence check |
 | `get_broken_links` | Find all links pointing to non-existent documents |
 | `get_similar` | Find semantically similar notes by document path |
+| `get_toc` | Heading outline for a note or a folder subtree |
 | `get_recent` | Get the most recently modified notes |
 | `get_context` | Get a consolidated context dossier for a note (backlinks, outlinks, similar, folder peers, tags, modified time) |
 | `get_orphan_notes` | Find all notes with no inbound or outbound links |
--- a/docs/design.md
+++ b/docs/design.md
@@ -1812,6 +1812,7 @@ pattern). Each tool is annotated with MCP `ToolAnnotations`:
 | `get_outlinks` | Find links from a document (with exists flag) | `True` | `False` | `True` |
 | `get_broken_links` | Find links to non-existent documents | `True` | `False` | `True` |
 | `get_similar` | Find semantically similar notes by path | `True` | `False` | `True` |
+| `get_toc` | Heading outline for a note or folder subtree | `True` | `False` | `True` |
 | `get_recent` | Get most recently modified notes | `True` | `False` | `True` |
 | `get_context` | Get consolidated context dossier for a note | `True` | `False` | `True` |
 | `get_orphan_notes` | Find notes with no inbound or outbound links | `True` | `False` | `True` |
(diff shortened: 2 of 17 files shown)


pvliesdonk:
Addressed the `max_notes=0` finding in b6b1edd. Rather than only the `max_notes` guard, `DocumentManager.get_toc` now validates **both** bounds at its entry point and raises `ValueError` for `max_notes < 1` or `max_level < 1` — so the misleading `{notes: [], truncated: True}` (and the analogous all-headings-filtered `max_level=0` case) are now explicit errors. Added regression tests (`test_get_toc_rejects_max_notes_below_one`, `test_get_toc_rejects_max_level_below_one`) and documented the `>= 1` requirement in the manager and tool docstrings.

On the other notes: the cross-document heading dedup (`GROUP BY document_id, heading, heading_level`) is intentional and matches single-note `get_toc`; the vault-root block and the absence of a `max_notes` ceiling are deliberate (the latter consistent with `list_documents`). Tightening the `dict[str, Any]` payloads to `TypedDict`/dataclass is tracked separately in #779.

pvliesdonk:
Thanks — re the FTS-layer `max_notes` guard suggestion: declining intentionally. Input validation in this codebase lives at the manager/facet boundary, not in the private `FTSIndex` methods — e.g. `_validate_path` / `_validate_dir_path` are on `DocumentManager`, and `FTSIndex` trusts its callers (it never validates `path`, `limit`, etc. either). `DocumentManager.get_toc` now rejects `max_notes < 1` / `max_level < 1` before any FTS call, so the only production path is covered. Duplicating the guard inside `get_subtree_toc` would be inconsistent with that layering for a private method with a single guarded caller. Happy to revisit if `get_subtree_toc` ever gains an unguarded caller.

All other points in this review are confirmations — no further changes. CI, type-check, Vale, and 100% patch coverage are green on b6b1edd.

PR #801, opened 2026-07-01
feat: surface skipped files via get_index_status (#775)

### LINKED ISSUE #775: Indexer silently skips notes with invalid YAML frontmatter — not surfaced via MCP API
## Summary

A note whose YAML frontmatter fails to parse is dropped from the index, but the skip is reported only as a `docker` WARNING and is not discoverable through any MCP tool — so a parse-skipped note is indistinguishable from one that simply has not synced yet.

## Observed

While writing several Markdown notes into the vault, two were absent from the index. The MCP-side diagnostics gave no indication why:

- `get_index_status` → `status: "queryable"`, `dirty_paths: 0`, `last_reindex_error: null` (no error reported)
- `read` on the path → `ValueError: ... document is not indexed or does not exist`
- `get_broken_links` → the missing note surfaced only *indirectly*, as broken inbound links from other notes that referenced it

The actual reason appeared only in `docker logs ai-obsidian-vault-mcp`:

```
WARNING markdown_vault_mcp.managers.index: reindex: skipping <path>.md — parse error (mapping values are not allowed in this context
  File ".../managers/index.py", line 382, in reindex
    note = parse_note(abs_path, self._source_dir, self._chunk_strategy)
  File ".../scanner.py", line 962, in parse_note
    post = frontmatter.loads(text)
  ...
yaml.scanner.ScannerError: mapping values are not allowed in this context
```

The reindex summary log line also carries a `skipped` counter (`reindex: N added, …, 0 skipped`) that is not exposed through the API.

Cause of the parse error itself — `[verified: reproduced + docker logs]` — was an unquoted colon-space (`: `) in a frontmatter scalar (e.g. a title `Foo — Style: Clarity and Grace`). That part is understood and is on the authoring side. **This issue is about the silent, hard-to-discover skip, not the YAML.**

## Expected

A file skipped because of a frontmatter/parse error should be discoverable through the MCP API (without reading container logs), so a dropped note can be told apart from an unsynced one.

## Context

- Observed against the deployed `ai-obsidian-vault-mcp` container, 2026-06-29.
- Internal locations from the traceback: `managers/index.py:382` (`reindex` → `parse_note`) and `scanner.py:962` (`parse_note` → `frontmatter.loads`).
- Exact version/commit `[unverified]` — diagnosed from runtime logs, not from a checked-out source tree.
- Workaround that confirmed the diagnosis: writing the file through the `write` tool re-emits valid (quoted) YAML, after which it indexes normally.

## Scope boundary

Not about correcting the specific notes' YAML (already fixed), and not about preventing authors from writing invalid frontmatter. Only about surfacing parse-skips through the API rather than swallowing them.

## Open questions

- `[unverified]`: whether other parse/scan failures (non-YAML, e.g. encoding errors) are skipped through the same path and would be hidden the same way.
- `[unverified]`: whether the existing `skipped` counter in the reindex summary already aggregates these and only needs exposing.

pvliesdonk:
## Summary

Makes deterministic, non-excluded index skips **discoverable through `get_index_status`**, so a note dropped from the index (invalid YAML, non-UTF-8, or missing a required frontmatter field) can be told apart from one that simply hasn't synced yet — without reading container logs.

Previously such a skip was only a container-log `WARNING`: `get_index_status` reported `queryable`/`dirty_paths: 0`, and `read` on the path raised `ValueError: document is not indexed`. Now the reason is recorded in tracker state and surfaced via a new `skipped_files` field.

Closes #775.

## What changed

- **New `SkippedFile` type + `SKIP_CATEGORIES`** (`types.py`) — frozen dataclass `{path, category, detail}`; the three legal categories (`parse_error`, `encoding_error`, `missing_frontmatter`) are pinned in one place and test-enforced.
- **`skip_reasons` map in the tracker state file** (`tracker.py`) — path → `{category, detail}`, maintained in lockstep as a strict subset of the existing `skipped` hash map (carried forward for unchanged skips, cleared when a path indexes or leaves disk). Back-compatible: a v2 `state.json` without the key loads as `{}`, no migration. Malformed persisted entries are dropped with a `WARNING` so the non-blocking `get_index_status` read never raises.
- **Both index-build paths feed it** — `build_index` via a new `on_skip` callback on `scan_directory`; `reindex` inline via `_record_skip`. Exclude-pattern matches and transient `OSError` skips are intentionally **not** surfaced (the latter self-heals on the next scan).
- **Surfaced via `get_index_status.skipped_files`** — `IndexManager.skipped_files()` reads the map; the facet merges it (as `asdict` dicts) into the status response. Tool docstring, `docs/tools/index.md`, and `docs/design.md` updated.

## Test plan

- Full suite green (2479+ passing); ruff + mypy clean; patch coverage ≥ 80% (the four touched modules at ~100%); structural diff-gate 100%.
- New coverage: `SkippedFile` serialization + category enforcement; tracker `skip_reasons` round-trip / carry-forward / drop-on-index / malformed-value sanitization / back-compat; `scan_directory` `on_skip` per category (incl. the generic-exception branch) and not-for-excluded; both build paths across all three categories; carry-forward through `IndexManager.reindex`; and an end-to-end FastMCP `Client` test asserting `get_index_status.skipped_files`.

## Review

Built via subagent-driven development (5 tasks, per-task spec+quality review) plus an Opus whole-branch review (0 critical/important) and a 3-round local preflight review circus that converged clean.

**Known sub-threshold follow-ups (non-blocking, tracked separately):** a distinct category for genuinely-unexpected `parse_note` exceptions (vs. reusing `parse_error`), and a `WARNING` log when a `build_index` pass-2 hash re-read races a vanishing file and drops an already-surfaced reason (rare, self-healing).

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -356,9 +356,12 @@ for built-in names, or pass a custom instance.
 
 - **State file** (the JSON persistence layer for hash-based change detection):
   versioned format `{"version": 2, "indexed": {relative_path: sha256_hash},
-  "skipped": {relative_path: sha256_hash}}` as JSON (#665). The legacy flat
-  `{relative_path: sha256_hash}` format still loads (every entry is treated
-  as indexed), so upgrades need no migration step.
+  "skipped": {relative_path: sha256_hash}, "skip_reasons": {relative_path:
+  {"category": ..., "detail": ...}}}` as JSON (#665, #775). `skip_reasons`
+  is a strict subset of `skipped` covering the surfaced deterministic skips
+  (parse / encoding / missing-frontmatter); a version-2 file without it loads
+  it as `{}`. The legacy flat `{relative_path: sha256_hash}` format still
+  loads (every entry is treated as indexed), so upgrades need no migration.
 - **Default path**: `{source_dir}/.markdown_vault_mcp/state.json` (when
   `state_path=None`).
 - On `reindex()`: scan all files, compare hashes to stored state, re-parse and
--- a/docs/design.md
+++ b/docs/design.md
@@ -1801,15 +1801,20 @@ class ChangeTracker:
     def detect_changes(self, source_dir: Path,
                        glob_pattern: str = "**/*.md") -> ChangeSet: ...
     def update_state(self, notes: list[ParsedNote],
-                     skipped: dict[str, str] | None = None) -> None: ...
+                     skipped: dict[str, str] | None = None,
+                     skip_reasons: dict[str, dict[str, str]] | None = None) -> None: ...
+    def skip_reasons(self) -> dict[str, dict[str, str]]: ...
     def reset(self) -> None: ...
 ```
 
 `tracker.py` is entirely new code (no ifcraftcorpus equivalent). State file
-format (version 2, #665): `{"version": 2, "indexed": {"Journal/note.md":
-"sha256hex", ...}, "skipped": {"CLAUDE.md": "sha256hex", ...}}` as JSON.
-The legacy flat `{"Journal/note.md": "sha256hex", ...}` format loads with
-every entry treated as indexed.
+format (version 2, #665, #775): `{"version": 2, "indexed": {"Journal/note.md":
+"sha256hex", ...}, "skipped": {"CLAUDE.md": "sha256hex", ...}, "skip_reasons":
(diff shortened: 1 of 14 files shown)


pvliesdonk:
Sub-threshold review follow-ups (diagnostic fidelity) are tracked in #802 — non-blocking for this PR.

PR #826, opened 2026-07-02
fix: floor the semantic candidate pool so recall does not depend on limit

### LINKED ISSUE #820: Semantic search recall depends on the caller's limit: candidate-pool floor hides best-scoring documents
**Version:** 3.0.4.

`managers/search.py` `_semantic_search` sizes its candidate pool as `max(limit * (chunks_per_file + 4), 50)` before grouping by document. `VectorIndex.search` does a full linear scan and sort regardless, so the floor is a pure recall cap: when a few chunk-heavy documents fill the pool, a smaller document whose best chunk ranks just past it is absent from grouped results entirely, even when it is the best-scoring document for the query.

Observed on a real 6,916-chunk vault: for one query the true best document (cosine 0.685) was absent at `limit` 3-12 while the returned top-1 scored 0.495; it appeared only at `limit >= 13` (pool 78). Scores are also not comparable across limits, since the pool changes what gets grouped.

Unit repro: 50 chunks across two large documents at cosine ~0.75 plus one small target document at 0.7499 (global rank 51) — target absent at limit 3/5/8 on stock, present at every limit after raising the floor.

Fix incoming as a PR: floor the pool at 1000 (wider pool costs only the slice, not extra distance computation) + a regression test making document recall limit-independent.

michael-denyer:
Closes #820

## Summary

`managers/search.py` `_semantic_search` sizes its vector candidate pool as `max(limit * (chunks_per_file + 4), 50)` and then groups the returned chunks into documents. `VectorIndex.search` does a full linear scan and sort over every stored vector regardless, slicing to the requested limit afterwards, so the floor is a pure recall cap rather than a cost control.

At a small limit the floor of 50 chunks can be consumed entirely by many chunks from a few large documents, so a smaller document whose best chunk ranks just past the floor never enters the pool and is absent from the grouped results, even when it is the best-scoring document for the query. The same query at a larger limit widens the pool and the document reappears, so recall depends on the caller's `limit` rather than on relevance. Scores are also not comparable across limits, since the pool changes what gets grouped.

The fix raises the floor to 1000 so the best-scoring documents are reachable at any limit in practice. Because the underlying search already scans and sorts every vector, the wider pool costs only the slice, not additional distance computations.

## Evidence

From the issue, observed on a real 6,916-chunk vault: for one query the true best document (cosine 0.685) was absent at `limit` 3 through 12 while the returned top-1 scored 0.495; it appeared only at `limit >= 13` (pool 78).

## Test plan

Added `test_semantic_recall_of_best_document_is_independent_of_limit` in `tests/test_managers_search.py`. It builds a three-document store where a small TARGET document's only chunk sits at global cosine rank 51, behind 50 chunks from two large documents, and asserts TARGET is returned at `limit=5`. The test drives chunk ranks exactly by writing document vectors onto the index and embedding every query to a fixed unit vector, so the cosine of each chunk is its own first coordinate.

- Full suite: 2522 passed, 2 skipped (playwright browser test and an empty domain-wizard placeholder; both environmental).
- Patch coverage: 100% on the changed line (diff-cover against `main`).

## Reviewer notes

The change is scoped to `_semantic_search` only. The floor-50 pools in `_keyword_search`, `_hybrid_search`, the snippet re-query, and the attachment path are intentionally left unchanged; the recall pathology and the "scan is already full" argument are specific to the semantic vector path. The test reaches into `VectorIndex._embeddings` and `SearchManager._vectors` to fix ranks precisely, matching the existing pattern in this test module, and asserts on returned paths rather than internal calls.

### DIFF
--- a/src/markdown_vault_mcp/managers/search.py
+++ b/src/markdown_vault_mcp/managers/search.py
@@ -56,6 +56,13 @@ logger = logging.getLogger(__name__)
 # RRF constant — standard value recommended in the original paper.
 _RRF_K = 60
 
+# Floor for the vector-search candidate pool, applied by both the semantic and
+# the hybrid path. Vector search does a full linear scan and sort regardless of
+# limit (see VectorIndex.search), so a wide floor only costs the final slice,
+# not extra distance work. It keeps recall of the best-scoring documents from
+# depending on how large a limit the caller passed.
+_SEMANTIC_CANDIDATE_FLOOR = 1000
+
 # Regex for extracting query tokens (alphanumeric sequences).
 _QUERY_TOKEN_RE = _re.compile(r"[A-Za-z0-9]+")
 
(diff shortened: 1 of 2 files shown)


pvliesdonk:
@claude review

michael-denyer:
Adopted the hybrid finding in 9d919ea: module-level `_SEMANTIC_CANDIDATE_FLOOR = 1000` now feeds both `_semantic_search` and the vector channel of `_hybrid_search`, while the FTS channel keeps its floor of 50 (BM25 does real per-candidate work, so the channels size independently; the snippet re-query stays on the FTS limit). Added `test_hybrid_recall_of_best_document_is_independent_of_limit` on the same pinned-embeddings technique — it fails against the previous commit and passes with the fix. Full suite 2,523 passing, ruff/mypy clean.

PR #827, opened 2026-07-02
fix: retry transient descriptor-exhaustion when hashing during a scan

### LINKED ISSUE #821: detect_changes silently drops files that fail to hash; retry transient EMFILE/ENFILE
**Version:** 3.0.4. **Framing: hardening, not a field-observed bug** — demonstrated under fault injection only.

`tracker.py` `detect_changes` hashes every discovered file and, on any `OSError`, logs a WARNING and drops the path from the disk snapshot. A brand-new file that fails to hash is then reported as neither added, modified, unchanged, nor deleted; it never reaches the FTS upsert and the reindex summary logs a self-consistent "0 added, 0 modified, 0 deleted, N unchanged", so the drop is invisible at INFO.

Demonstrated end to end with an injected-EMFILE shim against a live http server: the watcher's reindex silently omitted the new file while logging a clean summary. Not observed in the wild on my deployment (fd pressure never materialized); filing because the drop path is plainly reachable and unobservable.

This deliberately preserves the documented "transient skips are not recorded, retry next scan" contract: the fix incoming as a PR retries only `EMFILE`/`ENFILE` with short backoff within the scan, leaves every other `OSError` on its current first-failure path, and logs exhausted retries at ERROR with a distinct event name. Related to the skipped-files-visibility direction of #775/#802 — a per-scan "unreadable" counter on `ReindexResult` would complement this but changes a public schema, so it is left out.

michael-denyer:
Closes #821

## Summary

Framing: hardening, not a field-observed bug. The drop path is demonstrated under fault injection only, not observed in the wild on my deployment (descriptor pressure never materialized there). Filing because the drop path is plainly reachable and unobservable at the default log level.

`tracker.py` `detect_changes` hashes every discovered file and, on any `OSError`, logs a WARNING and drops the path from the disk snapshot. A brand-new file that fails to hash is then reported as neither added, modified, unchanged, nor deleted; it never reaches the FTS upsert and the reindex summary logs a self-consistent "0 added, 0 modified, 0 deleted, N unchanged", so the drop is invisible at INFO.

The fix retries the hash read a few times with a short backoff, but only for the transient descriptor-exhaustion errnos (`EMFILE`/`ENFILE`), so a momentary spike no longer drops the file for the whole scan. Every other `OSError` (`EACCES`, `ENOENT`, ...) still propagates on the first failure and is dropped with the existing warning, preserving the documented "transient skips are not recorded, retry next scan" contract. When retries are exhausted the failure is logged at ERROR with a distinct event name so a persistently unreadable file is visible instead of lost among a busy watcher's INFO summaries. This mirrors the existing `_retry_on_sqlite_locked` discriminate-then-retry idiom.

## Evidence

Demonstrated end to end with an injected-EMFILE shim against a live http server: the watcher's reindex silently omitted the new file while logging a clean summary at INFO. Retries: three attempts with backoffs of 0.05s, 0.1s, 0.2s.

## Test plan

Added two tests in `tests/test_tracker.py`:

- `test_transient_fd_exhaustion_is_retried_within_the_scan`: a single EMFILE on the first hash attempt recovers on retry, and the file is reported as added rather than dropped.
- `test_persistent_fd_exhaustion_is_dropped_and_logged_at_error`: a persistent EMFILE that outlasts every retry is dropped for the scan (nothing can be read) and surfaced at ERROR with the file name.

Both assert on the returned `ChangeSet` and on log records rather than internal state.

- Full suite: 2523 passed, 2 skipped (playwright browser test and an empty domain-wizard placeholder; both environmental).
- Patch coverage: 100% on the changed lines (diff-cover against `main`).

## Reviewer notes

Deliberately kept out of scope: a per-scan "unreadable" counter on `ReindexResult`. It would complement this and align with the skipped-files-visibility direction of #775/#802, but it changes a public schema, so it belongs in its own change. The retry is bounded (three short sleeps) so a persistently unreadable file adds at most ~0.35s before being dropped and logged.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -378,8 +378,13 @@ for built-in names, or pass a custom instance.
   `ReindexResult` instead of `added`; it is re-evaluated (and indexed, if it
   gained valid frontmatter) only when its content hash changes. Transient
   `OSError` skips are deliberately not recorded, so those files retry on
-  every scan. A skipped file deleted from disk is dropped silently; it was
-  never indexed, so it is not counted as `deleted`.
+  every scan. Descriptor-exhaustion errors (`EMFILE`/`ENFILE`) additionally
+  get up to three in-scan retries with a short backoff (0.05 s, 0.1 s,
+  0.2 s) before the file is dropped for that scan, and exhausted retries
+  log at `ERROR` (`hash_read_failed_after_retry`); every other `OSError`
+  keeps the single-attempt `WARNING` path. A skipped file deleted from disk
+  is dropped silently; it was never indexed, so it is not counted as
+  `deleted`.
   The surfaced deterministic skips (parse / encoding / missing-frontmatter /
   internal-error, but not exclude-pattern or transient `OSError`) are also
   recorded with a `{category, detail}` reason in the state file's
--- a/src/markdown_vault_mcp/tracker.py
+++ b/src/markdown_vault_mcp/tracker.py
@@ -465,12 +465,16 @@ class ChangeTracker:
             OSError: The last error if a transient failure outlasts every
                 retry, or immediately for any non-transient ``OSError``.
         """
-        for sleep_s in _HASH_RETRY_SLEEPS_S:
+        attempt = 0
+        while True:
             try:
                 return self._compute_hash(path)
             except OSError as exc:
                 if exc.errno not in _TRANSIENT_HASH_ERRNOS:
                     raise
+                if attempt >= len(_HASH_RETRY_SLEEPS_S):
+                    raise
+                sleep_s = _HASH_RETRY_SLEEPS_S[attempt]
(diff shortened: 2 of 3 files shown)


pvliesdonk:
@claude review

michael-denyer:
Both points addressed in 509662a. The design.md tracker contract now records the bounded EMFILE/ENFILE retry (three attempts, 0.05/0.1/0.2 s backoff) and the ERROR-level `hash_read_failed_after_retry` event on exhaustion, distinct from the single-attempt WARNING path for other OSError types. The retry helper is restructured to a single call site in a while-True loop keyed off the attempt count, matching the `_retry_on_sqlite_locked` idiom it cites. Tracker tests 38 passing, ruff and mypy clean on the touched files.

PR #828, opened 2026-07-02
feat: scoped file discovery and watches for large vaults (prune excluded subtrees, deliver dot-roots)

### LINKED ISSUE #822: File discovery descends excluded subtrees: minutes-scale scans and watcher storms on large vaults
**Version:** 3.0.4.

`tracker.detect_changes` and `IndexManager.build_index` glob `**/*.md` across the whole vault and apply `is_path_excluded` per file afterwards, so every excluded subtree (`node_modules`, `.venv`, `go`, `Library`, ...) is fully descended on every scan.

Measured on a home-directory vault (4,4xx documents, 23 exclude patterns): one `detect_changes` call took **5,043 s**; a second stock run sat 28 minutes wedged inside a single `opendir` syscall (stack-sampled) before being killed. With the file watcher enabled, any churn retriggers scans continuously, so effective save-to-searchable latency is minutes and the process scans all day.

With exclude-aware pruning (walk that skips a directory before descent only when the exclude patterns provably cover everything beneath it, per-file filter unchanged as the correctness layer): the same scan takes **3.2 s** discovering 4,429 files across 25,263 directories, and save-to-searchable drops to ~15 s end to end. Both measurements taken while the deployed service was scanning concurrently.

Fix incoming as a PR: shared `iter_markdown_files()` in `utils/fs.py` used by `detect_changes`, `build_index`, and `scan_directory`; prunes only the two pattern shapes in real use (anchored `PREFIX/**`, any-depth `**/NAME/**`), descends on anything else; symlink and unreadable-dir semantics preserved; includes a walker-vs-glob equality test.

### LINKED ISSUE #823: Watcher: single recursive watch on source_dir registers over the whole tree (macOS TCC prompts); dot-root events never delivered
**Version:** 3.0.4, macOS 26 (Tahoe), watchdog 6.0.0. Two coupled defects in `_file_watcher.py`.

**1. Whole-tree registration.** The watcher schedules one recursive observer on `source_dir`. With a home-directory vault on macOS this creates an FSEvents stream over every other app's data containers and produces repeated "would like to access data from other apps" consent prompts naming the python binary. Exclude patterns never narrow the registration. Note watchdog's `recursive=False` is emulated on macOS: a non-recursive watch still opens an OS-level recursive FSEvents stream (verified: a non-recursive parent watch received raw callbacks for a file three levels down before watchdog filtered them), so even a "floor" watch on source_dir re-registers the whole tree.

**2. Dot-root events dropped.** `_has_hidden_component` ignores every event whose path relative to source_dir contains any dot component. A vault whose content lives under a dot directory (e.g. `.claude/**` when the vault root is $HOME) never triggers a reindex from its own edits; changes are only picked up when unrelated churn fires a scan.

Fix incoming as a PR (builds on the pruning helper from #822): derive watch roots from source_dir's immediate children minus provably-excluded ones (94 scoped watches on the field vault, zero source_dir-rooted), evaluate hiddenness relative to the delivering watch root so deliberately watched dot-roots deliver while `.git`-style noise below any root is still ignored, and gate the optional source_dir floor watch behind `FILE_WATCHER_ROOT_FLOOR` (default on, preserving current behavior). Verified live: a file written under a watched dot-root becomes searchable without restart; red before the fix, green after. Related: #508 (symlink-farm layouts get no effective event coverage either way — watchdog delivers realpath-prefixed events that a root-relative filter drops; kept out of this PR).

michael-denyer:
Closes #822, closes #823

## Summary

Two coupled changes to file discovery and the file watcher on large vaults, plus the env var and wizard question that expose the new watcher knob.

**Discovery pruning (#822).** `tracker.detect_changes` and `IndexManager.build_index` globbed `**/*.md` across the whole vault and applied `is_path_excluded` per file afterwards, so every excluded subtree (`node_modules`, `.venv`, `go`, `Library`, ...) was fully descended on every scan. This PR adds a shared `iter_markdown_files()` in `utils/fs.py`: an `os.walk`-based walker that prunes a directory before descending when the exclude patterns provably cover everything beneath it. It prunes only the two pattern shapes in real use, anchored `PREFIX/**` and any-depth `**/NAME/**`, and descends on anything else, so the per-file `is_path_excluded` filter stays the sole correctness layer. It is used by `detect_changes`, `build_index`'s skipped-count walk, and `scan_directory` (which falls back to the raw glob for a non-default `glob_pattern`).

**Scoped watches (#823).** The watcher scheduled one recursive observer on `source_dir`. With a home-directory vault on macOS that registers an FSEvents stream over every other app's data container and produces repeated "would like to access data from other apps" prompts. This PR watches a derived set of roots instead. Each non-excluded immediate child directory is watched recursively, reusing `_dir_prune_rules`/`_should_prune_dir` from `utils/fs` as the single prune-rule source, and `source_dir` itself is watched non-recursively as a floor. It also fixes a dot-root bug: `_has_hidden_component` ignored every event whose path had any dot component, so a vault whose content lives under a dot directory never triggered a reindex from its own edits. Hiddenness is now evaluated relative to the watch root that delivered the event. The non-recursive `source_dir` floor watch is gated behind `FILE_WATCHER_ROOT_FLOOR` (default true); watchdog emulates `recursive=False` in Python but the OS-level FSEvents stream for the floor is still recursive on macOS, so deployments that need zero `source_dir`-rooted registration can set it false and accept that root-level files are then only picked up by scans.

## Evidence

From the issues, measured on a home-directory vault (4,4xx documents, 23 exclude patterns) while the deployed service was scanning concurrently:

- Discovery: one stock `detect_changes` call took 5,043 s; a second run sat ~28 minutes wedged inside a single `opendir` before being killed. With pruning the same scan takes 3.2 s discovering 4,429 files across 25,263 directories, and save-to-searchable drops to ~15 s end to end.
- Watches: the field vault derived 94 scoped watches with zero `source_dir`-rooted streams. A file written under a watched dot-root became searchable without restart (red before the fix, green after).

## Test plan

Commit 1 (pruning) adds `tests/test_utils_fs.py` covering `_dir_prune_rules`, `_should_prune_dir`, and a walker-vs-glob equality test over the 23 live exclude patterns, plus non-default-`glob_pattern` fallback tests for `scan_directory` and `detect_changes`.

Commit 2 (scoped watches) adds `_derive_watch_roots` tests (anchored and any-depth pruning, symlinked children, enumeration-failure fallback with `root_floor` on/off), `_is_hidden_relative_to_root` tests, a `start()` test that a root whose schedule raises is logged and skipped without aborting, `SyncConfig`/`ProjectConfig` env tests for `FILE_WATCHER_ROOT_FLOOR`, and the wizard question.

Commit 3 documents the env var and the scoped-watch behavior.

- Full suite: 2571 passed, 2 skipped (playwright browser test and an empty domain-wizard placeholder; both environmental).
- `mypy src/ tests/`: no issues in 155 source files.
- Structural-diff gate (`diff-quality` with `C901,PLR0911,PLR0912,PLR0913,PLR0915,S`): 100%, 0 violations over 1130 diff lines.
- Patch coverage: 99% (diff-cover against `main`). The one uncovered diff line is the `(+N more)` truncation in the watcher's start-log formatter, which only fires with more than 12 watch roots.

## Reviewer notes

Rebased onto current `main`; both features build on the shared prune helpers, so please read `utils/fs.py` first.

Two conflict resolutions worth a look:

- `scanner.py` import block: `main` had added `SkippedFile` and `Callable` (the skipped-files surfacing work) while this change adds `iter_markdown_files` and `Iterable`. Resolved by unioning both, so all four are imported and used.
- `tests/test_file_watcher.py`: the cherry-picked `FILE_WATCHER_ROOT_FLOOR` env tests imported `VaultConfig`, which `main` has renamed to `ProjectConfig`. Re-applied the tests against `ProjectConfig`, matching the sibling tests in the same file. `managers/index.py`'s skipped-count walk merged cleanly onto the newer `skipped_state`/`skip_reasons` recording, and because excluded files are now pruned during discovery they no longer enter `skipped_state`, matching the excluded-files-are-invisible contract.

On the wizard: `main` added `tests/test_config_wizard_drift.py`, which hard-fails if any `ServerConfig`/`ProjectConfig` setting is neither offered by the wizard nor listed in `_COVERED_BY_INFERENCE`. Adding `FILE_WATCHER_ROOT_FLOOR` to `SyncConfig` made that test require coverage, so rather than defer it behind an allowlist I added the wizard question (a bool under the existing "File watcher" advanced group, matching `file_watcher` and `file_watcher_debounce_s`) and dropped the `NOT_IN_WIZARD` deferral entry. `_COVERED_BY_INFERENCE` was not the right escape hatch since this setting has a real dedicated env var.


michael-denyer:
On the Docs Prose (Vale) failures: the ai-tells style landed on main between this PR's first and second CI runs, so the two runs judged different rules. After bb601db, this branch's added lines are clean against the current style (the one hit in our lines was a 'for example' connector in docs/configuration.md, now reworded). The remaining failures in the job log sit on lines this PR does not touch (README feature bullets, docs/design.md, and other pre-existing prose flagged by the new rules) — nothing in the workflow's docs backlog was linted before the style landed, so they surface on whichever PR runs next. Happy to handle those however you prefer: separate cleanup PR, or leave them for a docs pass on main.

PR #838, opened 2026-07-02
fix: never watch the vault's own state dir or .git (self-feedback reindex loop)

### LINKED ISSUE #830: fix: file-watcher self-feedback reindex loop — state dir & .git become recursive watch roots (#828 regression of #720/#721)
## Summary

With a default configuration (no `exclude_patterns`), the file watcher schedules a recursive watch on the vault's own state directory (`.markdown_vault_mcp/`) and on `.git/`, and writes into those directories re-trigger `reindex()`, producing a self-perpetuating reindex loop. This is a regression of the #720/#721 self-feedback-loop fix, introduced by the scoped-watch work in #828.

## Observed

At `main` (commit range `3f29919..5527f66`, PR #828 merged):

- `_derive_watch_roots(source_dir, exclude_patterns, ...)` promotes every non-pruned immediate child directory of `source_dir` to a **recursive** watch root, with pruning driven **only** by `exclude_patterns`. `exclude_patterns` defaults to `None` (`config_sections/indexing.py`), so nothing is pruned by default.
- The default state directory is `{source_dir}/.markdown_vault_mcp/`, created at lifespan startup before the watcher starts, so it becomes its own recursive watch root. `.git/` likewise on a git vault.
- `_VaultEventHandler` now evaluates hiddenness **relative to the delivering watch root** (`_is_hidden_relative_to_root`), not relative to `source_dir` as at base `3f29919`. So a write to `.markdown_vault_mcp/state.json` delivered under `watch_root=.../.markdown_vault_mcp` yields `rel="state.json"` — no hidden component → **not ignored** → schedules a reindex.
- `reindex()` writes tracker state **unconditionally** every run (`IndexManager.reindex` → `_tracker.update_state(...)` → `_save_state` writes `.markdown_vault_mcp/state.json`), so the state write is itself a mutating event under the state-dir watch root → another reindex after the debounce → repeats.

[verified] Reproduced in-process: `_derive_watch_roots(vault, None)` returns recursive roots including `.git` and `.markdown_vault_mcp`, and `_is_hidden_relative_to_root(".../.markdown_vault_mcp/state.json", ".../.markdown_vault_mcp")` returns `False` (not ignored).

## Expected

Writes into the vault's own state/index/embeddings directory and into `.git/` do not trigger reindexing; the watcher does not enter an idle reindex loop in a default configuration. This is the invariant the `VaultFileWatcher` class docstring still asserts ("Changes inside hidden directories (e.g. `.git/`, `.markdown_vault_mcp/`) are silently ignored so ... state-file writes do not trigger spurious reindexes").

## Context

- Surfaced by a local multi-lens review of the merged #824/#826/#827/#828 range; two independent lenses (git-history and a general code review) reached this with the repro above.
- Existing tests cover `.git` nested *under* `.claude` and a floor handler bound to `source_dir`, but none exercise a handler bound to a **top-level** `.markdown_vault_mcp`/`.git` watch root, which is the regressing path.

## Scope boundary

About the watcher scheduling watches on the state dir / `.git` and the resulting loop. Not about the discovery-pruning change in the same PR (that path keeps `is_path_excluded` as its correctness layer).

## Deliverables

- Fix so the state directory and `.git` are not active watch roots (while preserving deliberate dot-root delivery such as `.claude`).
- Regression test: a mutating event under a top-level `.markdown_vault_mcp`/`.git` watch root must not schedule a reindex.

pvliesdonk:
Closes #830.

## Problem

#828's scoped-watch change (`_derive_watch_roots`) promotes every non-pruned immediate child of `source_dir` to a **recursive** watch root, pruned only by `exclude_patterns` (which defaults to `None`), and `_VaultEventHandler` now judges hiddenness **relative to the delivering watch root**. So the default state dir `.markdown_vault_mcp/` and `.git/` each became their own recursive watch root, and a write to `.markdown_vault_mcp/state.json` — which `reindex()` performs **unconditionally** on every run — was no longer seen as hidden.

That closes a self-feedback reindex loop (a regression of the #720/#721 fix), reachable in a default configuration, plus reindex storms on `.git` activity (e.g. the local-git-write auto-commit path).

Verified in-process: `_derive_watch_roots(vault, None)` returned recursive roots including `.git` and `.markdown_vault_mcp`, and `_is_hidden_relative_to_root(".../.markdown_vault_mcp/state.json", ".../.markdown_vault_mcp")` returned `False`.

## Fix

`_derive_watch_roots` gains an `internal_dirs` parameter and skips any top-level child that **is or contains** one of them; `_server_deps` passes the vault's state / index / embeddings dirs and `.git`. User dot-roots (`.claude/`) are still watched recursively, preserving the #558/#828 behavior. Skipping a subtree that *contains* an internal dir (rather than only an exact match) guarantees the loop can never form even if the state dir is configured beneath a top-level directory.

## Tests

- `test_derive_watch_roots_skips_internal_state_and_git_dirs` — state dir + `.git` excluded, `.claude` retained.
- `test_derive_watch_roots_skips_top_level_dir_containing_internal_dir` — a dir containing a nested state dir is skipped.
- `test_start_does_not_schedule_watch_on_internal_state_dir` — exercises the `start()` wiring with a recording observer.

Full `tests/test_file_watcher.py` (61) green; ruff + mypy clean.

## Provenance

Surfaced by a local multi-lens review of the merged #824/#826/#827/#828 range; two independent lenses (git-history + general code review) reached this with the repro above.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -31,6 +31,12 @@ variables (`MARKDOWN_VAULT_MCP_TRANSPORT`, `MARKDOWN_VAULT_MCP_HOST`,
 | `MARKDOWN_VAULT_MCP_TEMPLATES_FOLDER` | string | `_templates` | No | Relative folder path used by the `create_from_template` prompt to discover/read template files |
 | `MARKDOWN_VAULT_MCP_PROMPTS_FOLDER` | path | (none) | No | Path to a directory of `.md` prompt files that extend or override built-in prompts. Each declared argument name must be a plain Python identifier (letters, digits, underscore; not a keyword); a prompt with a non-conforming argument name is skipped with a logged warning. |
 
+The file watcher never watches the directories that hold `INDEX_PATH`,
+`EMBEDDINGS_PATH`, and `STATE_PATH` (or `.git`), so the writes the server makes
+while indexing do not trigger it again. If you place one of these paths inside a
+content directory, that whole top-level directory stops being watched, so keep
+them at the vault root or outside the vault to keep sibling content watched live.
+
 ## Index Build Timeout
 
 ### `MARKDOWN_VAULT_MCP_BUILD_TIMEOUT_S`
--- a/src/markdown_vault_mcp/_file_watcher.py
+++ b/src/markdown_vault_mcp/_file_watcher.py
@@ -123,20 +123,69 @@ class _WatchRoot:
     recursive: bool
 
 
+def _resolve_internal_dirs(internal_dirs: Sequence[Path]) -> list[Path]:
+    """Resolve *internal_dirs*, dropping any that cannot be resolved.
+
+    Resolving up front lets :func:`_contains_internal_dir` compare against
+    realpaths, so a symlinked state dir still matches its watch-root candidate.
+
+    Args:
+        internal_dirs: Vault-internal directories to protect from watching.
+
+    Returns:
+        The resolvable directories as resolved paths.
+    """
+    resolved: list[Path] = []
+    for d in internal_dirs:
+        try:
+            resolved.append(d.resolve())
+        except OSError:
+            continue
+    return resolved
+
+
+def _contains_internal_dir(child: Path, internal_dirs: frozenset[Path]) -> bool:
+    """Return ``True`` if *child* is, or contains, a vault-internal write dir.
+
+    ``internal_dirs`` must already be resolved. A recursive watch on a directory
+    that is or contains the state / index / embeddings dir (or ``.git``) would
+    deliver the writes ``reindex`` makes into those paths, re-triggering itself;
+    such a directory is skipped entirely so the loop cannot form (#830).
+
+    Args:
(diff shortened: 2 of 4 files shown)


pvliesdonk:
Addressed in a39d690:

- **Coverage gap (gate #4):** added `test_lifespan_passes_configured_internal_dirs_to_watcher`, which drives `make_vault_lifespan()` with an explicit `state_path`/`index_path`/`embeddings_path` and asserts each parent dir plus `.git` reaches the watcher's `internal_dirs` — covering the previously-untested non-`None` branches in `_server_deps`.
- **Nested-state caveat:** documented in `docs/configuration.md` (the state/index/embeddings dirs and `.git` are never watched; nesting one under a content directory stops that whole top-level directory being watched).
- **Debug log on skip:** left out for now — the "why isn't this dir watched" case is now covered by the config note, and adding a per-candidate debug line to the hot start path felt like more noise than value. Happy to add it if you'd prefer.

PR #840, opened 2026-07-02
fix: keep an already-indexed file that fails to re-hash instead of purging it

### LINKED ISSUE #831: fix: transient hash-read failure purges an already-indexed file from the index
## Summary

When `tracker.detect_changes` fails to hash an **already-indexed** file (transient EMFILE/ENFILE exhausting the retries added in #827, or any other `OSError`), the file is dropped from the disk snapshot and is then computed as `deleted`, causing the downstream reindex to purge that live document from FTS and its embedding until a later scan re-hashes it. The #827 ERROR log and adjacent comment describe the outcome as benign ("dropped from this scan and retried on the next one"), which is only true for a brand-new file.

## Observed

`tracker.py` `detect_changes`: on hash failure the path is not added to `disk_state` and the loop `continue`s. `deleted` is computed as `rel_path in indexed_state and rel_path not in disk_state and not excluded`, so a previously-indexed file that fails to hash lands in `deleted`. [verified: read `detect_changes` and the deleted-set computation]

## Expected

A transient (or otherwise non-permanent) failure to re-hash an already-indexed file leaves that document in the index (treated as unchanged this scan, re-hashed next scan), rather than purging it; and operator-facing messages describe the actual impact.

## Context

The underlying drop-then-look-deleted behavior predates #827 (the pre-#827 `except OSError` did the same for all errnos). #827 added the retry and an ERROR message on exhaustion whose wording understates the indexed-file case.

Cause: [verified] behavior traced in `detect_changes`; the severity mislabel is in the #827 ERROR string/comment.

## Deliverables

- Carry an already-indexed file's prior hash forward into `disk_state` when its re-hash fails (so it is neither reported modified nor deleted), or an equivalent that keeps the document indexed.
- Correct the exhausted-retry ERROR message/comment to state the real impact.
- Test: an indexed file that fails to hash is not reported deleted.

pvliesdonk:
Closes #831.

## Problem

`ChangeTracker.detect_changes` dropped **any** file that failed to hash from the disk snapshot. For a file already in the indexed state, that routed it into the `deleted` list, and the downstream reindex purged the live document from FTS and its embedding — on a merely transient (`EMFILE`) or a still-present-but-unreadable (`EACCES`) read error. #827's retry lowered the probability for `EMFILE`, but the exhaustion branch and every non-transient `OSError` kept the wrong outcome. Surfaced by the multi-lens review of the #824–#828 range (silent-failure lens).

## Fix

When hashing fails for a path already in `indexed_state`, carry its prior hash forward into `disk_state` so it counts as **unchanged** for that scan and is re-hashed on the next one. Only a newly discovered (unindexed) file that fails to hash is dropped (unchanged behavior). A genuinely deleted file is absent from discovery on the next scan and reported `deleted` then.

## Tests

- `test_indexed_file_failing_to_hash_is_kept_not_deleted` — an indexed file that raises `EACCES` on re-hash is not in `deleted`, not `modified`, counted `unchanged` (fails on `main`: the file is purged).
- `test_non_transient_errno_is_not_retried` — asserts a non-transient errno is attempted exactly once, locking in the discriminate-then-retry contract (part of #836).

Full `tests/test_tracker.py` (41) green; ruff + mypy clean; structural-diff gate 100%.

**Note:** the matching `docs/design.md` contract update rides with the design.md-relocation PR (design.md currently fails the local Vale hook on pre-existing content and is being moved under `docs/design/`).

### DIFF
--- a/src/markdown_vault_mcp/tracker.py
+++ b/src/markdown_vault_mcp/tracker.py
@@ -152,11 +152,30 @@ class ChangeTracker:
             try:
                 content_hash = self._hash_with_retry(abs_path)
             except OSError as exc:
+                prior = indexed_state.get(rel_str)
+                if prior is not None:
+                    # Already indexed: a failed re-hash must NOT drop it, or the
+                    # file would be absent from disk_state and look "deleted"
+                    # below, purging a live document from FTS and its embedding
+                    # on a merely-unreadable-right-now file (transient EMFILE, or
+                    # a still-present EACCES). Carry the prior hash forward so it
+                    # counts as unchanged this scan and is re-hashed next scan
+                    # (#831). A genuinely deleted file is absent from discovery
+                    # on the next scan and reported deleted then.
+                    disk_state[rel_str] = prior
+                    logger.warning(
+                        "hash_read_failed_keeping_indexed path=%s errno=%s: %s",
+                        abs_path,
+                        exc.errno,
+                        exc,
+                    )
+                    continue
                 if exc.errno in _TRANSIENT_HASH_ERRNOS:
-                    # Retries were exhausted: file-descriptor pressure outlasted
-                    # the retry window. The file is dropped from this scan and
-                    # retried on the next one, but surface it at ERROR so it is
-                    # not lost among a busy watcher's INFO reindex summaries.
+                    # A brand-new (unindexed) file that failed even after
+                    # retries: nothing to carry forward, so it is dropped from
+                    # this scan and retried on the next one. Surface it at ERROR
+                    # so it is not lost among a busy watcher's INFO reindex
+                    # summaries.
                     logger.error(
                         "hash_read_failed_after_retry path=%s errno=%s: %s",
                         abs_path,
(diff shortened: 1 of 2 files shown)


pvliesdonk:
Thanks — addressed in 978e645:

- **caplog nit:** the kept-indexed test now asserts the `hash_read_failed_keeping_indexed` WARNING is emitted for `keep.md`.
- **Severity change (ERROR → WARNING for an exhausted-retry *indexed* file):** intentional. Once the prior hash is carried forward the document stays indexed and no data is at risk, so WARNING is the honest level; the ERROR path (`hash_read_failed_after_retry`) is retained for a *newly discovered* file that exhausts retries, where the file really is dropped for the scan.
- **design.md contract gap:** resolved in #841, which relocates `docs/design.md` under the Vale-excluded `docs/design/` subtree and adds the already-indexed carry-forward paragraph there (it couldn't be edited here without tripping the local Vale hook on pre-existing content).

PR #841, opened 2026-07-02
chore: relocate docs/design.md into docs/design/ (resolves template#234 downstream)

### LINKED ISSUE #837: chore(vale): pre-commit exclude drifted out of lockstep with CI glob (superpowers vs superpowers,design,decisions)
## Summary

The Vale exclusion scope declared in the pre-commit `- id: vale` hook has drifted out of lockstep with the CI `vale_flags` glob: pre-commit excludes only `^docs/superpowers/`, while CI excludes `docs/{superpowers,design,decisions}/**`. The generated CLAUDE.md documents them as required to stay in lockstep.

## Observed

- `.pre-commit-config.yaml` `- id: vale`: `exclude: ^docs/superpowers/`.
- `.github/workflows/ci.yml`: `vale_flags: "--glob=!docs/{superpowers,design,decisions}/**"`.
- CLAUDE.md (Documentation Conventions) states the pre-commit exclude is `^docs/(superpowers|design|decisions)/` and that the CI glob and pre-commit exclude "must stay in lockstep."

Effect: a local `pre-commit run` Vale-lints `docs/design/**` and `docs/decisions/**` that CI does not, so pre-commit and CI disagree on which docs are linted. [verified: read both files + CLAUDE.md]

## Expected

The pre-commit Vale exclude matches the CI glob's subtree set.

## Context

`.pre-commit-config.yaml` is `_skip_if_exists` (write-once, not re-rendered by `copier update`), so when the template widened the exclusion set the local pre-commit hook was not updated. Distinct from the upstream question of whether the top-level `docs/design.md` file should be excluded at all (filed upstream as pvliesdonk/fastmcp-server-template#234).

## Deliverables

- Update the pre-commit `- id: vale` `exclude:` regex to match the CI glob's current subtree set.

pvliesdonk:
Resolves the downstream side of pvliesdonk/fastmcp-server-template#234. Closes #837.

## Why

`docs/design.md` is the project's **authoritative internal design spec**, but at the top level it sat outside the Vale-exclusion subtrees (`docs/{superpowers,design,decisions}/**`), so ordinary spec vocabulary tripped error-level Vale rules on every PR that edited it (e.g. #827's `additionally` / `backoff`). mkdocs already excluded it from the published site (`exclude_docs`); only the linters lagged.

## What

- `git mv docs/design.md docs/design/design.md` — into the already-excluded `docs/design/` subtree (the convention the generated CLAUDE.md documents for internal dev-docs).
- Repointed every **live** reference: `CLAUDE.md` (×3), `examples/{para,zettelkasten}/README.md`, `docs/deployment.md`, `docs/guides/{para,zettelkasten}.md`, and four `src/` docstrings (`facets/__init__.py`, `fts_index.py` ×2, `vault.py`). Fixed one now-`../` relative link inside the spec. Historical `docs/superpowers/plans/**` records are left as-is.
- `mkdocs.yml`: dropped the now-redundant top-level `design.md` from `exclude_docs` (`design/**` covers it).
- Widened the pre-commit `- id: vale` exclude to `^docs/(superpowers|design|decisions)/` to match the CI glob (it had drifted to `superpowers`-only because `.pre-commit-config.yaml` is `_skip_if_exists`) — **this is #837**.
- Carries the #831 already-indexed carry-forward contract note in the spec, now that design.md is editable without a Vale failure.

## Verification

ruff + mypy clean on touched `src/`; `tests/test_tracker.py` green; the local pre-commit `vale` hook now passes (the relocated spec is excluded). Chosen over changing the template glob because Vale honours only a single `--glob` and the template's convention is that internal specs live under `docs/design/`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -117,9 +117,10 @@ repos:
         # the packs aren't populated, Vale prints E201 config errors and
         # exits 0 — silent pass.
         # Upstream hook has `types: [text]`; this `files:` scopes to docs/.
-        # Together: any text file under docs/ except the working-specs
-        # subtree. Matches the CI step's `vale ... --glob='!docs/superpowers/**' docs/`
+        # Together: any text file under docs/ except the internal dev-doc
+        # subtrees. Matches the CI step's
+        # `vale ... --glob='!docs/{superpowers,design,decisions}/**' docs/`
         # scope exactly (both lint markdown + asciidoc + rst + plain
         # text — every format Vale supports — under docs/).
         files: ^docs/
-        exclude: ^docs/superpowers/
+        exclude: ^docs/(superpowers|design|decisions)/
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -105,8 +105,8 @@ repos:
         files: ^(docs/|\.vale\.ini$)
         # Align trigger scope with the `vale` lint hook below — no point
         # checking sync state on commits that only touch the excluded
-        # working-specs subtree.
-        exclude: ^docs/superpowers/
+        # internal dev-doc subtrees.
+        exclude: ^docs/(superpowers|design|decisions)/
 
   - repo: https://github.com/vale-cli/vale
     rev: v3.14.2
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -4,7 +4,7 @@ Generic markdown vault MCP server with FTS5 + semantic search, frontmatter-aware
 
 ## Design
 <!-- DOMAIN-START -->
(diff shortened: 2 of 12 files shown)


pvliesdonk:
Both addressed in 936ed31:

1. **`vale-sync` exclude drift:** widened it to `^docs/(superpowers|design|decisions)/` to match the `vale` lint hook — good catch, that's the same class of drift this PR fixes.
2. **Forward-referencing spec paragraph:** you're right — the carry-forward paragraph documents #840's behavior, not anything in this PR's diff. Removed it, so the relocated spec now matches `main`'s content exactly (plus the one `../guides/…` relative-link fix). The #831 contract note will land in a small follow-up once this move merges and unblocks editing `docs/design/design.md`.

PR #865, opened 2026-07-05
fix: inherit host neutrals, keep Paper terracotta accent (drop blue info alias)

### LINKED ISSUE #861: Graph view leaks host (Claude) theme colors — pin graph to Paper palette
### Summary
The Graph view inherits Claude's host theme instead of the Paper palette. The focus pill, semantic edges, and highlights come out in Claude blue, and other host styles bleed into the canvas. The graph should render in the Paper identity (terracotta accent, warm neutrals) in both light and dark — not the host theme.

### Root cause
The Paper tokens the graph reads are aliased to host variables, e.g.:

```css
--accent:     var(--color-text-info);      /* → Claude blue */
--panel:      var(--color-background-primary);
--ink:        var(--color-text-primary);
--border:     var(--color-border-primary);
--accent-ink: var(--color-text-inverse);
```

Because the graph paints from these, whatever the host pushes wins — so `--accent` resolves to Claude's blue and the canvas stops looking like Paper.

### Fix — pin the graph palette to literal Paper values (no host indirection)
In `getColors()`, resolve to the Paper hexes directly (per theme) instead of the host `--color-*` tokens. Keep the probe-based resolution, just point the fallbacks at literals and drop the host aliases for the graph:

```js
function getColors() {
  const dark = document.documentElement.getAttribute('data-theme') === 'dark';
  return dark ? {
    panel:'#252118', panel2:'#2e281d', ink:'#ece4d4', ink2:'#bcb29c',
    muted:'#8f8571', edge:'#4c4230', accent:'#d68a5f', accentInk:'#241a12',
    accentSoft:'#39291d', border:'#39311f',
  } : {
    panel:'#fbf8f1', panel2:'#f2ecdf', ink:'#2a2620', ink2:'#5c5446',
    muted:'#928975', edge:'#d3c7ad', accent:'#b25a35', accentInk:'#ffffff',
    accentSoft:'#f3e4d8', border:'#e4dbc9',
  };
}
```

pvliesdonk:
## Summary

Closes #861.

The SPA (Vault Explorer MCP App) aliased its accent to the host's **semantic info** slot:

```css
--accent: var(--color-text-info);   /* host info state = blue */
```

`--color-text-info` is the host's semantic *info* color (blue), not a UI accent — so this alias injected Claude's blue into links, active tabs, the graph focus node, semantic edges, and buttons. That blue appears nowhere else in the client; **that** was the clash (#861's root-cause line), not blending itself.

This decouples the two layers:

- **Neutrals inherit the host** — `--panel/-2`, `--ink/-2`, `--muted`, `--border/-2`, `--edge` now resolve to `var(--color-*, <Paper fallback>)`, expanded to also pick up `--color-text-tertiary` and `--color-border-secondary` for a fuller native blend (light + dark).
- **Accent is pure Paper terracotta literals** — `--accent`/`--accent-ink`/`--accent-soft`, never aliased to a host semantic slot. Per the ext-apps SDK convention the host supplies only the neutral/semantic ramp; `--color-accent` is the app's own token (see the SDK reference themes, `examples/basic-server-*/src/global.css`, where `--color-accent` is app-defined "customize for your brand"). So Paper owns the accent.
- **`--warn` → `var(--color-text-danger, …)`** — a genuine semantic state (the "missing note" badge).
- Repointed `.action-btn` / `.action-btn--secondary` and the JS error screen off the purged `--color-text-info`/`--color-text-inverse` slots; updated a now-stale `getColors()` comment in `views/graph.js`.

The graph canvas follows automatically via the existing probe (`getColors()` in `views/graph.js`): focus node + semantic edges render terracotta, surfaces/edges take host/Paper neutrals.

## Relationship to #861

#861 proposed pinning the graph's **entire** palette (neutrals included) to Paper literals. This PR instead keeps the graph's neutrals **host-inherited** and pins only the accent to Paper — a deliberate, newer design decision (native blend: surfaces match the host client, terracotta identity stays). #861's actual reported symptom (blue focus pill / semantic edges / highlights) is resolved either way; the "fully-Paper graph" mechanism is intentionally not adopted.

## Not in scope

- **#859 (mobile display-mode/sizing)** is a separate subsystem (already handled by #860, in `main`); this change touches only the color/token layer.
- **Graph-canvas color resolution** remains covered by the `getColors()` probe (older Electron/Chromium can't resolve `var()` indirection on canvas), not by these token tests — unchanged by this PR.

## Tests

`tests/test_mcp_apps_foundation.py`: new `test_accent_is_paper_literal_not_host_semantic` (pins all three accent tokens as literals in both themes; asserts the `--color-text-info`/`--color-text-inverse` class is purged everywhere) and `test_neutrals_inherit_host_with_paper_fallback` (pins every neutral + `--warn` as host-inherited with its theme-specific fallback in both the `:root` and `[data-theme="dark"]` blocks). `tests/test_mcp_apps_context.py`: rewrote two tests that had encoded the old host-semantic aliases. Full suite: 2666 passed; mypy, ruff, structural diff (100%), and build/vendor reproducibility all green.

## Note

Includes an incidental `uv.lock` sync (`3.1.0rc1 → rc2`) that `uv` regenerates on any invocation — `main`'s lock was stale relative to the `3.1.0-rc.2` version bump.

## Verification

Token wiring is confirmed against the ext-apps host-styling contract, so no live host probe is needed to validate correctness. What remains is an on-device visual pass in Claude Desktop (light + dark) to confirm the exact host hues balance well; fallbacks render as Paper in a plain browser.

### DIFF
--- a/src/markdown_vault_mcp/static/app.html
+++ b/src/markdown_vault_mcp/static/app.html
@@ -128,33 +128,27 @@ if(__exports != exports)module.exports = exports;return module.exports}));
 </script>
 <style>
   :root {
-    /* Host theme vars — the fallback VALUES carry the Paper "light" identity,
-       so a host that pushes its own --color-* overrides Paper; otherwise Paper
-       shows. The Paper semantic tokens below map onto these. */
-    --color-background-primary: #fbf8f1;
-    --color-background-secondary: #f2ecdf;
-    --color-text-primary: #2a2620;
-    --color-text-secondary: #7a7161;
-    --color-text-inverse: #ffffff;
-    --color-text-info: #b25a35;
-    --color-text-success: #22c55e;
-    --color-text-danger: #b4472c;
-    --color-border-primary: #e4dbc9;
-    /* Paper semantic tokens (map onto host vars so the host can override) */
-    --panel: var(--color-background-primary);
-    --panel-2: var(--color-background-secondary);
-    --ink: var(--color-text-primary);
-    --muted: var(--color-text-secondary);
-    --border: var(--color-border-primary);
-    --accent: var(--color-text-info);
-    --accent-ink: var(--color-text-inverse);
-    /* Paper-only tokens (no host equivalent; flip under [data-theme="dark"]) */
-    --page: #eceae4;
-    --ink-2: #5c5446;
-    --border-2: #efe8d9;
-    --edge: #d3c7ad;
+    /* Neutrals: inherit the host's chrome, fall back to Paper "light". A host
+       that pushes its own --color-* wins; otherwise the Paper fallback shows. */
+    --panel:     var(--color-background-primary,   #fbf8f1);
+    --panel-2:   var(--color-background-secondary, #f2ecdf);
+    --ink:       var(--color-text-primary,         #2a2620);
+    --ink-2:     var(--color-text-secondary,       #5c5446);
+    --muted:     var(--color-text-tertiary,        #928975);
+    --border:    var(--color-border-primary,       #e4dbc9);
+    --border-2:  var(--color-border-secondary,     #efe8d9);
+    --edge:      var(--color-border-primary,       #d3c7ad);   /* graph link color */
+    /* Accent: ALWAYS Paper terracotta — never aliased to a host semantic slot,
+       or the host's blue "info" color leaks into links/tabs/buttons. By SDK
+       convention the host supplies the neutral/semantic ramp; the accent is the
+       app's own token (ext-apps ref theme: basic-server-*/global.css). */
+    --accent:      #b25a35;
+    --accent-ink:  #ffffff;
     --accent-soft: #f3e4d8;
-    --warn: #b4472c;
+    /* Semantic state: host danger with a Paper fallback (the "missing" badge). */
+    --warn: var(--color-text-danger, #b4472c);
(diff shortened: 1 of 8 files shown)


pvliesdonk:
Thanks — good catch to raise, but the `--edge → --color-border-primary` mapping is intentional, not an oversight.

The host theme contract exposes exactly two border slots (`--color-border-primary`, `--color-border-secondary`). `--border-2` already takes `--color-border-secondary`, so an *inherited* `--edge` must share a host slot with either `--border` (primary) or `--border-2` (secondary) — there's no third slot that would keep all three distinct under a host override. The design intent for the graph link is to read as the host's **primary** border tone (the more prominent one), so primary is the deliberate choice; routing `--edge` to `--color-border-secondary` would collapse it onto the *subtler* `--border-2` and make graph links **less** visible under host themes, not more.

The Paper-only distinctness (`--edge` slightly darker than `--border`) is a nicety the "inherit host neutrals" decision (see #861) consciously trades for a native blend — the graph's identity is carried by the terracotta focus node + semantic edges, which stay Paper literals. Under a plain browser (no host tokens) the Paper fallbacks preserve the original edge/border distinction.

Leaving as-is and pinned by the test; the edge/border contrast under a real host is on the on-device visual checklist. If it reads poorly on a specific client we can revisit (e.g. a small graph-only literal override in `getColors()`).

PR #884, opened 2026-07-10
fix: recover keyword/hybrid search for hyphenated terms (#866)

### LINKED ISSUE #866: Hyphenated terms in keyword/hybrid search silently return zero results (raw FTS5 MATCH, no token quoting)
## Summary

A keyword/hybrid search query containing a hyphenated term (e.g. `File-back Ingest Lint`) silently returns zero results, even though the indexed documents contain all the terms. The same query with the hyphen replaced by a space (`file back Ingest Lint`) returns the expected document as top-1 with a high BM25 score.

Hyphenated terms are extremely common in markdown vaults (file slugs like `vault-mcp`, terms like `File-back`, `e5-large`), so this silently degrades a large class of real queries. In `hybrid` mode the failure is masked: the keyword leg contributes nothing and the query falls back to semantic-only ranking without any signal to the caller.

## Repro

Environment: `markdown-vault-mcp` runtime v3.0.3, `fastmcp-pvl-core` 4.0.1, vault of ~1000 md docs, FastEmbed `intfloat/multilingual-e5-large`.

1. Index a document containing the words `Ingest`, `File-back`, `Lint`.
2. `search(query="Ingest", mode="keyword")` -> document found (sanity check, index is fine).
3. `search(query="file back Ingest Lint", mode="keyword")` -> document is top-1, score ~13.
4. `search(query="File-back Ingest Lint", mode="keyword")` -> `{"result": []}` - empty, no error surfaced.

## Suspected cause

`fts_index.py::search()` passes the user string directly as an FTS5 query ("query: FTS5 query string"). In FTS5 query syntax an unquoted `-` between barewords is not a valid token sequence, so the MATCH either raises (and the error appears to be swallowed, returning an empty set) or parses to something that matches nothing. There is `_escape_like()` for folder filters, but no sanitization/quoting of the search terms themselves.

## Suggested fix (either would work)

1. Auto-quote tokens containing FTS5-special characters (`-`, `.`, `:` etc.) as string literals (`"file-back"`) before building the MATCH expression - keeps power-user syntax (AND/OR/NEAR) for well-formed queries, protects natural-language ones; or
2. On `sqlite3.OperationalError` from MATCH, retry once with all tokens quoted (graceful degradation instead of a silent empty result), and/or surface a warning field in the response so hybrid callers know the keyword leg failed.

Happy to provide more details from our deployment. Thanks for the very actively maintained project - it is the backbone of our team knowledge base search.

pvliesdonk:
## Summary

Closes #866.

A keyword/hybrid search query containing a hyphenated term or slug (`File-back Ingest Lint`, `vault-mcp`, `e5-large`) silently returned **zero results**. FTS5 parses the `-`/`:` between barewords as an operator/column filter and raises `OperationalError: no such column: ...`, which `FTSIndex.search()` caught and swallowed to `[]`. In `hybrid` mode this was worse: the keyword leg contributed nothing and the query fell back to semantic-only ranking with no signal to the caller.

## Fix

On an FTS5 **query-syntax** error (not a lock or other environmental error — those still propagate on both the first and the retry execute), `search()` retries the MATCH once with every whitespace token quoted as an FTS5 phrase literal (`"file-back" "ingest" "lint"`, implicit AND). Special characters are then matched literally, and a compound like `File-back` becomes an adjacency phrase over its tokens. Well-formed operator queries (`AND`/`OR`/`NEAR`/phrases/prefixes) parse cleanly and never hit the fallback, so the documented FTS5-query contract is preserved. The one fix in `search()` covers keyword and the hybrid keyword leg (both delegate to it).

Two new helpers: `_is_fts5_query_error` (extracted from the prior inline discriminator, preserving the #327/#560 "don't swallow real DB errors" invariant) and `_quote_fts_query`.

## Behaviour

- `search("File-back Ingest Lint")` / `search("vault-mcp")` now find the doc (keyword **and** hybrid).
- `search("Ingest OR zzzmissing")` keeps operator semantics (bypasses the fallback).
- A non-FTS5 error (DB lock, disk I/O) still propagates on the first **and** the retry execute.
- A query rejected even after quoting degrades to `[]` and logs at **WARNING** (per the logging standard), no longer silently.

## Tests

Hyphen recovery (FTSIndex + SearchManager keyword leg), well-formed operator/phrase/prefix bypass, retry-path non-FTS5 propagation, retry-path FTS5-again → `[]` + WARNING, empty-fallback guard, fallback preserving folder filters, and the quote-doubling escape. The two prior tests asserting `malformed → []` are rewritten to assert graceful degradation. Full suite: 2867 passed. Reviewed locally with the preflight-circus (five context lenses + specialist reviewers) before opening.

## Documentation

`docs/tools/index.md` search note; `FTSIndex.search()` docstring.

### DIFF
--- a/docs/tools/index.md
+++ b/docs/tools/index.md
@@ -81,6 +81,8 @@ Find documents matching a query using full-text or semantic search.
     - Use `mode="semantic"` for meaning-based similarity
     - Check `stats` to see if `semantic_search_available` is true
 
+    Keyword and hybrid modes accept FTS5 operators (`AND`, `OR`, `NEAR`, `"exact phrase"`, `prefix*`). A natural-language query whose terms contain characters FTS5 reserves (a hyphenated slug such as `vault-mcp`, or a colon) is matched literally rather than failing, so plain queries do not need escaping.
+
 **Example usage:**
 
 ```json
--- a/src/markdown_vault_mcp/fts_index.py
+++ b/src/markdown_vault_mcp/fts_index.py
@@ -220,6 +220,49 @@ def _escape_like(value: str) -> str:
     return value.replace("\\", "\\\\").replace("%", "\\%").replace("_", "\\_")
 
 
+def _is_fts5_query_error(exc: sqlite3.OperationalError) -> bool:
+    """True when an ``OperationalError`` is an FTS5 *query-syntax* error.
+
+    Distinguishes a malformed MATCH expression (user input problem, safe to
+    recover from) from an environmental error such as a database lock (must
+    propagate). Matches the error strings SQLite emits for FTS5 query
+    problems: ``fts5: syntax error``, ``no such column`` (a bareword parsed as
+    a column filter), and ``unterminated string``.
+    """
+    msg = str(exc).lower()
+    return (
+        "fts5" in msg
+        or "syntax error" in msg
+        or "no such column" in msg
+        or "unterminated" in msg
+    )
+
+
+def _quote_fts_query(query: str) -> str:
+    """Quote each whitespace-separated token as an FTS5 string literal.
+
+    FTS5 query syntax treats characters like ``-`` and ``:`` between barewords
+    as operators/column filters, so a natural-language query such as
+    ``File-back Ingest Lint`` or a slug like ``vault-mcp`` is rejected with an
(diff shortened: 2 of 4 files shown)


pvliesdonk:
@claude review please review again, but your subagents don't inherit permissions

PR #911, opened 2026-07-14
refactor(server): de-fork server.py onto the template skeleton; config-honoring DOMAIN-WIRING (#901)

### LINKED ISSUE #901: de-fork server.py: sentinel fold + instructions module
Part of #898. **Sequence: parallel with config (depends on the gate).**

## Summary

`server.py` (281 lines) carries domain code outside the skeleton's `DOMAIN-UPSTREAM`/`DOMAIN-WIRING` sentinels, including a sentinel the skeleton does not define.

## Observed `[verified]`

- `_build_default_instructions` (module function) — domain logic, no skeleton equivalent.
- A `DOMAIN-FILE-EXCHANGE-START/END` sentinel that exists in **no** template ref (skeleton defines only `DOMAIN-UPSTREAM` + `DOMAIN-WIRING`). copier only preserves template-shipped sentinels, so this block is not preserved on update.
- A tag-visibility `mcp.disable(tags={...})` block after the wiring, outside any sentinel.

Pristine skeleton is 129 lines.

## Target (approved plan)

Move `_build_default_instructions` to a domain module (`_instructions.py`). Fold the transfer-link wiring (currently under the invented `DOMAIN-FILE-EXCHANGE`) and the `mcp.disable(...)` visibility block into the existing `DOMAIN-WIRING` sentinel; delete `DOMAIN-FILE-EXCHANGE`. Result: only `DOMAIN-UPSTREAM` + `DOMAIN-WIRING` remain, matching skeleton.

## Acceptance criteria

- [ ] `server.py` removed from `RATCHET`; conformance gate strict-passes.
- [ ] Only `DOMAIN-UPSTREAM` + `DOMAIN-WIRING` sentinels remain.
- [ ] Transfer-link registration and read-only/tag-visibility behavior unchanged (server + transfer suites green).
- [ ] `pytest -x -q`, ruff, `mypy`, patch coverage ≥ 80%, structural gate green.

## Scope boundary

`server.py` only.

pvliesdonk:
## What

De-forks the template-owned `server.py` (#901, epic #898). `make_server()`'s body is now **byte-identical to the pristine copier-template skeleton** outside the `DOMAIN-UPSTREAM` / `DOMAIN-WIRING` sentinels; all MVM customization moves into `DOMAIN-WIRING` or dedicated domain modules. `server.py` is removed from the conformance `RATCHET` and now strict-passes — the removal is the proof.

**Domain customization relocated:**

- **`_build_default_instructions` → new `_instructions.py`** (`build_default_instructions`), applied post-construction via the `mcp.instructions` setter in `DOMAIN-WIRING`.
- **Server icon** attached post-construction (`mcp._mcp_server.icons` — FastMCP's `icons` property is read-only and the skeleton ctor carries no icons).
- **Invented `DOMAIN-FILE-EXCHANGE` sentinel folded into `DOMAIN-WIRING`** (transfer wiring); the tag-visibility disable block folded in too.
- **`register_prompts(mcp)`** matches the template's no-arg shape. Prompt registration is **split** by config-dependency: `register_prompts(mcp)` registers the config-independent built-ins at the skeleton call site; a new `register_domain_prompts(mcp, templates_folder, prompts_folder)` — called from `DOMAIN-WIRING` with `make_server`'s already-resolved `config.content` — registers `create_from_template` + user prompts. A user prompt shadowing a built-in prunes it first (`mcp.local_provider.remove_prompt`, guarded against `KeyError`) so the override replaces cleanly with no "component already exists" warning.

## Config is honored, not re-read from env (#609)

The skeleton body sources `SERVER_NAME` / `INSTRUCTIONS` from the environment. `DOMAIN-WIRING` re-applies them from the **passed** config so `make_server(config=X)` honors `X` even when the env var is unset (the `from_env` path is unchanged, since `config.*` mirrors the same env var):

- `mcp.instructions = config.instructions or build_default_instructions(...)` (unconditional).
- server name: when `config.server_name` differs from the env-derived value, `mcp._mcp_server.name` is overridden. `get_server_info` is intentionally **not** re-registered — it stays on its single skeleton registration (keeping the `DOMAIN-UPSTREAM` block the sole source of truth for upstream wiring) and reports the `SERVER_NAME` env identity.
- prompt folders come from `config.content` (via `register_domain_prompts`), never a second env read.

Regression tests pin all three (`test_instructions_resolved_from_provided_config_not_env`, `test_server_name_resolved_from_provided_config_not_env`, `test_user_prompt_resolved_from_provided_config_not_env`).

## Review

Local preflight-circus ran to convergence — **5 full rounds** (5 core lenses + code-reviewer + test/type analysts each round). It caught, and this PR fixes, several real defects the first pass introduced:

1. `register_prompts` re-read env for prompt folders → `make_server(config=X)` registered prompts against the wrong root. Fixed by the config-dependency split.
2. Same class for **instructions** and **server_name** (baseline read env; override gated on config) → operator's programmatic config silently dropped. Fixed to honor config.
3. The built-in override logged a spurious `Component already exists` WARNING on every startup (`add_prompt` is not silent). Fixed by pruning the shadowed built-in first — and the guard test was found **vacuous** (the `fastmcp` logger sets `propagate=False` and `configure_logging_from_env` resets it, so `caplog` saw nothing) and rewritten to capture the logger directly.
4. The prune could crash `make_server` (`remove_prompt` raises `KeyError`) if a built-in failed to register and a user prompt shadowed it. Guarded.
5. The server-name fix initially re-registered `get_server_info`, creating a latent lockstep hazard with the `DOMAIN-UPSTREAM` block. Simplified away (Option above).

**Known follow-up (not fixable in this PR):** the skeleton `make_server` docstring calls the `transport` param *"logging only,"* but `transport` gates the webhook/transfer wiring in `DOMAIN-WIRING`. That docstring is template-owned and conformance-frozen; correcting it requires an upstream fix, filed as **pvliesdonk/fastmcp-server-template#246**. The accurate behavior is documented at the wiring sites.

## Gates

Full suite **2909 passed, 18 skipped**; `mypy src/ tests/` clean; ruff clean; structural diff-gate **0 violations**; conformance gate green (`server.py` now strict). Patch coverage on changed modules ≥ 94%.

Part of #898. Closes #901.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -46,7 +46,8 @@ src/markdown_vault_mcp/
   write_callback.py    -- WriteCallbackDispatcher: deferred git-commit callback worker (#599)
   config.py            -- configuration loading
   domain.py            -- Service: owns the Vault lifecycle (build, boot index/reindex/embeddings jobs, file watcher); get_vault/get_config DI + vault singleton (#902)
-  server.py            -- generic FastMCP server factory (make_server) with tool annotations
+  _instructions.py     -- build_default_instructions: domain server-instructions prose, applied in server.py's DOMAIN-WIRING (#901)
+  server.py            -- generic FastMCP server factory (make_server); template-owned, domain customization confined to DOMAIN-UPSTREAM/DOMAIN-WIRING (#901)
   cli.py               -- CLI entry point
 ```
 <!-- DOMAIN-END -->
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -2121,23 +2121,34 @@ queries the existing ``sections`` table (no file I/O).
 
 The library layer returns typed structs (``TocEntry``, ``SubtreeNote``, ``SubtreeToc``, defined in ``types.py``); the MCP boundary serializes them to the JSON shapes above via ``utils/serialization.toc_payload``, keeping the wire format unchanged.
 
-**Prompts**: 6 built-in prompt templates, plus optional user-defined prompts:
+**Prompts**: 7 built-in prompt templates, plus optional user-defined prompts:
 
 | Prompt | Parameters | Tags | Description |
 |-|-|-|-|
 | ``summarize`` | ``path`` | | Summarize a document |
 | ``research`` | ``topic`` | ``write`` | Search and consolidate as new note |
 | ``discuss`` | ``path`` | ``write`` | Analyze and suggest improvements |
-| ``create_from_template`` | ``template_name`` (optional) | ``write`` | Discover/read/fill/write from a template |
 | ``related`` | ``path`` | | Find related notes, suggest cross-references |
 | ``compare`` | ``path1, path2`` | | Compare two documents |
+| ``propose-links`` | ``scope`` (optional), ``per_note_limit`` (optional) | ``write`` | Propose new links between semantically-close notes |
+| ``create_from_template`` | ``template_name`` (optional) | ``write`` | Discover/read/fill/write from a template |
 
 Write-tagged prompts are hidden in read-only mode by the same
 ``mcp.disable(tags={"write"})`` call that hides write tools.
 
+Prompt registration is split by config-dependency (#901). The template-owned
+``make_server`` body calls ``register_prompts(mcp)``, which registers the six
+config-independent built-ins above (every one except ``create_from_template``).
+The config-dependent prompts — ``create_from_template`` (needs the templates
+folder) and user-defined prompts (need the prompts folder) — are registered by
+``register_domain_prompts(mcp, templates_folder, prompts_folder)``, called from
+``make_server``'s
+``DOMAIN-WIRING`` block with the already-resolved ``config.content`` folders (no
+second environment read; a caller-supplied config is honored — #609).
+
 **User-defined prompts**: when ``MARKDOWN_VAULT_MCP_PROMPTS_FOLDER`` is set,
-``register_prompts()`` scans the directory for ``.md`` files at startup and
-registers each as an MCP prompt. The file stem becomes the prompt name.
+``register_domain_prompts()`` scans the directory for ``.md`` files at startup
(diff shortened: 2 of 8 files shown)


pvliesdonk:
Thanks — addressed in `c8600016`.

**1. Documentation gap (fixed).** `docs/design/design.md`'s Prompts section now names both functions correctly: `register_prompts(mcp)` registers the six config-independent built-ins; the `PROMPTS_FOLDER` scan, `create_from_template`, the shadow-prune, and the per-prompt user backstop all live in `register_domain_prompts(mcp, templates_folder, prompts_folder)` (called from `DOMAIN-WIRING`). The override prose now describes the register-then-prune mechanism (remove the shadowed built-in, tolerate `KeyError`, no "component already exists" warning). While fixing it I also corrected a **pre-existing** gap in the same section's prompt table — it omitted `propose-links` and mislabeled the count as "6" — now 7 rows. Swept the rest of `docs/` + `README.md`: the user-facing prompt tables/prose were already accurate (`propose-links` present, override described as "replaces"), so this was the only stale spot.

**2. `get_server_info` server-name divergence.** Confirmed intended, as you noted — the live `mcp.name` honors a programmatic `config.server_name` while `get_server_info` stays on its single skeleton registration (keeping `DOMAIN-UPSTREAM` the sole source of truth). It matches on the `from_env` path and is pinned by `test_server_name_resolved_from_provided_config_not_env`. No change.

**3. Dependency Audit / pillow CVE.** Agreed it's pre-existing lockfile drift (transitive via `fastembed`), unrelated to any file this PR touches — leaving it out of scope here so the de-fork stays clean; worth a separate `uv lock --upgrade-package pillow` PR.

Local review note: the stale design-doc prose was a blind spot in my pre-push review — the circus lenses scope to the diff and the changed source files, so a doc that *describes* changed behavior but isn't itself in the diff slipped through. Good catch.

PR #950, opened 2026-08-04
feat: add first-class Voyage embedding provider

### LINKED ISSUE #949: First-class Voyage AI embedding provider
## What works today, and why it is not enough

Voyage's `/v1/embeddings` is OpenAI-shaped, and the `openai` provider's `OPENAI_BASE_URL` already reaches it:

```bash
MARKDOWN_VAULT_MCP_EMBEDDING_PROVIDER=openai
OPENAI_API_KEY=pa-...          # actually a Voyage key
OPENAI_BASE_URL=https://api.voyageai.com/v1
OPENAI_EMBEDDING_MODEL=voyage-4
```

This works — I verified it end to end, including the sharp edges: Voyage rejects `dimensions` and `user` with HTTP 400 and accepts only base64 for `encoding_format`, which happens to be exactly what the `openai` SDK sends by default. But nothing in the docs, the config wizard, or `server.json` tells anyone this is possible; you have to know the base URL, know a valid model name, and put a Voyage key in a variable named for OpenAI. And the vector sidecar records `provider: openai`, so the index's true identity is lost.

## Ask

A first-class `voyage` provider: a thin preset over the existing OpenAI-compatible transport with the base URL pinned, its own `VOYAGE_API_KEY` / model variables, a sensible default model, wizard + `server.json` wiring, and docs alongside the other three providers. No new HTTP client and no ABC change needed.

I have a PR ready that does exactly this and will link it here.

Possible follow-ups, deliberately not part of that PR: per-call `input_type` (Voyage's query/document asymmetry needs an `EmbeddingProvider.embed` signature change), and a larger embedding batch size for remote providers (`_EMBEDDING_BATCH_SIZE = 4` is sized for local ONNX memory bounds per #159 and is small for a network API).

salmonumbrella:
## What changed

- add `voyage` as a fourth value for `MARKDOWN_VAULT_MCP_EMBEDDING_PROVIDER`, alongside `openai`, `ollama`, and `fastembed`
- `VoyageProvider` is a preset over the existing `_OpenAICompatEmbeddings` transport — the same relationship `OllamaProvider` already has to it (#916): base URL pinned to `https://api.voyageai.com/v1`, reading `VOYAGE_API_KEY` and `MARKDOWN_VAULT_MCP_VOYAGE_MODEL` (default `voyage-4`). No new HTTP client, no change to the `EmbeddingProvider` ABC
- wire the new provider and both env vars into the config wizard (`VOYAGE_API_KEY` in `secretKeys`) and both `server.json` package blocks, keeping the drift guards green
- document the provider everywhere the other three are enumerated (design spec, embeddings guide, configuration reference, API docs, README, install/client guides, `packaging/env.example`)
- add a `context_length` table for the current Voyage families (32k for voyage-4 / voyage-3.5 / voyage-code-3 / voyage-finance-2, 16k for voyage-law-2, per docs.voyageai.com), `None` for anything unknown

## Why

Voyage already works today — you point the OpenAI provider at it:

```bash
MARKDOWN_VAULT_MCP_EMBEDDING_PROVIDER=openai
OPENAI_API_KEY=pa-...          # actually a Voyage key
OPENAI_BASE_URL=https://api.voyageai.com/v1
OPENAI_EMBEDDING_MODEL=voyage-4
```

That is wire-identical to what this PR does, and it keeps working. The problem is that it is folklore: nothing in the docs, the config wizard, `server.json`, or the provider list says it is possible, you have to know Voyage's base URL and a valid model name, and the vector sidecar records the provider as `openai`, so two vaults embedded with different backends look interchangeable when they are not. This PR turns that configuration into a named provider so the endpoint, the key variable, the model default, and the persisted provider identity are all discoverable.

## Usage

```bash
MARKDOWN_VAULT_MCP_EMBEDDING_PROVIDER=voyage
VOYAGE_API_KEY=pa-...
# optional, defaults to voyage-4:
MARKDOWN_VAULT_MCP_VOYAGE_MODEL=voyage-4-large
```

| Variable | Default | Notes |
|---|---|---|
| `MARKDOWN_VAULT_MCP_EMBEDDING_PROVIDER=voyage` | — | opt-in; never auto-detected |
| `VOYAGE_API_KEY` | (none) | bare name, matching the `OPENAI_API_KEY` / `OLLAMA_HOST` convention |
| `MARKDOWN_VAULT_MCP_VOYAGE_MODEL` | `voyage-4` | prefixed, matching `MARKDOWN_VAULT_MCP_OLLAMA_MODEL` |

## Design notes worth reviewing

- **Guarded params.** Voyage is strict about OpenAI fields it does not implement: `dimensions` → HTTP 400, `user` → HTTP 400, and `encoding_format: "float"` → 400 with *"accepted values are 'base64'"* (verified against the live API). The shared transport already sends only `model` and `input` and lets the `openai` SDK apply its base64 default, which Voyage accepts — so this needed no request shaping, but it is now an explicit constraint in the docstring and pinned by a test asserting the request carries exactly `{model, input}`.
- **Default model `voyage-4`.** Voyage's own guidance is *"voyage-4-large for the best quality, voyage-4-lite for the lowest latency and cost, and voyage-4 for a balance between the two"* — the balanced member is the default, the same posture as `text-embedding-3-small` for the OpenAI provider.
- **Not auto-detected.** `voyage` must be selected explicitly. Adding it to the auto-detect chain would mean a `VOYAGE_API_KEY` exported for an unrelated tool could silently take over an existing index and force a full re-embed. The existing chain (`openai` → `ollama` → `fastembed`) is unchanged.
- **`_explicit_provider()` extraction.** Not cosmetic: a seventh `return` in `get_embedding_provider` trips `PLR0911` under the structural diff gate, and splitting explicit dispatch from the auto-detect probe was the honest fix rather than a `# noqa`.

## Deliberately out of scope

- **`input_type`.** Voyage's asymmetric `query` / `document` prefixes measurably help retrieval, but they need a per-call parameter on `EmbeddingProvider.embed` and a way for the search path to say which side it is on. That is an ABC change affecting all four providers and belongs in its own PR.
- **Batch-size tuning.** `_EMBEDDING_BATCH_SIZE` is 4, sized for local ONNX memory bounds (#159); that is small for a remote API, but raising it per-provider is its own change.
- **`VOYAGE_BASE_URL`.** No override is offered. Anyone needing a proxy or gateway in front of Voyage still has the `openai` provider with `OPENAI_BASE_URL`, which is exactly the escape hatch this PR is formalising.

## Verification

- `uv run pytest -q` — 2983 passed, 25 skipped; no test touches the network, the `openai` SDK is faked in-process throughout
- `uv run ruff check .` / `ruff format --check .` — clean
- `diff-cover --compare-branch=origin/main` — 100% patch coverage (46 changed lines, 0 missing)
- `diff-quality --violations=ruff.check --options="--extend-select=C901,PLR0911,PLR0912,PLR0913,PLR0915,S" --fail-under=100` — 100%, 0 violations
- `uv run mkdocs build --strict` and `vale docs/` — clean

Closes #949.

### DIFF
--- a/.claude-plugin/plugin/README.md
+++ b/.claude-plugin/plugin/README.md
@@ -39,7 +39,7 @@ Optional environment variables:
 | Variable | Default | Purpose |
 |---|---|---|
 | `MARKDOWN_VAULT_MCP_READ_ONLY` | `true` | Disable write tools. |
-| `MARKDOWN_VAULT_MCP_EMBEDDING_PROVIDER` | *(empty)* | `fastembed` / `ollama` / `openai`. Leave blank for keyword-only search. |
+| `MARKDOWN_VAULT_MCP_EMBEDDING_PROVIDER` | *(empty)* | `fastembed` / `ollama` / `openai` / `voyage`. Leave blank for keyword-only search. |
 | `MARKDOWN_VAULT_MCP_EXCLUDE` | `.obsidian/**,.trash/**,.git/**` | Glob patterns to skip. |
 
 For the full list of env vars, see the
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Point it at a directory of Markdown files (an Obsidian vault, a docs folder, a Z
 
 <!-- DOMAIN-START -->
 - **Full-text search** — SQLite FTS5 with BM25 scoring, porter stemming
-- **Semantic search** — cosine similarity over embedding vectors (FastEmbed, Ollama, or OpenAI)
+- **Semantic search** — cosine similarity over embedding vectors (FastEmbed, Ollama, OpenAI, or Voyage AI)
 - **Hybrid search** — Reciprocal Rank Fusion combining FTS5 and vector results
 - **Diversity-aware ranking** — each search result list caps a single document at 2 chunks (configurable), downweights chunks of long documents, and returns sentence-scale snippets — bounded LLM context cost per query, with full-section recovery via `read(path, section=heading)`
 - **Adaptive heading-level chunking** — long sections are recursively re-split at deeper heading levels (H1 → H6) until each chunk fits a configurable word budget, improving retrieval precision on synthesising essays without manual restructuring
@@ -62,7 +62,7 @@ With optional dependencies:
 
 ```bash
 pip install markdown-vault-mcp[mcp]            # FastMCP server
-pip install markdown-vault-mcp[embeddings-api]  # Ollama/OpenAI embeddings via API
+pip install markdown-vault-mcp[embeddings-api]  # Ollama/OpenAI/Voyage embeddings via API
 pip install markdown-vault-mcp[embeddings]      # FastEmbed local embeddings
 pip install markdown-vault-mcp[all]             # MCP + FastEmbed + API embeddings
 ```
@@ -83,7 +83,7 @@ docker pull ghcr.io/pvliesdonk/markdown-vault-mcp:latest
 ```
 
 <!-- DOMAIN-START -->
-The Docker image uses `[all]` (MCP + FastEmbed + API embeddings). By default, semantic search works locally with FastEmbed and can switch to Ollama/OpenAI when configured. A `compose.yml` ships at the repo root as a starting point — copy `.env.example` to `.env`, edit, and `docker compose up -d`.
+The Docker image uses `[all]` (MCP + FastEmbed + API embeddings). By default, semantic search works locally with FastEmbed and can switch to Ollama/OpenAI/Voyage when configured. A `compose.yml` ships at the repo root as a starting point — copy `.env.example` to `.env`, edit, and `docker compose up -d`.
 
 To attach a remote Python debugger (development only; the protocol is unauthenticated), see [Remote debugging](docs/deployment/docker.md#remote-debugging).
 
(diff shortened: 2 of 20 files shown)


pvliesdonk:
@claude review

salmonumbrella:
@pvliesdonk can you please add voyage into release 4.0?

pvliesdonk:
Superseded by #1169, now merged — closing this as shipped rather than rejected.

Thank you, @salmonumbrella. The work here is what landed: your commit was cherry-picked into #1169 with **authorship preserved**, so `git log` credits you for the Voyage provider. This branch had gone stale against `main` and lives on a fork, so rebasing it in place was not something a maintainer-side agent could push to; carrying the commit forward was the way to ship it without losing your name on it.

Two adjustments were needed for the four months of drift, both worth knowing:

- The branch predated the `embed_timeout_s` plumbing (#954), so the extracted `_explicit_provider` built providers without a timeout. That is threaded through now, and `VoyageProvider` gained the same `timeout` knob as its sibling presets, with a test.
- `EmbeddingsConfig.from_env` no longer exists — `ProjectConfig` owns `from_env` now — so the Voyage vars became flat fields there, with `VOYAGE_API_KEY` declared in `config-presentation.domain.yml` as a bare-env read the AST scan cannot see.

Your design decisions all survived review untouched: the pinned base URL, `voyage-4` as the balanced default, and above all keeping `voyage` out of the auto-detect chain so a stray `VOYAGE_API_KEY` cannot silently take over an existing index.

The follow-ups you flagged are tracked: per-call `input_type` at #1135, and the documented base-URL recipe plus the written acceptance bar for named provider presets at #1134.

To answer your earlier question about 4.0 — it shipped after that release, so it rides the v4.1 milestone instead.

---

PR #981, opened 2026-08-09
feat(transfer): adopt fastmcp-pvl-core transfer subsystem; retire local store

### LINKED ISSUE #979: Transition the transfer tools to fastmcp-pvl-core's transfer subsystem
## Summary

markdown-vault-mcp ships a hand-rolled, in-memory HTTP transfer subsystem; `fastmcp-pvl-core` (already pinned at 4.6.1) now owns an equivalent subsystem. Adopt the upstream subsystem and retire the local copy.

## Observed

- Local subsystem in this repo: `src/markdown_vault_mcp/transfer/store.py`, `transfer/routes.py`, `_server_transfer.py`, `config_sections/transfer.py`, wired in `server.py` via `register_transfer(mcp, config)`. The store is an in-memory dict guarded by a lock (lost on restart) and hard-burns a link on success.
- `fastmcp-pvl-core` 4.6.1 provides `register_transfer_routes(mcp, config, transfer_config, *, sink, validate)` (ADR 0001), which owns the `/transfer/{token}` route, the two link tools (`create_download_link` / `create_upload_link`), the KV-backed token store, and the TTL / size-cap / lease / grace mechanics. The downstream supplies only two domain hooks: a `TransferSink` (where bytes come from / go to) and a `TransferValidator` (which refs are acceptable). It also exposes the primitives core uses behind those tools (`TransferSink`, `TransferConfig`, `TransferReadResult`, plus `fetch_url` / `decode_base64_capped`).
- The dependency was bumped to 4.6.1 without adopting the subsystem; the local wiring was kept.

## Desired outcome

markdown-vault-mcp uses pvl-core's transfer subsystem for note/attachment download and upload: a domain `TransferSink` + validator over the vault, wired via `register_transfer_routes`, with the local subsystem removed. `TransferConfig` comes from pvl-core (composed config field) rather than the local dataclass.

## Constraints (direction agreed with the maintainer)

- The upstream link tools are owned by pvl-core and are meant to be **identical on every server**. Use them as-is — do not mutate the registered tools' metadata (title/annotations/icons) downstream.
- Do not rebuild the store or route locally. Two legitimate paths exist and both are pvl-core's: use the default tools as-is, or build a domain-specific tool on the same primitives core uses. A generic file download/upload is the former.
- Retiring the local subsystem must preserve today's behaviour for existing note/attachment links (path validation, attachment-extension allowlist, size cap, one-time semantics).

## Scope boundary

- **No OKF bundle export here.** Bundle delivery is a separate, OKF-specific concern handled by overloading the download ref inside the domain hook (tracked on the OKF export issue #963).
- Whether pvl-core's default transfer tools should carry generic titles/annotations is an **upstream** question for `fastmcp-pvl-core`, not something this repo should force or work around locally.

## Open questions

- `[unverified]` How `create_upload_link` is gated in read-only mode under core's registration — whether the uniform tool carries a `write` tag that mvm's `mcp.disable(tags={"write"})` pass already hides, or whether another mechanism applies. Verify against pvl-core rather than assuming.
- `[unverified]` How the tool-metadata enforcement test (`tests/test_server.py::test_every_registered_tool_has_title`) should treat pvl-core-owned tools — e.g. the same exemption already applied to `get_server_info` — rather than any downstream mutation.
- `[unverified]` Whether the KV backend default (`config.server.kv_store_url`, defaulting to `file:///data/state`) is adequate for existing deployments, or whether operators need guidance.
- `[unverified]` Behavioural deltas from the switch (e.g. hard-burn → grace-settle on success; a mid-transfer sink error surfacing as 500 rather than the local subsystem's 503/415) and whether any are user-visible enough to document.

## Documentation

`docs/configuration.md` (transfer env vars, including the two new ones pvl-core's `TransferConfig` adds), `examples/*.env`, and the config-wizard surface (`scripts/gen_config_surface.py`).

Enables the OKF bundle export (#963). Refs #959.

pvliesdonk:
Closes #979.

## What & why

markdown-vault-mcp shipped a hand-rolled, in-memory HTTP transfer subsystem (`transfer/`, `_server_transfer.py`, `config_sections/transfer.py`, from #622). `fastmcp-pvl-core` 4.6.1+ folded that pattern into a shared, KV-backed subsystem (ADR 0001), and **4.8.0** made both adoption paths first-class:

- `register_transfer_routes` registers the two generic `create_download_link` / `create_upload_link` tools **with full metadata** — titles, hints, icons, and the `write` tag on upload — so downstream neither mutates them nor rebuilds the machinery.
- `build_transfer_links` exposes the link minter (`TransferLinks`) for domain-authored tools (path 2), for later use by OKF bundle export (#963).

The dependency had been bumped to 4.6.1 without adopting the subsystem; this adopts it and retires the local copy.

## Changes

- **Bump** `fastmcp-pvl-core` to `>=4.8.0,<5`.
- **`VaultTransferSink`** (`_transfer_sink.py`) — the one domain hook pvl-core consumes: `validate` / `read` / `write` over vault notes and attachments, lifting the path validation from the old `_server_transfer.py`. The sink handle is just the vault-relative path; note-vs-attachment is re-derived from the `.md` suffix.
- **Wiring** — `server.py`'s `DOMAIN-WIRING` now calls `register_transfer_routes(mcp, config.server, config.transfer, sink=…, validate=…)` on HTTP/SSE transport with `base_url` set, passing `download_note` / `upload_note` for vault-specific context. **No `_finalize_transfer_tool_metadata` / no downstream mutation** — the tool titles/hints/icons/`write` tag come from core.
- **Config** — compose pvl-core's `TransferConfig` as a field (replacing the local dataclass). Shared env var names preserved (`TRANSFER_TTL_DEFAULT_S`, `TRANSFER_TTL_MAX_S`, `TRANSFER_MAX_UPLOAD_BYTES`); two new optional ones appear (`TRANSFER_GRACE_TTL_S`, `TRANSFER_LEASE_S`). The config-surface generator discovers them from core's own field metadata (a new `transfer_inventory` tier in the wizard-spec drift test mirrors this).
- **Delete** `transfer/`, `_server_transfer.py`, `config_sections/transfer.py`, and the three old transfer test files.
- **Tests** — add `test_transfer_sink.py` (validate/read/write, note+attachment, traversal / bad-extension / missing-file / bad-UTF-8 paths); rewrite `test_transfer_routes.py` against `make_server` (present on HTTP+base_url, absent on stdio / without base_url, titled/hinted/iconed, upload hidden read-only); repoint `test_config.py` and `test_server.py`.
- **Docs** — regenerate the config surface (README, wizard-spec, env examples, `server.json`); update `docs/configuration.md` and the internal `docs/design/design.md` transfer section.

## Behaviour deltas (from the switch)

- Token store is now KV-backed (`kv_store_url`, on-disk by default) instead of in-memory, so live links survive a restart.
- On success a link is **grace-settled** (TTL shrinks to `TRANSFER_GRACE_TTL_S`) rather than hard-burned, so a served-but-stalled transfer can retry.
- Without `base_url` set on an HTTP transport, the transfer tools/route are now simply **not registered** (previously they registered and raised at call time).

## Gates

- `uv run pytest -x -q`: **3084 passed**, 27 skipped
- ruff check / format --check: clean · mypy `src/ tests/`: clean
- structural diff-quality: **100%** · `_transfer_sink.py` coverage: **100%**
- `scripts/gen_config_surface.py --check`: clean · manifest versions untouched

## Out of scope

OKF bundle export (#963) — it builds on this transition's sink and the `TransferLinks` minter in a follow-up.

https://claude.ai/code/session_01268CY9P38eZx1JvB9a12Mr

---
_Generated by [Claude Code](https://claude.ai/code/session_01268CY9P38eZx1JvB9a12Mr)_


pvliesdonk:
Thanks — both findings verified as real, and both are **upstream pvl-core gaps** that this adoption surfaces rather than mvm-side bugs, so they're being fixed there rather than worked around downstream (which would re-diverge from the subsystem this PR just adopted).

**1. Vault-unavailable → 500 instead of 503 (`_transfer_sink.py`).** Confirmed: `Service.stop()` clears the vault singleton and the lifespan is ref-counted, so the vault is torn down when all sessions close while the `/transfer` route stays mounted — a link followed in that window raises `RuntimeError` → generic 500. The adopted pvl-core route maps *every* sink exception to 500, so a sink cannot signal 503 through it. This is exactly [fastmcp-pvl-core#233](https://github.com/pvliesdonk/fastmcp-pvl-core/issues/233) (sink→status mapping); I've added the transient-unavailable/503 scenario there. Restoring 503 downstream isn't possible without reaching around the adopted route.

**2. Wizard deployment gating dropped (`config.py` / wizard-spec).** Confirmed: pvl-core's `TransferConfig` field metadata has no `when: server`, so the five transfer settings now render for stdio deployments too. Since transfer is HTTP-only for *every* adopter, the gating belongs on core's field metadata; filed as [fastmcp-pvl-core#255](https://github.com/pvliesdonk/fastmcp-pvl-core/issues/255). The domain-presentation override is documented as for vars the scan can't see, not for re-annotating a scanned core var, so it's not the right lever here.

Both are minor edge deltas (a 500-vs-503 status during a server-idle window with the link still retryable; a few wizard fields shown where they're inert). Once #233 and #255 land upstream and the pin bumps, both resolve with no mvm-side workaround.

---

pvliesdonk:
Both review findings are now **resolved upstream and adopted here** (pushed in `d486eca`, bumping `fastmcp-pvl-core` to 4.9.0):

**1. Vault-unavailable → 503 (was 500).** [fastmcp-pvl-core#233](https://github.com/pvliesdonk/fastmcp-pvl-core/issues/233) shipped a `TransferSinkError` family. `VaultTransferSink` now raises `TransferUnavailableError` (retryable **503**) when the vault singleton is torn down mid-request, and `TransferResourceGoneError` (**410**) when a note/attachment validated at mint time has since been removed — upgrading the mint-then-delete race from a generic 500 to a semantic 410. The route maps these; everything else still maps to 500. New tests cover 503 (read + write) and 410 (note + attachment); sink coverage is back to 100%.

**2. Wizard deployment gating.** [fastmcp-pvl-core#255](https://github.com/pvliesdonk/fastmcp-pvl-core/issues/255) added a `when: "server"` wizard hint to `TransferConfig`'s fields, so regenerating the config surface restored `showIf: {deployment: ["server"]}` on all five transfer vars — with no domain-side override, the gating now comes from core's field metadata.

No downstream workarounds; both fixes are the upstream contract. Full suite green (3088 passed), gates clean. This makes the adoption fully clean rather than carrying tracked deltas.

---

pvliesdonk:
@claude review without subagents

PR #984, opened 2026-08-09
feat(okf): enforced write layer — provenance stamping, verification invalidation, okf_verify (#964)

pvliesdonk:
## Summary

Implements the frontmatter-enforcement half of the OKF **enforced write layer** (design §6, issue #964), gated behind the new `MARKDOWN_VAULT_MCP_OKF_WRITE` flag (default `false`). When the flag is off the write path is byte-for-byte untouched.

On an **OKF-active** vault with the flag on, every `write`/`edit`:

- **Stamps provenance** — sets/replaces `generated: {by, at}`. The actor is `human:<subject>` when the caller is authenticated, else a `markdown-vault-mcp/<version>` tool actor. `sources` and all other fields are left alone.
- **Invalidates prior review** — clears any `verified` attestation, because a content change (including a frontmatter-only edit) means an earlier human review no longer describes the note. `rename` bypasses the enricher and preserves both fields.

New **`okf_verify`** tool (tagged `okf-enforce`, hidden unless `OKF_WRITE` is on): appends `{by: human:<subject>, at}` to a note's `verified` list, promoting its trust tier to `human-reviewed`. Verification is attributable, so it raises a `ToolError` when the server runs with no auth. Its own write is suppressed so the enricher does not clear the attestation it just added.

## How the actor reaches the write path

The provenance actor is a FastMCP request-context dependency (`get_subject`), resolvable only inside a tool handler — but the write runs deep under `asyncio.to_thread`. The tool binds an `OkfWriteIntent` on a contextvar around the `to_thread` call; `to_thread` copies the context into its worker, so the enricher (running in that worker) sees it. This mirrors pvl-core's own `_current_auth_mode` contextvar and avoids widening the `DocumentManager.write`/`edit` signatures. The one-shot migration transforms (`okf_convert_links`/`okf_generate_index`/`okf_seed_log`) bind a *suppressing* intent so a mechanical rewrite never re-stamps or discards a human attestation.

## Config

`ContentConfig` gains an `okf_write: bool` field with cross-validation (`okf_write=true` requires `okf_mode != off`, else a `ConfigurationError`) and wizard metadata. The config surface (README, env examples, wizard spec, `server.json`) is regenerated.

## Scope note

Convention maintenance on writes (`log.md` append + affected-folder `index.md` refresh, the second bullet set of design §6) is deliberately **out of scope here** and tracked as the second half of #964 — hence `Refs #964`, not `Closes`.

## Tests

`tests/test_okf_write.py` — the pure transforms, actor resolution, the enrichment builder + suppress contextvar, config cross-validation, the full invalidation matrix (body write / body edit / frontmatter-only edit clear `verified`; rename preserves it) through a writable OKF-active vault, and `okf_verify` end-to-end through the MCP server under both an authed fixture and a no-auth server.

## Documentation

- `docs/configuration.md` — `MARKDOWN_VAULT_MCP_OKF_WRITE` row
- `docs/tools/index.md` — `okf_verify` summary row + detail section
- `docs/guides/okf.md` — new "The enforced write layer" section
- `README.md` + env examples + wizard spec — regenerated via `gen_config_surface.py`
- design §6 (`docs/design/okf.md`) already specifies this phase; the implementation conforms

## Gates

Tests (3144 passed), ruff, mypy, patch coverage (new modules exercised), structural diff-quality (100%), config-surface `--check`, and Vale on the changed docs all pass locally.

Refs #964

---
_Generated by [Claude Code](https://claude.ai/code/session_01268CY9P38eZx1JvB9a12Mr)_


pvliesdonk:
All four review findings addressed in `facaf1a`:

1. **`fetch` provenance (correctness)** — bound `okf_write_intent(OkfWriteIntent(actor=resolve_write_actor()))` around fetch's `.md` write, so an authenticated caller's fetched note is now attributed to `human:<subject>` instead of the tool actor. Attachments go through `write_attachment` (the enricher doesn't touch them), so no intent there. Added a test with a mocked httpx client asserting the human actor.

2. **`okf_verify` optimistic concurrency** — the read-modify-write now passes `if_match=note.etag`; a concurrent write/edit/delete in the window raises `ToolError` ("changed since it was read; retry") rather than clobbering the change or resurrecting a deleted note. Added a test that simulates a racing write between the read and the write. (The pure-migration siblings share the older pattern but are pre-existing and out of scope here.)

3. **Triplicated `"human:"` prefix** — collapsed to the single `okf.py._HUMAN_ACTOR_PREFIX`, now imported by both `_okf_write.py` (removed its `_HUMAN_PREFIX`) and `okf_verify`, so a future convention change can't silently desync `derive_trust_tier`'s matching.

4. **Convention-maintenance spec drift** — the split was intentional (PR-A vs PR-B), but you're right that the spec didn't record it. Rather than pull the deferred work into this PR, I made the deferral explicit: design §9 now splits Phase 5 into **5a** (shipped: stamping, invalidation, `okf_verify`) and **5b** (deferred: `log.md`/`index.md` maintenance), §6 carries a Status note, and the user guide gains a "What the layer does not yet do" note pointing at the migration tools. 5b stays tracked under #964.

All gates green locally (3146 tests, ruff, mypy, structural diff-quality 100%, Vale, patch coverage).

---

pvliesdonk:
@claude review without subagents

PR #1002, opened 2026-08-11
OKF ranking downweight (phase 6) — baseline run

pvliesdonk:
Adds the OKF downweight ranking factor and wires it into the search channels with tests and docs. (Routine PR.)


pvliesdonk:
Test complete — pin-combo full-skill review verified (PR #1003 posted validated inline findings). Closing.

PR #1038, opened 2026-08-14
feat(prompts): ship client-side summarization recipe as summarize-subtree prompt

### LINKED ISSUE #1035: [Feature]: Client-side summarization for agentic MCP clients — a server-shipped prompt that delegates planning and map-reduce to client subagents
*(Using feature-request.yml — the request is an enhancement, not a bug or decay observation.)*

Summarizing more than a couple of notes currently has exactly one route: the server-side `summarize` tool, which requires an operator-provisioned OpenAI-compatible backend. That leaves three groups underserved:

- **Keyless installs have no summarization at all.** Without `OPENAI_API_KEY` or a base URL the tool is hidden (`[verified: has_provider() gate in server.py]`), even when the connected client is a fully capable agent (Claude Code over the HTTP channel) that could do the work with its own model.
- **Privacy-sensitive vaults can't use it.** The tool sends note content to the external provider by design; its own docs warn against summarizing content that must not leave the environment. A client-side route keeps content inside the user's existing model relationship.
- **Agentic clients that improvise do it badly.** Without a recipe from the server, an agent asked to "summarize this folder" tends to read every note into its *primary* context window — the exact context blowout the summarize tool exists to avoid. The server knows how to partition this work (it does so internally, `[verified: SummarizeManager._pack_batches + _SYSTEM_MAP/_SYSTEM_REDUCE prompts]`) but currently shares none of that with clients.

There is also a quality ceiling on the server route: the backend receives a frozen text blob per request, so it cannot follow a link mid-summary or pull in a referenced note — a subagent with vault tools can.

Context: this is the parked deferral from #1034 / #1033. MCP sampling as the mechanism for "use the client's model" was evaluated and rejected — deprecated by the 2026-07-28 spec and never supported by Anthropic clients (design decision log entry 18) — which points at client-side *orchestration* rather than server-initiated inference.

## Desired outcome

An agentic MCP client connected over plain MCP — HTTP hosting, the common deployment channel; no Claude Code plugin required — can produce faithful multi-note/subtree summaries using its own model, such that:

- **The primary context stays small.** Note bodies are confined to subagent contexts; the primary conversation holds only the partition plan (a path list) and the final summary. Per discussion, planning itself is also delegated to a subagent rather than done by the primary or a server tool.
- **The recipe travels with the server**, not per-client folklore: partitioning guidance, the map-phase instructions that preserve per-note path attribution, and the reduce-phase instructions arrive from the vault itself, so every agentic host runs the same tested procedure.
- **The two routes advertise each other.** Where a summarize backend is configured, the client-side route names the tool as the single-call alternative; where none is configured, the situation stops being "no summarization" and becomes "use the client-side route".
- **Nothing changes for non-agentic clients** — the dual-mode `summarize` tool (#1034) remains the baseline.

## Implementation ideas

Non-binding starting points from the discussion that produced this request:

- **Delivery vehicle: an MCP prompt** (e.g. `summarize-subtree`), since prompts are served over HTTP with the server itself and the machinery already exists (`_server_prompts.py` + `static/prompts/`, which today includes a trivial single-note `summarize` prompt) — unlike plugin skills, which only reach the Claude Code plugin channel. A plugin skill could later wrap the same recipe as channel sugar.
- **Recipe shape**: planner subagent (`get_toc` / `list_documents` → batches of paths) → primary agent fans out mapper subagents in parallel (each reads its own batch, returns a partial summary with path attribution) → reduce step (subagent, or inline when partials are small). The plan must return to the primary for the fan-out because subagents cannot spawn subagents in Claude Code `[unverified for other hosts]`; a degenerate sequential single-subagent mode could cover hosts without parallel subagents.
- **Reuse the tuned prose**: the map/reduce system-prompt lessons in `managers/summarize.py` (`_SYSTEM_MAP` / `_SYSTEM_REDUCE`: preserve specifics and path attribution, no meta-commentary) encode exactly what mappers and reducers need.
- `[unverified]`: budget-aware batch packing needs a size signal the toc does not carry — `SubtreeNote` is path/title/headings only (`[verified: types.py]`) — so the planner either batches by note count or a size field gets added to `get_toc`/`list_documents`; the latter may deserve its own issue.
- `[unverified]`: how MCP prompts surface across hosts (Claude Code slash commands vs. claude.ai prompt UI) and whether the prompt text can assume subagent facilities exist, or must detect and degrade.

pvliesdonk:
## Closes / Refs

Closes #1035  (Refs #1036)

## What & why

Adds the `summarize-subtree` MCP prompt: a server-shipped recipe that lets clients produce faithful multi-note/folder summaries with their own model, processing notes in batches so note bodies stay out of the retained conversation context. The mapper/reducer instructions are derived from the tuned `_SYSTEM_MAP` / `_SYSTEM_REDUCE` prose in `managers/summarize.py` (SYNC comments on both sides).

**The prompt adapts to the server's configuration.** It is registered config-dependently via `register_domain_prompts` (the `create_from_template` pattern, honoring user-prompt overrides); a `${route_note}` slot in the static body resolves at registration time from `config.summarize.has_provider()`:

- Backend configured → the prompt opens by preferring the single-call `summarize` tool, keeping the recipe as the fallback when the user wants their own model to do the work.
- No backend → the slot is removed; the prompt carries the recipe alone and never names a tool the server doesn't have. The server instructions point clients at the prompt in this case (keyless installs no longer read as "no summarization").

**The recipe is subagent-optional.** Parallel fan-out, sequential mappers, and inline batch-by-batch processing are all first-class execution modes; the invariant is that only partial summaries survive each batch.

**Route framing stays off model-facing surfaces.** The tool docstring does not mention the prompt; the instructions name it only in the no-backend case; route choice is an operator decision (configure a backend or not). The trade-offs — context overhead (the tool costs the primary model nothing), cost accounting (backend usage is billed outside plan-based client subscriptions; irrelevant for local or pay-per-token backends), data recipients (neither route inherently more private — the backend can be a local model), link-following capability — are operator-facing and documented in `docs/prompts.md` only.

Batching is by note count with heading counts as a rough size proxy, since the toc carries no note sizes (noted as `[unverified]` in the issue; see deferrals). The existing single-note `summarize` prompt is kept (different, trivial use case) and refers folder-scale asks to `summarize-subtree`.

- Plugin-channel delivery (a `vault-summarize` skill + mapper agent definition in the shipped Claude Code plugin, wrapping this same recipe): #1036
- A size/character field on `get_toc` / `list_documents` for budget-aware batch packing (the issue flags this as possibly deserving its own issue; the recipe batches by note count with heading counts as a proxy until such a signal exists): no issue filed yet — can be split out if wanted

## Local review

- [x] Ran a local code-review pass on the cumulative diff before creating the PR.

## Docs impact

- [x] `README.md` — prompt count, prompts table row, summarize tool row cross-ref
- [x] `docs/` site pages — `docs/prompts.md` (new `summarize-subtree` section with the operator-facing route discussion), `docs/tools/index.md` (summarize backend admonition)
- [x] `docs/design/` — prompts table, registration split, client-side route paragraphs (adaptive registration + trade-off-placement rule), decision log entry 20
- [x] Inline docstrings — `register_domain_prompts` / `_register_summarize_subtree`, `build_default_instructions`; `.claude-plugin/plugin/README.md` prompt list; Vale vocab gains `agentic`/`subagent(s)`

https://claude.ai/code/session_011FYTwmTyuzNASbBRSEB6ws


pvliesdonk:
@claude review with a focus on the summarization strategies, and on the various paths an operator/user can follow (availability of external api, availability of subagents)

pvliesdonk:
Fixed in e6458ef — the truncation-recovery finding was real. Step 1.2 now recovers by re-enumerating completely (retry `get_toc` with a higher `max_notes`, or call `list_documents(folder=<entry>)`) and states explicitly that a truncated listing must never be planned from; the fragile subfolder walk is gone and `list_folders` left the recipe's tool lists. A test now pins the recovery wording.

One correction to the review's fix analysis: `list_documents(folder=...)` is recursive (folder **and** sub-folders, `SearchManager.list`) and has no cap, so it single-handedly closes the loose-notes gap — the review's suggestion to use it *alongside* a subfolder walk assumed it was non-recursive. That also resolves the "named but never invoked" observation: it is now the uncapped enumeration fallback in the plan step.

---

PR #1048, opened 2026-08-15
feat(plugin): vault-setup skill and SessionStart doctor hook for in-session bootstrap

### LINKED ISSUE #1042: [Feature]: Plugin-shipped bootstrap — a setup skill and SessionStart doctor hook so a misconfigured install repairs itself in-session
*(Using feature-request.yml — enhancement request. Companion to #1040: that issue covers the `userConfig` configuration screen shown at enable time; this one covers what happens when that screen was skipped, answered wrong, or the config later breaks. Filed separately because the two ship independently and #1040 stays useful without this.)*

When the plugin's required setting is missing or wrong (`MARKDOWN_VAULT_MCP_SOURCE_DIR` unset, or pointing at a moved/renamed directory), the MCP server cannot start and the user gets a dead `Failed to connect` entry in `/mcp` with no in-product path to a fix — the fix lives in a README and a shell profile. The same dead end appears later in the plugin's life, not just at install: vault moved, git token expired, embedding provider misconfigured.

The plugin container itself is the natural place for the repair path, and research against current Claude Code docs shows the pieces exist and compose:

- When a plugin's MCP server fails to start, the plugin's skills, commands, agents, and hooks **all keep working** — "server down, setup skill available" is a supported state `[verified: https://code.claude.com/docs/en/mcp-quickstart.md troubleshooting section]`.
- A settings file `env` block reaches plugin-launched MCP server processes at startup, and takes precedence over the shell environment — so config written to `.claude/settings.json` genuinely satisfies the `${VAR}` expansion in the plugin's `.mcp.json` `[verified: https://code.claude.com/docs/en/env-vars.md and settings.md precedence rules]`.
- SessionStart hooks can run bundled scripts from `${CLAUDE_PLUGIN_ROOT}` and emit `systemMessage` / `additionalContext` that Claude sees `[verified: https://code.claude.com/docs/en/hooks.md]`.
- MCP servers only (re)start at session start; nothing picks up new env mid-session, so any bootstrap flow ends with "restart Claude Code" — one restart is the floor `[verified: mcp-quickstart.md]`.
- A survey of the official marketplace and known community plugins found **no** plugin using SessionStart hooks for guided setup — the ecosystem norm is userConfig-or-README `[verified: survey of anthropics marketplace + community plugins, Aug 2026; absence claims are inherently best-effort]`.

## Desired outcome

- A user whose server is down for config reasons gets told so at session start, in-session, with an offer of help — instead of a silent dead connection.
- Asking for setup help (or accepting the offer) triggers an interactive flow where Claude discovers candidate vaults on disk (e.g. `.obsidian/` markers), validates the chosen path, writes the configuration where the plugin's `.mcp.json` expansion will find it, and ends with a clear "restart to connect."
- The same flow serves later-life repair (moved vault, expired git token), not only first run.
- The flow stays consistent with #1040 rather than competing with it: the config screen is the front door; this is the fallback and the recovery path, and the two must agree on where values live and which wins.

## Implementation ideas

Non-binding starting points:

- Two plugin artifacts: a `skills/vault-setup/SKILL.md` (interview → discover → validate → write settings `env` block → instruct restart) and a `hooks` SessionStart entry running a small doctor script from `${CLAUDE_PLUGIN_ROOT}/scripts/` that checks the effective config and emits guidance only when something is wrong.
- The doctor script could reuse the existing readiness/diagnostic knowledge (`get_server_info`, index/embeddings status semantics) in prose, without needing the server up.
- Precedence needs documenting: settings-file `env` beats shell env `[verified: env-vars.md]`, which is surprising to power users who configured via shell profile first.
- `[unverified]`: whether project-scope vs user-scope settings is the right default write target for a per-user vault path — likely user-scope, needs a decision.
- `[unverified]`: interaction with #1040's `${user_config.*}` values — which source wins when both are set is not documented and should be tested during implementation.
- Documentation touchpoints: `.claude-plugin/plugin/README.md`, `docs/guides/claude-code-plugin.md` (new "Troubleshooting / guided setup" section), README plugin section.

pvliesdonk:
Closes #1042. Part of epic #1043 — the fallback-and-recovery half of the plugin channel's install story (#1040's screen is the front door; this is what catches everything the screen can't).

## What changed

- **`scripts/doctor.sh` + `hooks/hooks.json`** — a SessionStart doctor. It resolves the effective `MARKDOWN_VAULT_MCP_SOURCE_DIR` exactly as the server does — the `~/.claude/settings.json` `env` block first (it reaches the server process and wins), shell environment as fallback — and speaks **only when something is wrong**: vault unset → "no vault configured, ask Claude to set up your vault"; configured directory gone → "vault moved, ask Claude to repair". Each message pairs a one-line `systemMessage` for the user with `additionalContext` telling the model to reach for the setup skill. Healthy configs produce zero output, so the hook is invisible in the normal case. All paths exercised locally, including both-sources-set-and-disagreeing (the settings value is the one diagnosed).
- **`skills/vault-setup/SKILL.md`** — the guided flow the doctor points at, usable for first-run setup *and* later-life repair: diagnose effective config (naming the settings-beats-shell precedence explicitly when both are set and disagree), discover candidate vaults (`.obsidian/` markers, note-like folder names, moved-vault matches), validate, write the **user-scope** settings `env` block (per the verified constraint that project-scope is wrong for per-user plugin config), and always finish with "restart Claude Code" — servers only start at session start, so one restart is the floor and the skill is forbidden from claiming success before it. Git-token and embedding-provider repairs are covered as variants of the same flow.
- **`packaging/pre-release-checks.sh`** — asserts the skill, the executable doctor, and the hooks entry all ship in the artifact.
- **Docs** — plugin README bootstrap section; a new "Troubleshooting and guided setup" section in `docs/guides/claude-code-plugin.md` (Vale-clean, including the precedence rule that surprises shell-profile users); design-doc decision-log entry #21 recording the architecture and why no server-side "setup mode" is needed (epic non-goal).

## Verification

- Doctor exercised end-to-end in all states (unset / missing dir / healthy / both-sources-disagreeing) — correct JSON hook output on the failure paths, silence on health, plus a plain-text fallback if `python3` is somehow absent.
- Full gates: 3271 tests passed, ruff check/format clean, mypy clean, diff-quality (no Python in diff), Vale 0 errors on the edited guide, `bash -n` + `jq` validation on the new script/config.
- Review round: the precedence-order finding from the first automated pass was fixed in commit `6b51195` and confirmed resolved by the follow-up review.

## Deliberately not done

- No `${user_config.*}` interaction test — the plugin's `userConfig` screen is #1040 (blocked on the v3.6.0 template tag); when it lands, the doctor's env-resolution order stays valid because `userConfig` values reach the server via `.mcp.json` substitution, not the settings `env` block.
- The generic template version of this pattern (a scaffolded doctor for any sibling) is the follow-up migration flagged in template#332's scope — proven here first, per the epic's ordering.
- The plugin README's pre-existing shell-profile setup instructions (the reviewer's non-blocking note) get modernized alongside #1040's `userConfig` screen, which changes that section anyway.

https://claude.ai/code/session_01LpzfNGAaCapWDf58vswzH3


pvliesdonk:
The review's non-flagged observation was correct and is now fixed: the doctor resolved the shell environment first, but the server's effective value is settings-beats-shell, so with both set to different paths the doctor could validate a path the server never uses. The follow-up commit reverses the resolution order (settings-file `env` first, shell fallback) and the both-set case is exercised — the settings value is the one diagnosed.

---

PR #1051, opened 2026-08-15
docs: prototype agent-written release-notes page for 3.1.0 (reference sample, not for merge)

### LINKED ISSUE #1059: docs/guides/transfer-links.md contradicts current behaviour after the pvl-core transfer adoption
## Summary

`docs/guides/transfer-links.md` still documents the retired local transfer
store. PR #981 (closes #979) adopted `fastmcp-pvl-core`'s transfer subsystem and
deleted `src/markdown_vault_mcp/transfer/`, but the guide was not updated, so it
now describes behaviour the server no longer has.

## Observed `[verified: git show origin/main]`

The guide has **zero commits** in `v3.1.0..origin/main` while the subsystem
underneath it was replaced. Four specific contradictions:

- **Line 52** — "After a successful download the token is consumed. A second
  request returns HTTP 404."
- **Lines 101-103** — a "One-time use (burn-on-success)" section: "A token is
  consumed on the **first successful** transfer."
- **Line 107** — "stale entries are purged from memory the next time a link is
  minted (no background thread is needed)."
- **Line 123** — "**In-memory token store.** Tokens are not persisted to disk.
  A server restart invalidates all outstanding tokens."

After #981 the store is KV-backed, so outstanding links **survive a restart**,
and a successful transfer **grace-settles** — the TTL shrinks to
`TRANSFER_GRACE_TTL_S` rather than the token burning immediately, so a stalled
transfer can retry.

Two environment variables introduced by the adoption are documented in
`docs/configuration.md` but absent from the guide:
`MARKDOWN_VAULT_MCP_TRANSFER_GRACE_TTL_S` and
`MARKDOWN_VAULT_MCP_TRANSFER_LEASE_S` `[verified: grep both files]`.

`[unverified]` One further behaviour change was reported during research but not
confirmed by me: that without `BASE_URL` set on an HTTP deployment the transfer
tools are now **not registered at all**, where previously they registered and
raised at call time. Worth checking while fixing the rest, since it changes what
an operator sees in the tool listing.

## Expected

The guide describes the shipped behaviour: persistence across restart, the
grace-settle semantics, and the full configuration surface.

## Why this is worth fixing now

The v4.0.0 release notes describe the new transfer behaviour. Linking readers to
a page that states the opposite would amplify the error rather than contain it,
so the notes currently describe the change **without** linking the guide.

## Documentation

- `docs/guides/transfer-links.md` — the fix.
- Correct behaviour is already recorded in `docs/configuration.md` and
  `docs/design/design.md`; the guide is the only stale surface found.

## Scope boundary

This page's accuracy. Not a behaviour change, and not a review of the other
guides — though the fact that a subsystem swap left a guide untouched suggests a
sweep for the same class of staleness would be worthwhile, as a separate issue.

pvliesdonk:
## Closes / Refs

Refs #1050.

> ## Not intended for `main`
>
> This is a **reference sample** for evaluating the agent-written release-notes approach, kept open so the output can be read and discussed against #1050. It is not proposed for merge.
>
> The two documentation fixes that briefly shared this branch have been split out to **#1061** (closes #1059, #1060) and are based on current `main`.

## What & why

Adds `docs/releases/3.1.md`, synthesised from the `v3.0.0..v3.1.0` range (72 commits, 71 linked issues, 64 PRs, 8 copier bumps) by walking commit → PR → linked issue and writing from the issue bodies rather than the commit subjects.

It exists to answer one question before any workflow is built: does digging into linked issues produce genuine rationale, or does the model pad?

Evidence that the corpus carries real signal — none of the following appears in any commit subject in the range:

- #822: 5,043 s to 3.2 s on a 4,400-document vault, 28 minutes wedged in one `opendir`
- #819: a 68,796-vector store destroyed by a 604-chunk watcher delta
- #820: the true best document at cosine 0.685 invisible at limit 3 through 12
- #731: about 190 false broken links from `\|` in tables
- #767: `VaultConfig` renamed to `ProjectConfig` with no alias, shipped in a *minor*
- Three of the reporters (@michael-denyer, @Finomosec, @Denzilla04) are outside contributors, which is the release's actual theme and is visible only from issue authorship

Also adds four domain terms to the Vale vocabulary (`accessor`, `backoff`, `globbed`, `subcommands`) surfaced while linting the page, and a `PATCH-RELEASES-START` / `-END` sentinel pair so later patch releases can be inserted without regenerating the narrative.

## Superseded by a second trial

A deeper trial has since been run against the unreleased `v3.1.0..main` range (95 commits) to draft the 4.0.0 release body. Its findings are recorded on `pvliesdonk/fastmcp-server-template#347` and materially change the implementation requirements — most importantly that a single agent holding a whole range will fall back to commit subjects, so the workflow must fan out per theme. That trial also surfaced the two documentation bugs now fixed in #1061.

Read this page for the *shape* and the evidence standard; treat #347 as the current state of the design.

- No workflow, action, or script — generation was done by hand so the output could be judged before committing to an implementation: #1050
- Not wired into `mkdocs.yml` `nav:` or the `llmstxt` `sections:` block, so the page builds but is unlisted: #1050
- No `docs/releases/index.md` landing page: #1050
- No change to `CHANGELOG.md` or any PSR configuration: #1050
- Covers one range only. Historical backfill is #1058

## Verification

`vale docs/releases/3.1.md` clean, `mkdocs build --strict` passes. Vale's `ai-tells` style caught two `VerbTricolon` hits and one `OverusedVocabulary` in the first draft, which is itself a finding — the repo already owns an LLM-prose gate pointed at exactly this directory.

## Docs impact

- [ ] `README.md`
- [x] `docs/` site pages
- [ ] `docs/design/`
- [ ] Inline docstrings

---
_Generated by [Claude Code](https://claude.ai/code/session_01AhD7QQxeanv14VyQdFaatF)_

### DIFF
--- a/.vale/styles/config/vocabularies/Base/accept.txt
+++ b/.vale/styles/config/vocabularies/Base/accept.txt
@@ -245,3 +245,9 @@ Zotero
 # --- Fonts / typography ---
 Newsreader
 Plex
+
+# --- Release-notes prose vocabulary ---
+accessor
+backoff
+globbed
+subcommands
--- a/docs/releases/3.1.md
+++ b/docs/releases/3.1.md
@@ -1,0 +1,319 @@
+# 3.1
+
+Released July 8, 2026.
+
+This is the release where other people's vaults broke it. Three outside
+reporters ran markdown-vault-mcp against a 4,400-document home directory, a
+symlink farm, and an Obsidian vault full of table navigators, and between them
+filed the issues behind most of what follows. The headline work is not a new
+feature: it is that discovery, watching and semantic recall now behave the same
+on a large real vault as they did on a small test one.
+
+Alongside that, the four in-panel MCP App views were rebuilt on a shared visual
+foundation, and two new read tools landed.
+
+## Upgrade notes
+
+Read these before upgrading. Two affect existing deployments.
+
+### `VaultConfig` is now `ProjectConfig`, with no alias
+
+The configuration dataclass was renamed from `VaultConfig` to `ProjectConfig`
+([#767](https://github.com/pvliesdonk/markdown-vault-mcp/issues/767)). No
+back-compatible alias was kept, so `from markdown_vault_mcp.config import
+VaultConfig` now raises `ImportError`. Only Python API consumers are affected.
+Environment variables, MCP tools and the CLI are unchanged.
+
+The rename was forced by an upstream template gate,
+`tests/test_config_wizard_drift.py`, which hard-imports `ProjectConfig` and
+could not collect against this repository at all. The same change taught the
+gate to see configuration reads that are delegated into `config_sections/`
+submodules, which it had previously been scanning past. Roughly 40 domain
+environment variables were unchecked before this landed.
+
+### Check `EMBEDDINGS_PATH` if it ends in `.npy`
+
+Setting `MARKDOWN_VAULT_MCP_EMBEDDINGS_PATH` to a path with a `.npy` extension
+could destroy the vector store
+([#819](https://github.com/pvliesdonk/markdown-vault-mcp/issues/819)). The
+loader probed for the sidecar by appending `.npy` to the configured base, while
+save and load derived it with `Path.with_suffix`. The two agreed only when the
+base had no extension. With `embeddings.npy` configured, the loader looked for
+`embeddings.npy.npy`, found nothing, cold-built an empty index, and the next
+save overwrote the real store.
+
+The reporter lost a 68,796-vector store to a 604-chunk watcher delta on the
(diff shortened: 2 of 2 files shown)


pvliesdonk:
Closing per the maintainer's decision on the epic plan review (#1054): this PR served its purpose — it answered the "does issue-walking produce genuine rationale" question and its findings are folded into pvliesdonk/fastmcp-server-template#347, which is the current design state. The branch `claude/auto-release-notes-generation-770xjo` is deliberately kept so `docs/releases/3.1.md` stays readable as the reference sample; it also remains input for the 3.1 page in the backfill (#1058). Closed, not merged; the branch should not be deleted.

---

PR #1061, opened 2026-08-15
docs: correct transfer-links after the pvl-core adoption; document EMBED_TIMEOUT_S

### LINKED ISSUE #1059: docs/guides/transfer-links.md contradicts current behaviour after the pvl-core transfer adoption

### LINKED ISSUE #1060: docs/guides/embeddings.md documents EMBEDDING_BATCH_SIZE but not EMBED_TIMEOUT_S
## Summary

PR #1000 (closes #954) added two related settings for slow embedding backends.
The embeddings guide picked up one of them and not the other, so the guide
covers the batch-size half of the fix and omits the timeout half.

## Observed `[verified: grep at origin/main]`

| Variable | `docs/guides/embeddings.md` | `docs/configuration.md` | `README.md` |
|---|---|---|---|
| `MARKDOWN_VAULT_MCP_EMBEDDING_BATCH_SIZE` | present | present | present |
| `MARKDOWN_VAULT_MCP_EMBED_TIMEOUT_S` | **absent** | present | present |

The guide is where an operator goes when embeddings are slow or failing, so it
is the surface where the timeout matters most.

## Why it matters

The reporter's symptom in #954 was specifically a **timeout**: on a CPU-only
Ollama running `bge-m3`, a single chunk takes 4-11 s, so a batch of four could
need 20-44 s against a hard-coded 30 s limit — 61 failures in one afternoon.
Raising `EMBED_TIMEOUT_S` is the direct remedy; shrinking the batch is the
secondary one. The guide currently documents only the secondary one.

## Expected

`docs/guides/embeddings.md` documents both settings together, with the
CPU-only guidance from #954 — raise the timeout to 60-120 s and/or drop the
batch to 1-2 — so an operator hitting the reported symptom finds the fix in the
guide rather than only in the configuration reference.

Worth noting alongside: FastEmbed runs in-process and ignores the timeout.

## Documentation

- `docs/guides/embeddings.md` — the fix.
- No change needed to `docs/configuration.md` or `README.md`; both already carry
  the variable.

## Scope boundary

One missing variable in one guide. Not a behaviour change, and not a review of
env-var documentation coverage generally — although this is the second
documentation gap found while drafting the v4.0.0 notes, which may argue for a
coverage check as its own issue.

pvliesdonk:
## Closes / Refs

Closes #1059, closes #1060.

## What & why

Two documentation-accuracy fixes, found while drafting the v4.0.0 release notes and verified against `docs/design/design.md` and the source.

### `docs/guides/transfer-links.md` (#1059)

The guide had **zero commits** in `v3.1.0..main` while #979/#981 replaced the transfer subsystem beneath it, so it still documented the retired local store. Four claims were wrong:

- **Burn-on-success → grace-settle.** A successful transfer no longer consumes the token outright; its remaining lifetime shrinks to `TRANSFER_GRACE_TTL_S` (default 60 s), so a served-but-stalled transfer can reclaim the link instead of being stranded.
- **Transient failure and crash behaviour** was undocumented. A failed request releases the reservation with the full remaining TTL; a crashed handler's in-flight reservation frees after `TRANSFER_LEASE_S` (default 60 s).
- **"In-memory token store … a server restart invalidates all outstanding tokens"** was listed as a *limitation* and is now false. The store is KV-backed, so live links survive a restart. Moved out of Limitations into its own Persistence section, since it is a property rather than a shortcoming.
- **"stale entries are purged from memory the next time a link is minted"** removed. The TTL is the security-relevant bound.

Both new variables added to the configuration table.

Also corrected the two registration bullets, which claimed the tools "raise an error" on stdio or without `BASE_URL`. `server.py:232` gates registration on `transport != "stdio" and config.server.base_url`, so the tools are **absent from the tool listing** rather than present and failing. This was filed as `[unverified]` on #1059 and is verified here.

### `docs/guides/embeddings.md` (#1060)

The guide documented `EMBEDDING_BATCH_SIZE` but not `EMBED_TIMEOUT_S`, and only framed batch size as a *memory* control. `EMBED_TIMEOUT_S` is the direct remedy for the symptom reported in #954 — 61 timeout failures in one afternoon on a CPU-only Ollama running `bge-m3` — and the guide is where an operator looks when embeddings time out.

Adds a **Slow or CPU-only backends** section covering both settings together, with the reported numbers (4–11 s per chunk, so a default batch of 4 can need 20–44 s against a 30 s budget), the CPU-only guidance, and a note that the two settings trade off against each other on throughput, so the timeout should be raised before the batch is shrunk.

- **No sweep of other guides** for the same class of staleness that #1059 represents. A subsystem swap leaving a guide untouched suggests others may be affected; that belongs in its own issue, noted on #1059.
- **No behaviour change.** Documentation only.
- **No changes to `docs/configuration.md`** — it already carried all three variables correctly. The gap was in the guides.

## Local review

- [x] Ran a local review pass on the cumulative diff before opening.

## Docs impact

- [ ] `README.md`
- [x] `docs/` site pages
- [ ] `docs/design/`
- [ ] Inline docstrings

## Verification

`vale` clean on both changed pages, `mkdocs build --strict` passes.

---

Split out of #1051 at the maintainer's request so these can merge independently of the release-notes prototype, which is not intended for `main`.

https://claude.ai/code/session_01AhD7QQxeanv14VyQdFaatF

---
_Generated by [Claude Code](https://claude.ai/code/session_01AhD7QQxeanv14VyQdFaatF)_


pvliesdonk:
Good catch, and correct. Fixed in 7d82b34.

Verified independently before acting: `README.md:533` did still carry the retired wording, and `README.md:257-258` already documented `TRANSFER_GRACE_TTL_S` and `TRANSFER_LEASE_S` correctly — so that one sentence was the last stale spot in the file. A sweep for the other retired claims (`in-memory token`, `restart invalidates`, `burn-on-success`) across `README.md` and the published `docs/` tree now returns nothing.

Wording aligned with the guide's "One-time use (grace-settled)" section rather than adopting the suggestion verbatim, so the two read consistently:

> Each token grants exactly one operation. On success the link is **grace-settled**: its remaining lifetime shrinks to `MARKDOWN_VAULT_MCP_TRANSFER_GRACE_TTL_S` (default 60 seconds), so a stalled transfer can still retry. A failed or interrupted transfer releases the reservation with the full remaining TTL, so retry is permitted until the TTL expires.

`vale README.md` clean.

Worth noting what this says about the underlying problem. The PR was opened to fix a guide that went stale when #979/#981 replaced the subsystem beneath it, and the fix itself repeated the same mistake one file over. That is the second occurrence of a single behaviour change leaving inconsistent documentation behind, which strengthens the case for the sweep noted on #1059 — the failure mode is a subsystem swap not knowing which prose describes it, and it will keep recurring until something checks.

---

PR #1064, opened 2026-08-16
feat(config): WRITE_PROTECT_EXISTING guard against blind overwrites

### LINKED ISSUE #1063: [Feature]: opt-in guard against write overwriting an existing file

On a writable vault, `write` on an existing path replaces the whole file,
frontmatter included, and there is no server-side setting that can stop it.
The two existing controls do not cover the case: `READ_ONLY` turns off every
write tool, and `if_match` only helps when the caller already decided to pass
an etag.

Where this bites: an LLM client reaching for `write` when `edit` was the right
tool. The tool docstring warns about it and the built-in prompts tell the model
not to overwrite without confirmation, but nothing enforces it — the note's
previous content is gone, and on a vault without git write-through there is
nothing to restore it from. I hit this on a personal vault used as long-term
notes storage; the operator who wants "this vault is append/edit-only for
agents" has no switch to set.

Observed on 3.0.3 and re-checked against `main` at 91f6a9c: no config key,
`Vault` argument, or `DocumentManager` guard exists for it.
[verified: grepped the config surface, `docs/configuration.md`, and the write
path in `managers/document.py`.]

### Desired outcome

An operator can configure a vault so that a `write` which would replace an
existing file is refused, while normal editing keeps working. Writes to new
paths, `edit`, `append`, `delete`, and `rename` should be unaffected, and a
caller who has actually read the file should still be able to replace it
deliberately. The default has to stay today's behaviour so existing
deployments do not change.

### Implementation ideas

Non-binding: a boolean env var next to `READ_ONLY` that makes `write` /
`write_attachment` raise `DocumentExistsError` when the target exists and no
`if_match` was supplied, since an `if_match` write already proves the caller
read the file.

### Documentation

`README.md` (env table), `docs/configuration.md`, `docs/tools/index.md`
(the `write` entry), `docs/design/design.md` (write semantics + env table),
and the generated env/wizard artifacts via `scripts/gen_config_surface.py`.

Finomosec:
## Closes / Refs

Closes #1063

## What & why

`write` replaces the whole file, so a client that reaches for `write` where
`edit` was meant destroys the note, and no server-side setting could prevent
it. This adds `MARKDOWN_VAULT_MCP_WRITE_PROTECT_EXISTING` (default `false`,
so nothing changes for existing deployments): with it on, `write` /
`write_attachment` raise `DocumentExistsError` when the target exists and no
`if_match` etag was supplied. An `if_match` write proves the caller read the
file first, so a deliberate replacement still works, and `edit`, `append`,
`delete`, and `rename` are untouched.

The guard runs before the `if_match` check, so a stale etag on an existing
file still reports `ConcurrentModificationError` rather than the new error.

Wiring follows `read_only` exactly: flat `ProjectConfig` field with `help` /
`tags` metadata → `to_vault_kwargs` → `Vault` → `DocumentManager`. The
generated artifacts (`.env.example`, `packaging/env.example`, the example
env files, `README.md`, `server.json`, `wizard-spec.json`) come from
`scripts/gen_config_surface.py`; `if_match` was added to the Vale vocabulary
because the generated README row spells it outside a code span.

The server's own upkeep of generated OKF files (`_okf_convention`'s `log.md`
bullet, `OkfMigrationManager`'s `index.md` regeneration and link conversion)
would otherwise have been caught by the guard: those flows read the file and
rewrite it in the same breath, without an etag. They now pass a keyword-only
`allow_overwrite=True` on `DocumentManager.write`, which is not reachable from
the `write` tool. `okf_verify` already passed `if_match` and needed nothing.

Note one consequence worth a maintainer's opinion: an upload through a
`create_upload_link` capability URL lands via `write_attachment`, so with the
flag on it cannot replace an existing attachment either. That reads as
consistent with the flag's intent, but it is a behaviour change for that route
and easy to scope out if you would rather it stay unguarded.

- No per-call override on the `write` tool (an `allow_overwrite` parameter):
  the point is that a model should not be able to opt out of the operator's
  policy. No tracking issue; say the word if you want one.
- No protection for `edit` / `append` / `delete` / `rename` — none of them
  replace unread content.
- No `okf_*` tool changes: `okf_seed_log` already refuses to overwrite, and
  `okf_generate_index` deliberately regenerates.

## Local review

Gates run locally: `uv run pytest -x -q` (3294 passed, 25 skipped),
`uv run ruff check .` + `ruff format --check .`, `uv run mypy src/ tests/`,
`uv run diff-quality --violations=ruff.check --options="--extend-select=C901,PLR0911,PLR0912,PLR0913,PLR0915,S" --fail-under=100`
(100%), `scripts/gen_config_surface.py --check` (regenerated in this commit),
and Vale over `docs/` + `README.md` (0 errors). New lines are covered by the
tests below.

## Docs impact

- [x] `README.md`
- [x] `docs/` site pages
- [x] `docs/design/`
- [x] Inline docstrings

`docs/configuration.md` gains the env-table row and a "Write protection"
section; `docs/tools/index.md` notes it on the `write` entry;
`docs/design/design.md` documents the `write()` semantics and the env var.

Tests: `TestWriteProtectExisting` in `tests/test_managers_document.py`
(new file, blind overwrite, valid/stale `if_match`, attachments, `edit` /
`append` unaffected, `allow_overwrite`, default off), in `tests/test_vault.py`
(Vault-level wiring), and in `tests/test_config.py` (env parsing +
`to_vault_kwargs`). `tests/test_okf_migrate.py` and
`tests/test_okf_convention.py` gain a protected-vault case each, so the
generated-file upkeep stays covered with the flag on.


pvliesdonk:
Maintainer decision on scope, so this can move:

- **The mechanism is accepted as designed** — refuse `write`/`write_attachment` on an existing target unless `if_match` is supplied, `allow_overwrite` kept internal, and the upload-link route staying guarded is the consistent call. `if_match` as the proof-of-intent is exactly right: it forces the caller to have actually read the note, which is the failure mode this flag exists for.
- **v4.1 ships it default-off, and 5.0 will flip the default to on.** That decision is tracked separately; please mention the planned flip in the PR body and in the docs' "Write protection" section (one sentence — "from 5.0 this guard is on by default"), so the release notes and operators see it coming.
- **One ask on the refusal message**: make sure `DocumentExistsError`'s text steers the agent to the sanctioned path — *"path exists; overwriting requires proof of read: call `read`, then retry with `if_match=<etag>` — or use `edit` for targeted changes"* — and not toward delete-then-recreate, which destroys first and proves nothing.
- **The branch needs a refresh**: it's based on `91f6a9c` and the 4.0 line has landed since (including #1119's `READ_ONLY` default flip, which touches the same config/generated surface), so it currently has conflicts. Please merge `main` in and re-run `scripts/gen_config_surface.py` rather than resolving the generated files by hand.

---

pvliesdonk:
Superseded by #1169, now merged — closing this as shipped rather than rejected.

Thank you, @Finomosec. Both of your commits were cherry-picked into #1169 with **authorship preserved**, so the history credits you for `WRITE_PROTECT_EXISTING` and for the OKF generated-file exemption. This branch sits on a fork and had conflicts against `main`, so carrying the commits forward was the way to ship the work without losing your name on it.

The maintainer's three review asks are applied:

- **The refusal message no longer points at delete-then-recreate.** It now steers to `read` + `if_match`, offers `edit` / `append`, and says outright not to delete and recreate, since that destroys the note first and proves nothing. A test locks the wording so it cannot regress quietly.
- **The 5.0 default flip is announced** in the docs' Write protection section, both env tables, the `write` tool entry, and the design spec, pointing at #1136.
- **The branch was refreshed** against `main` with the generated artifacts produced by `scripts/gen_config_surface.py` rather than resolved by hand.

One thing worth flagging, because it was a live trap: this branch predated #1119's `READ_ONLY` default flip and still carried `default=True` in `config.py` and two env tables. A naive merge would have silently reverted that flip. `main`'s `False` was kept in all three places.

The automated review added one finding, now fixed: `fetch` reaches the same guarded `write` / `write_attachment` path with an equally optional `if_match`, so `DocumentExistsError` is reachable through it too — its docstring and docs entry say so now. Your own scope calls all held up under review, including the `allow_overwrite` bypass staying unreachable from the public tools, and the guard running before the `if_match` check so a stale etag still reports `ConcurrentModificationError`.

On the open question in your description: the upload-link route stays guarded, which the maintainer confirmed as the consistent call.

---

PR #1118, opened 2026-08-20
feat(fetch)!: follow redirects, report the final URL, and classify the change

### LINKED ISSUE #1116: [Bug]: adopting pvl-core 4.11.2 flips the fetch tool from refusing redirects to following them, unclassified and with its new failure paths unhandled
## What happened?

Adopting `fastmcp-pvl-core` 4.11.2 changes what the `fetch` tool does: it stops refusing redirects and starts following them. The behaviour it replaces shipped in the last stable release, and the failure paths that replace it are neither handled, documented, nor tested.

## Observed

Attempted on branch `fix/1114-template-5.3.0-pvl-core-4.11.2` (commit `f1080b31`, unpushed). Three things surfaced, all consequences of the same posture change.

**1 — The removed behaviour is in the last stable release, not just the unreleased range.**

```
$ git show v3.1.0:src/markdown_vault_mcp/_server_tools/writer.py | grep follow_redirects
596:            httpx.AsyncClient(timeout=timeout_s, follow_redirects=False) as client,
```

`v3.1.0` (2026-07-08) is the newest non-prerelease tag; every 4.0.0 tag is an rc. Under 4.11.1 a redirecting URL raises `ValueError`; under 4.11.2 the same URL succeeds and returns bytes from whatever host the chain ended at. A caller that used `fetch` erroring on a redirect as a way to reject indirection gets content instead of an error, with no signature change to notice.

**2 — A failure mode that did not previously exist is unhandled.**

pvl-core 4.11.2's `fetch_url` docstring lists `httpx.TooManyRedirects: When the redirect chain exceeds httpx's max_redirects`. That class is unreachable in 4.11.1, whose `fetch_url` refused every 3xx up front with a `ValueError` the tool's existing `except ValueError` already caught. The tool's `Raises:` section does not mention it and no handler catches it. Sibling propagate-uncaught classes have tests (`test_fetch_http_error_status`, `test_fetch_timeout`); this one would not.

**3 — A comment written during the attempt describes upstream error text that does not exist.**

The attempt added: `# (e.g. a blocked-host refusal whose redacted URL contains the words)`. pvl-core's blocked-host errors carry a bare hostname, not a redacted URL:

```
f"Host {hostname!r} resolves to a non-public address ({ip}); "
"...are not allowed."
```

Both messages were also observed live earlier while probing the guard: `Host '127.0.0.1.nip.io' resolves to a non-public address (127.0.0.1); refusing to fetch.`

## Expected

The posture change is classified deliberately against the versioning policy before it ships, the failure surface it introduces is handled or documented as propagating, and prose describing upstream behaviour matches what upstream emits.

## Reproduction

Repo at `f4cf3977`; the attempt is `f1080b31` on `fix/1114-template-5.3.0-pvl-core-4.11.2`:

1. `git show v3.1.0:src/markdown_vault_mcp/_server_tools/writer.py | grep follow_redirects` → `follow_redirects=False`
2. Raise the `fastmcp-pvl-core` floor to `>=4.11.2` and re-lock
3. `uv run python -c "import inspect, fastmcp_pvl_core as c; print(inspect.getdoc(c.fetch_url))"` → the `Raises:` section lists `httpx.TooManyRedirects`
4. `grep -n "TooManyRedirects" src/markdown_vault_mcp/_server_tools/writer.py` → no match

## Suspected cause

`[verified: read v3.1.0's writer.py, pvl-core 4.11.2's fetch_url docstring and error-message source, and the current writer.py at f1080b31]`

The redirect posture is not a local decision — it lives entirely in pvl-core's `fetch_url`, so it changes as a side effect of the floor moving for an unrelated reason (the `build_instructions` signature, #1114). Nothing in this repo opts into it or gates it.

I have not verified whether any consumer actually relied on the refusal, only that the guarantee was documented in the tool's own docstring (`redirects are NOT followed`) and present in the last stable release.

## Open questions / scope

- The versioning policy in `CLAUDE.md` states that when only the MCP surface changes, "if the old behaviour is gone, treat it as breaking even though the schema re-discovers cleanly", assessed against the last stable release. Whether that lands on `!` here, or whether the prior refusal is judged an implementation detail nobody could rely on, is the call this issue exists to force — it should not be made first by a release-PR reviewer.
- `[unverified]` `httpx.HTTPStatusError` and `httpx.TransportError` are also documented as propagating from 4.11.2's `fetch_url`; whether their current handling is adequate was not examined.
- `[unverified]` The upgrade is a net security improvement in one respect — every redirect hop re-runs the resolve-validate-pin chain — but whether following redirects at all is the posture this vault should have was never decided, only inherited.
- Not about the `build_instructions` signature incompatibility or the template adoption, which is #1114.

Refs #1114.

---

pvliesdonk:
## Closes / Refs

Closes #1116
Refs #1114

Second of three (#1114 → #1116 → #1113). **#1117 is merged and this PR now targets `main` directly**; its diff is this change alone (4 files). #1119 is stacked behind this one.

## What & why

Adopting pvl-core 4.11.2 changes what `fetch` does: it stops refusing redirects and starts following them. This PR makes that call explicitly, rather than letting a release-PR reviewer discover it.

**The posture is not locally configurable.** 4.11.2 hardcodes `follow_redirects=True` in `fetch_url` with no opt-out parameter, so the only options were to accept it or re-assert the refusal domain-side by comparing `FetchResult.final_url` against the requested URL. **Accepting it**, for two reasons:

- **Per-hop revalidation is a stronger guarantee than a blanket refusal.** The guard is not a pre-flight check any more — it lives in `_GuardedTransport`, which httpx sends *every* redirect hop through. Each hop repeats the scheme check, the address validation, and the IP pin, so a `Location` pointing at an internal target is refused exactly as a directly supplied one is.
- **Ordinary indirection now resolves instead of erroring** — `http`→`https`, shortened links, CDN hand-offs. A refusal-based tool fails on a large share of real URLs.

### Why this carries `!`

Assessed against **v3.1.0**, the last stable release (every 4.0 tag is an rc), per the policy in `CLAUDE.md`:

```
$ git show v3.1.0:src/markdown_vault_mcp/_server_tools/writer.py | grep follow_redirects
596:  httpx.AsyncClient(timeout=timeout_s, follow_redirects=False) as client,
```

The tool's own docstring promised *"redirects are NOT followed"*. That behaviour is **gone and not reachable through any parameter** — not additive, not dual-mode. The policy is explicit that an MCP-surface change where the old behaviour is gone earns the marker even though the schema re-discovers cleanly. A caller that used a fetch error as a way to reject indirection now silently receives content, with no signature change to notice.

It rides inside the **unreleased 4.0 major** rather than forcing a 5.0 — same reasoning as #1113.

## Three consequences, handled

1. **`final_url` in the response.** The bytes need not come from the host in `url`, so validating the supplied URL is no longer sufficient on its own; pvl-core's own note says a caller enforcing a host policy must re-check `final_url`. It is surfaced rather than logged — `fetch_url` already logged the redacted source, and this string keeps its query string, which may be a pre-signed token.
2. **`httpx.TooManyRedirects` was unreachable before** — a redirect was refused up front with a `ValueError` the size-cap handler already caught. It is now documented as propagating and has a test, matching its `HTTPStatusError` and `TimeoutException` siblings, which were the only propagate-uncaught paths with coverage.
3. **The bogus comment is corrected.** The size-cap suffix match justified itself with *"a redirect refusal whose redacted URL contains the words"*. No such refusal exists any more, and pvl-core's blocked-host errors carry a bare hostname (`Host '127.0.0.1.nip.io' resolves to a non-public address…`), not a redacted URL — so the example was wrong on both counts. The rationale is restated on the message that genuinely does embed a caller-influenced URL: the size-cap error itself.

- **Does not re-assert the refusal, or add a knob for it.** Both were considered; see the reasoning above. A `FETCH_FOLLOW_REDIRECTS` env var would have kept the change non-breaking, at the cost of a permanent operator-surface knob, a config field, install-screen and wizard regeneration, and docs across every guide — for a mode the argument above says is the worse one.
- **Does not audit `httpx.HTTPStatusError` / `httpx.TransportError` handling**, listed `[unverified]` in #1116. Both predate 4.11.2, both already propagate, and both already have tests; nothing in this floor move changed them.
- **The floor move itself**: #1117 (merged).
- **The `READ_ONLY` default**: #1113 / #1119.

## Note on the `ci:` commit

The second commit on this branch is empty. It exists because CI genuinely never ran on the first one, and the recovery attempt made it worse:

- The branch was pushed *before* the PR was retargeted from its stacked base to `main`, so `synchronize` was filtered out by `ci.yml`'s `branches: [main, "release/**"]`; a base change fires `edited`, which is not one of `pull_request`'s default trigger types.
- A `workflow_dispatch` run then reported **`CI Success: success` while `PR Title` and `Structural Quality (diff)` were skipped** and the patch-coverage steps never ran — all three gated on `github.event_name == 'pull_request'`. A green aggregate with three required gates unexecuted is worse than no CI.

The empty commit fires `synchronize` against the correct base so the full PR-triggered set actually runs. Both previously-skipped gates now pass. The correct order — retarget first, then push — is what #1119 will use.

## Local review

- [x] Ran a local code-review pass on the cumulative diff.
- [x] The `!` breaks a surface that existed at the last stable release — verified by reading `v3.1.0`'s `writer.py`, not inferred. The `git tag --contains` sanity check does not apply inversely here: the surface being *removed* shipped in v3.1.0, so the marker is not spurious.

## Docs impact

- [ ] `README.md` — does not describe fetch's redirect behaviour
- [x] `docs/tools/index.md` — `url` parameter row corrected, `final_url` added to the return example, and a warning admonition covering the change, what replaces the guarantee, and the too-many-redirects failure
- [x] `docs/design/` — the SSRF paragraph no longer lists "redirect refusal"; a new paragraph records the posture, why it was adopted rather than inherited, and that it is pvl-core's to own
- [x] Inline docstrings — `url` argument, `Returns`, `Raises`, and both explanatory comments

## Gates

All 19 checks green on `2d8c3c2` (`deploy` correctly skipped). `claude-review`: no issues found — its passes independently confirmed the `!` justification, that `final_url` is returned and never logged, that `TooManyRedirects` falls outside the `except ValueError` handler, and that userinfo stripping in `final_url` holds against pvl-core's `_GuardedTransport`.

Locally: 3442 passed, ruff/mypy clean, 100% patch coverage, structural gate 100%, Vale clean.

### DIFF
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -2559,11 +2559,25 @@ overrides it.
   CGNAT/shared 100.64/10, and reserved ranges are all blocked, with NAT64
   embedded-IPv4 validation), DNS-rebind pinning of the validated IP,
   `trust_env=False` (ambient `HTTP(S)_PROXY`/`.netrc` cannot divert the
-  pinned connection), redirect refusal, a streaming size cap, and
-  userinfo/query redaction in logs and errors. The tool keeps only the
-  vault-side policy: which cap applies (attachment limit vs. uncapped
-  markdown) and the operator-facing size-limit message naming
-  `MARKDOWN_VAULT_MCP_MAX_ATTACHMENT_SIZE_MB`.
+  pinned connection), a streaming size cap, and userinfo/query redaction in
+  logs and errors. The tool keeps only the vault-side policy: which cap
+  applies (attachment limit vs. uncapped markdown) and the operator-facing
+  size-limit message naming `MARKDOWN_VAULT_MCP_MAX_ATTACHMENT_SIZE_MB`.
+
+  **Redirects are followed, and every hop re-runs that chain** (#1116).
+  Through 3.1 the primitive refused every 3xx outright; from the 4.11.2
+  floor the guard lives in the transport httpx sends each hop through, so a
+  `Location` pointing at an internal target is refused exactly as a directly
+  supplied one is. The posture is not locally configurable — pvl-core owns
+  it — and it was adopted deliberately rather than inherited silently: per-hop
+  revalidation is a stronger guarantee than a blanket refusal, and ordinary
+  indirection (`http`→`https`, shortened links, CDN hand-offs) now resolves
+  instead of erroring. The cost is that the bytes need not come from the host
+  in `url`, so the tool returns `final_url` — the URL the bytes actually came
+  from — and a caller enforcing a host policy must check it rather than
+  trusting the URL it supplied. A chain longer than httpx's `max_redirects`
+  raises `httpx.TooManyRedirects`, which propagates uncaught like the other
+  transport-tier failures.
 
 These semantics are intentionally close to Claude Code's file tools for
 familiarity. LLMs that know how to read/write/edit files can use these tools
--- a/docs/tools/index.md
+++ b/docs/tools/index.md
@@ -503,7 +503,7 @@ Download a file from a URL and save it to the vault as a note or attachment. Des
 
 | Parameter | Type | Default | Description |
 |-----------|------|---------|-------------|
-| `url` | string | required | Source URL to download. Only `http`/`https` schemes allowed; the host is resolved and blocked unless every address is publicly routable (private, loopback, link-local, CGNAT/shared, and reserved ranges are all refused); the validated IP is pinned for the connection; ambient `HTTP(S)_PROXY`/`.netrc` settings are ignored; redirects are not followed (SSRF protection) |
+| `url` | string | required | Source URL to download. Only `http`/`https` schemes allowed; the host is resolved and blocked unless every address is publicly routable (private, loopback, link-local, CGNAT/shared, and reserved ranges are all refused); the validated IP is pinned for the connection; ambient `HTTP(S)_PROXY`/`.netrc` settings are ignored. Redirects are followed, and each hop repeats every check above (SSRF protection) |
 | `path` | string | required | Destination path in vault. Extension determines handling: `.md` for notes, anything else for attachments |
 | `frontmatter` | object | `null` | Optional YAML frontmatter dict for `.md` files. Ignored for attachments |
 | `if_match` | string | `null` | Optional etag from a previous `read` call for optimistic concurrency |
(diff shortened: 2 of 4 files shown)


pvliesdonk:
**Why this PR shows no checks yet**

`ci.yml`'s `pull_request` trigger is filtered to `branches: [main, "release/**"]`, so a PR based on a topic branch gets no CI run at all — and the ruleset requires `CI Success` to merge. This one is based on `fix/1114-pvl-core-4.11.2` (#1117), so it stays check-less and unmergeable until it is retargeted to `main`.

That is the stack working as intended rather than a fault: merge #1117 first, retarget this to `main`, and CI runs against the correct base. The diff is already scoped to this change alone (4 files) — GitHub diffs against the merge base, so retargeting after #1117 lands does not widen it.

Retargeting now instead would run CI immediately but show #1117's 7 files on top of this PR's 4, which is the reviewability this split exists to avoid.

All seven gates were run locally against this branch; results are in the PR body.

---

PR #1027, opened 2026-08-12
chore(deps): update dependency openai to v3.3.1

pvliesdonk:
This PR contains the following updates:

| Package | Change | [Age](https://docs.renovatebot.com/merge-confidence/) | [Confidence](https://docs.renovatebot.com/merge-confidence/) |
|---|---|---|---|
| [openai](https://redirect.github.com/openai/openai-python) | `3.1.0` → `3.3.1` | age | confidence |

---

### Release Notes

<details>
<summary>openai/openai-python (openai)</summary>

### [`v3.3.1`](https://redirect.github.com/openai/openai-python/blob/HEAD/CHANGELOG.md#331-2026-08-19)

[Compare Source](https://redirect.github.com/openai/openai-python/compare/v3.3.0...v3.3.1)

##### Bug Fixes

- **deps:** update dependencies with published security fixes ([#&#8203;3680](https://redirect.github.com/openai/openai-python/issues/3680)) ([53aa4fc](https://redirect.github.com/openai/openai-python/commit/53aa4fc68b65f42456cafaaa5ff4b0d317184752))

##### Chores

- **build:** migrate to uv ([#&#8203;3653](https://redirect.github.com/openai/openai-python/issues/3653)) ([b37e85d](https://redirect.github.com/openai/openai-python/commit/b37e85d8fe9cd2b23862dbc62716d38e2cacaf92))
- **deps:** remove jsonschema and unused fixture-validation dependencies ([0dfdfdd](https://redirect.github.com/openai/openai-python/commit/0dfdfdddfbeb591a1bb5b3b4aa62e4a9870fcaa6))
- lock the repository Pyright toolchain ([#&#8203;3678](https://redirect.github.com/openai/openai-python/issues/3678)) ([3079be2](https://redirect.github.com/openai/openai-python/commit/3079be224c05479b6cf8aa329480c249f66f0599))
- run the mock server from locked local tooling ([#&#8203;3679](https://redirect.github.com/openai/openai-python/issues/3679)) ([370fcc6](https://redirect.github.com/openai/openai-python/commit/370fcc60ce9006db5a75dcd0333790dfe71a907c))

##### Refactors

- **deps:** use the standard library for platform detection ([d5b0065](https://redirect.github.com/openai/openai-python/commit/d5b00659d6659c5def665cb04e6563bdf683ea03))

##### Build System

[...]

PR #1021, opened 2026-08-12
chore(deps): update dependency ruff to v0.16.4

pvliesdonk:
This PR contains the following updates:

| Package | Change | [Age](https://docs.renovatebot.com/merge-confidence/) | [Confidence](https://docs.renovatebot.com/merge-confidence/) |
|---|---|---|---|
| [ruff](https://docs.astral.sh/ruff) ([source](https://redirect.github.com/astral-sh/ruff), [changelog](https://redirect.github.com/astral-sh/ruff/blob/main/CHANGELOG.md)) | `0.16.3` → `0.16.4` | age | confidence |

---

### Release Notes

<details>
<summary>astral-sh/ruff (ruff)</summary>

### [`v0.16.4`](https://redirect.github.com/astral-sh/ruff/blob/HEAD/CHANGELOG.md#0164)

[Compare Source](https://redirect.github.com/astral-sh/ruff/compare/0.16.3...0.16.4)

Released on 2026-08-20.

##### Preview features

- \[`flake8-use-pathlib`] Add autofix for `PTH116` ([#&#8203;26460](https://redirect.github.com/astral-sh/ruff/pull/26460))
- \[`refurb`] Restrict `delete-full-slice` to lists (`FURB131`) ([#&#8203;27711](https://redirect.github.com/astral-sh/ruff/pull/27711))
- \[`refurb`] Skip `FURB101` and `FURB103` when the `open` argument is a file descriptor ([#&#8203;27643](https://redirect.github.com/astral-sh/ruff/pull/27643))

##### Bug fixes

- Fix `InvalidInstruction` on Windows CPUs that do not support `POPCNT` ([#&#8203;27803](https://redirect.github.com/astral-sh/ruff/pull/27803))
- \[`pyflakes`] Emit semantic syntax errors in string type definitions as `F722` ([#&#8203;27835](https://redirect.github.com/astral-sh/ruff/pull/27835))
- \[`pylint`] Allow `os._exit` imports in `import-private-name` (`PLC2701`) ([#&#8203;27738](https://redirect.github.com/astral-sh/ruff/pull/27738))

##### Rule changes

- \[syntax-errors] Align mixed t-string/bytes error message with CPython 3.14 ([#&#8203;27766](https://redirect.github.com/astral-sh/ruff/pull/27766))
- \[`ruff`] Add `ctypes.LittleEndianStructure` and related types to existing exception (`RUF012`) ([#&#8203;27753](https://redirect.github.com/astral-sh/ruff/pull/27753))
[...]

PR #1150, opened 2026-08-24
chore(deps): update dependency pygments to >=2,<2.22

pvliesdonk:
This PR contains the following updates:

| Package | Change | [Age](https://docs.renovatebot.com/merge-confidence/) | [Confidence](https://docs.renovatebot.com/merge-confidence/) |
|---|---|---|---|
| [pygments](https://redirect.github.com/pygments/pygments) ([changelog](https://redirect.github.com/pygments/pygments/blob/master/CHANGES)) | `>=2,<2.21` → `>=2,<2.22` | age | confidence |

---

### Release Notes

<details>
<summary>pygments/pygments (pygments)</summary>

### [`v2.21.0`](https://redirect.github.com/pygments/pygments/blob/HEAD/CHANGES#Version-2210)

[Compare Source](https://redirect.github.com/pygments/pygments/compare/2.20.0...2.21.0)

(released August 17th, 2026)

- New lexers:

  - BitBake ([#&#8203;3103](https://redirect.github.com/pygments/pygments/issues/3103))
  - Caddyfile ([#&#8203;3225](https://redirect.github.com/pygments/pygments/issues/3225))
  - CEL ([#&#8203;3048](https://redirect.github.com/pygments/pygments/issues/3048))
  - PureScript ([#&#8203;1077](https://redirect.github.com/pygments/pygments/issues/1077), [#&#8203;3054](https://redirect.github.com/pygments/pygments/issues/3054))

- Updated lexers:

  - Bash: Fix coloured keyword at the beginning of a name ([#&#8203;2926](https://redirect.github.com/pygments/pygments/issues/2926))
  - Boogie: Add missing Boogie and Civl Verifier keywords ([#&#8203;3156](https://redirect.github.com/pygments/pygments/issues/3156))
  - C#:
    - Recognize interpolated verbatim strings with either `$@` or `@$`
      prefixes ([#&#8203;2685](https://redirect.github.com/pygments/pygments/issues/2685))
    - Support dollar-prefixed and multi-quote raw strings ([#&#8203;3129](https://redirect.github.com/pygments/pygments/issues/3129), [#&#8203;2897](https://redirect.github.com/pygments/pygments/issues/2897))
    - Recognize `union` ([#&#8203;3182](https://redirect.github.com/pygments/pygments/issues/3182))
  - C/C++:
    - Add C23/C++26 attributes ([#&#8203;3084](https://redirect.github.com/pygments/pygments/issues/3084))
[...]

PR #1081, opened 2026-08-17
chore(deps): update dependency numpy to v2.5.2

pvliesdonk:
This PR contains the following updates:

| Package | Change | [Age](https://docs.renovatebot.com/merge-confidence/) | [Confidence](https://docs.renovatebot.com/merge-confidence/) |
|---|---|---|---|
| [numpy](https://redirect.github.com/numpy/numpy) ([changelog](https://numpy.org/doc/stable/release)) | `2.4.6` → `2.5.2` | age | confidence |

---

### Release Notes

<details>
<summary>numpy/numpy (numpy)</summary>

### [`v2.5.2`](https://redirect.github.com/numpy/numpy/releases/tag/v2.5.2): (Aug 9, 2026)

[Compare Source](https://redirect.github.com/numpy/numpy/compare/v2.5.1...v2.5.2)

##### NumPy 2.5.2 Release Notes

The NumPy 2.5.2 is a patch release that fixes bugs discovered after the 2.5.1
release. The big news is that it includes wheels for the newly released
Python 3.15.0rc1.

This release supports Python versions 3.12-3.15

##### `PyArray_StringDTypeObject` is opaque under the abi3t stable ABI

The `PyArray_StringDTypeObject` was accidentally exposed in NumPy
2.5 when targeting the free-threading-compatible stable ABI
(`Py_TARGET_ABI3T`). `PyArray_StringDTypeObject` is now an opaque
struct: extensions compiled that way cannot access its fields, since
the struct layout depends on the size of the object header. Any code
that accessed `PyArray_StringDTypeObject` fields in an abi3t build
would have crashed, so we are making this API change in a bugfix
release.

The `NpyString` allocator API remains usable by passing the
descriptor object pointer, e.g.
`NpyString_acquire_allocator((PyArray_StringDTypeObject *)descr)`.

([gh-31771](https://redirect.github.com/numpy/numpy/pull/31771))

##### Contributors

A total of 16 people contributed to this release. People with a "+" by their
names contributed a patch for the first time.

- Abhijeetsingh Meena +
- Charalampos Stratakis
- Charles Harris
- Chris Ninham +
- David Woods
- Geonho +
- Gopu Yeshwanth Reddy +
- Iason Krommydas
- Ijtihed Kilani
- Jelle Zijlstra +
- Joren Hammudoglu
- Kumar Aditya
- Mike Boyle
- Nathan Goldbaum
- Raghuveer Devulapalli
- Sebastian Berg

##### Pull requests merged

[...]

PR #1163, opened 2026-08-25
fix: v4.1 bug sweep — rename staging, webhook gating, blank query, changelog

### LINKED ISSUE #894: bug: _stage_and_commit rename path runs `git add -u` with no pathspec, sweeping unrelated modifications into the auto-commit
## What
The rename path of `_stage_and_commit` stages with `git add -u` and **no pathspec**, so every tracked modification anywhere in the repository — including files the MCP server never touched (e.g. concurrent Obsidian edits not yet committed) — is swept into the rename's auto-commit. It also misses never-committed old paths (untracked files being renamed are not covered by `-u`).

Documented only in a NOTE comment in the code, so it is invisible to TODO greps; surfaced by the 2026-07 maintainability audit (#879) and deferred from PR #892 because fixing it **changes commit contents**, i.e. it is not a behavior-preserving refactor.

## Where
`src/markdown_vault_mcp/git/strategy.py`, `_stage_and_commit` rename branch (post-#892 line numbers near the module-level commit helper; pre-#892 reference: :1902–1926 at HEAD `a06be7f`).

## Why it compounds
Auto-commits attributed to a rename can silently capture unrelated working-tree changes, polluting history and — worse — pushing content the user had not intended to commit yet. Any future conflict-sibling or revert tooling that trusts "one commit = one MCP operation" inherits the corruption.

## Suggested direction
Pass explicit pathspecs: stage `old_path` (as a deletion; `git add -A -- <old> <new>` handles both tracked and untracked cases) and `new_path` only. Add a characterization test: dirty unrelated tracked file + MCP rename → the rename commit must not contain the unrelated file, and the working tree keeps its dirty state.

## Documentation
- `docs/design/design.md`: note the per-operation staging guarantee in the git-write strategy section once fixed.

Refs #879, #883.

### LINKED ISSUE #1110: Flaky: test_vault_search_respects_limit queries before the writer drains
## What happened?

`tests/test_mcp_apps_browser.py::TestBrowserDataTools::test_vault_search_respects_limit` failed once in a full local run at `4c449dd` (branch `claude/v4-0-bug-reports-0ye6vc`), and passed on an immediate re-run of the same tree, both for the whole suite and for the file and test alone. Observed once in five full-suite runs on that branch.

Not caused by the changes on that branch: the failing test exercises the `vault_search` app tool with `mode="keyword"`, and it also passed in three earlier full-suite runs across the same series of commits.

## Suspected cause

`[unverified]` The test is one of the few in `TestBrowserDataTools` that does **not** call `wait_for_mcp_writer_drain(client)` before querying — its neighbours `test_notes_have_kind_field` and the `vault_list` tests do. A `make_server()` fixture whose index writer has not drained would answer from a partially built index, which is exactly the shape of a query-side flake that reproduces once in five runs and never in isolation.

I did not capture the assertion output (the run's tail was truncated past the failure), so I have not confirmed whether it failed on `len(data) <= 2` or inside `_parse_tool_data`. The `<= 2` assertion cannot fail from having *too few* results, which points at the parse rather than the bound.

## Suggested direction

Add the same `await wait_for_mcp_writer_drain(client)` the neighbouring tests use, then re-run the file in a loop to confirm. If it reproduces with the drain in place, the readiness gate itself is the thing to look at rather than the test.

## Related

Second flake of this shape on the suite; #1091 is the other one (a settrace race under coverage on 3.11), unrelated mechanically.

### LINKED ISSUE #1111: An empty or whitespace-only semantic search query is sent to the embedding provider and returns a raw 400
## What happened?

`search(query, mode="semantic")` embeds the query verbatim. An empty or
whitespace-only query therefore reaches the provider as `[""]`, which every
OpenAI-compatible endpoint rejects with HTTP 400 — the same rejection as
#1087, on the read side rather than the index side.

Found while fixing #1087; deliberately left out of that PR, which is scoped
to the indexing paths.

## Observed

`[verified: reproduced on `main` @ `1d78f5c` with a provider that rejects
blank input the way a real endpoint does]`

```
query='':    RuntimeError: OpenAI API error 400: Input cannot contain empty strings
query='   ': RuntimeError: OpenAI API error 400: Input cannot contain empty strings
```

`VectorIndex.search` (`vector_index.py:391`) calls `self._provider.embed([query])`
with no blank check. Its neighbouring guard covers `limit` only:

```python
        if limit <= 0:
            ...
            return []

        raw = self._provider.embed([query])
```

Nothing upstream fills the gap — `SearchManager` and the `search` tool
registration have no `query.strip()` check `[verified: grep over
managers/search.py and _server_tools/reader.py]`.

## Why this is much less severe than #1087

Worth stating so it is not over-prioritised: this is a read path. The call
fails, the caller gets an error, and nothing persists. There is no stale
index, no silent divergence, and no recurrence on later passes — the three
properties that made #1087 a release blocker. `mode="keyword"` and
`mode="hybrid"` are unaffected on the FTS side.

The cost is a confusing error surface: an LLM client that sends an empty
query gets a provider-shaped HTTP 400 rather than a statement about its own
input, and on a self-hosted endpoint the message wording varies by vendor.

## Expected

An empty or whitespace-only query resolves without a provider round-trip.
Two defensible shapes, maintainer's call:

1. **Return `[]`**, matching the adjacent `limit <= 0` guard's posture in the
   same method — cheapest, and consistent with what is already there.
2. **Raise a domain error** naming the empty query, so a caller learns it sent
   a bad argument instead of reading a vendor's 400.

Whichever is chosen should apply at the manager boundary rather than only in
`VectorIndex`, so `mode="hybrid"` takes the same path.

## Environment

Any configuration using a network embedding provider (`openai`, `ollama`, and
the Voyage-via-`OPENAI_BASE_URL` setup from #1087). FastEmbed runs in-process
and may accept a blank input, so the failure is provider-dependent — another
reason to settle it in first-party code.

### LINKED ISSUE #1123: The changelog reconstruction dropped every hand-written BREAKING block, and six release sections are now empty
### What happened?

`CHANGELOG.md` at HEAD carries no breaking-change warnings at all. The
hand-written `### Changed (BREAKING)` blocks that shipped in the tags were
replaced by generated commit subjects, and the operator migration instructions
they contained are gone.

### Observed

At `v3.0.0` the changelog carried an eight-entry `### Changed (BREAKING)`
block. One entry read:

> Environment variable `MARKDOWN_VAULT_MCP_READY_TIMEOUT_S` renamed to
> `MARKDOWN_VAULT_MCP_BUILD_TIMEOUT_S`. Running deployments that set the old
> variable will silently fall back to the 60-second default after upgrading;
> update operator configs (compose files, systemd units, `.env` files)
> accordingly.

The block also covered `Collection.get_index_status`'s `"ready"` →
`"queryable"` value change and its priority flip, `is_index_ready()` →
`is_queryable()`, `wait_for_index_ready()` → `wait_until_queryable()`,
`needs_index_ready` → `needs_queryable`, `_server_readiness.py` →
`_server_queryable.py`, and `IndexNotReadyError` → `IndexUnavailableError`,
closing with "External consumers ... must rename their references. No
deprecation shims ship."

At HEAD:

```
$ grep -c BREAKING CHANGELOG.md
0
$ grep -c READY_TIMEOUT_S CHANGELOG.md
0
$ grep -n -A3 "^## 3\.0\.0" CHANGELOG.md
923:## 3.0.0 (2026-06-17)
924-
925-
926-## 2.0.0-rc.5 (2026-06-17)
```

The `3.0.0` section is empty. Surviving mentions of the renamed symbols are
commit subjects, not warnings — e.g. line 1164,
`- **docs+server**: Catch remaining 'ready'/IndexNotReadyError stragglers +
document get_index_status`, which names the old symbol without saying anything
broke.

**This is a class, not one entry.** Six release sections in the current file
are empty:

```
$ awk '/^## [0-9]/{if(h && !c) e++; h=$0; c=0; next} /[^[:space:]]/{if(h)c++} END{if(h && !c) e++; print e+0}' CHANGELOG.md
6
```

`[verified: commands above, run 2026-08-20 against HEAD and `git show
v3.0.0:CHANGELOG.md`]`

### Expected

A reader of `CHANGELOG.md` looking for what broke in a release finds the
breaking changes for that release. At minimum, a release that shipped breaking
renames does not present an empty section.

### Reproduction

```bash
git show v3.0.0:CHANGELOG.md | sed -n "/BREAKING/,/^## /p"   # the block, as shipped
grep -c BREAKING CHANGELOG.md                                 # 0 at HEAD
```

### Suspected cause

`[unverified]` The 2026-08-16 reconstruction in PR #1067 (`3e18812`,
`bf848cc`), which replaced the hand-maintained Keep-a-Changelog file with a
form generated from commit subjects. Both commits remove a `READY_TIMEOUT_S`
line. I have not verified whether the loss was a deliberate accepted
consequence or collateral — the commit message describes a wholesale
replacement and does not mention the block.

I have not verified the cause.

### Open questions / scope

- `[unverified]` Whether the six empty sections all lost hand-written content,
  or whether some were empty at their tag too. Only `3.0.0` was checked
  against its tag.
- `[unverified]` Whether any generated section is *wrong* rather than merely
  thin. Not examined.
- Out of scope: whether the generated format should be replaced. This records
  that content which existed at the tags is absent at HEAD.

This surfaced from a support question — someone set
`MARKDOWN_VAULT_MCP_READY_TIMEOUT_S` and found it had no effect, which is the
exact failure the deleted entry predicted. Grepping today's changelog for the
variable returns nothing.

---

### LINKED ISSUE #1128: [Bug]: a webhook delivery outside managed git mode fetches anyway — 503, or an unhandled CalledProcessError on a non-git vault
## What happened?

Expected `GITHUB_WEBHOOK_SECRET` set on a deployment that is not in managed git mode to be inert; got a real `git fetch origin` per delivery, answered 503, and an unhandled exception out of the handler when the vault is not a git repository.

## Observed

`to_vault_kwargs` builds a `GitWriteStrategy` on every path, including the unmanaged / commit-only fallback (`src/markdown_vault_mcp/config_sections/_assembly.py:291-300`, `enable_sync=False` → `enable_pull=False`). `Vault._git_strategy` is therefore never `None` for a config-driven server, and `GitWriteStrategy.force_pull` → `_pull_pipeline` runs `git fetch origin` without consulting `enable_pull` (that flag gates the periodic loop only).

Exercised both non-managed shapes directly against 563affd:

```python
s = GitWriteStrategy(token=None, repo_url=None, managed=False,
                     enable_pull=False, enable_push=False, repo_path=<dir>)
s.force_pull()
```

- **git repo with no `origin`** (unmanaged / commit-only):

  ```
  Git force_pull: fetch failed: fatal: 'origin' does not appear to be a git repository
  fatal: Could not read from remote repository.
  norem -> applied=False reason=fetch_failed
  ```

  `applied=False` reaches `_github_webhook.py`'s "pull not applied" branch, which answers **503** so GitHub retries. Every push to the repository therefore spends the full retry budget and fails.

- **plain directory, not a git repo** (no-git mode):

  ```
  plain -> raised CalledProcessError Command '['git', '-C', '<dir>', 'rev-parse', 'HEAD']' returned non-zero exit status 128
  ```

  `_pull_pipeline`'s `_head_sha` call raises before the fetch. The webhook handler does not catch it, so the exception leaves the handler.

The handler's own docstring enumerates its outcomes as 401 / 200 / 503, and carries a `pull_result is None` → 200 "no git strategy" branch that a config-driven server cannot reach.

## Expected

A webhook secret set on a deployment with no managed remote either refuses at configuration time, or answers deliveries without running git and without an unhandled exception.

## Reproduction

1. Check out 563affd
2. Run the snippet above against (a) a `git init` repo with no remote and (b) a plain directory
3. Observe `applied=False reason=fetch_failed` and the raised `CalledProcessError` respectively

## Suspected cause

`[verified: read config_sections/_assembly.py, git/strategy.py force_pull/_pull_pipeline/_resolve_force_repo, _github_webhook.py, and ran force_pull against both directory shapes at 563affd]`

`enable_pull` is not consulted on the `force_pull` path, and the handler's `None` branch anticipates a strategy that config never leaves unset. I verified the behaviour, not the intent behind either.

## Open questions / scope

- `[unverified]` Whether the `git_sync` tool reaches the same two outcomes. It is hidden without `GIT_REPO_URL`, so it may be unreachable in these configurations; I did not check its gating.
- `[unverified]` Whether the "no git strategy" 200 branch is reachable by a library consumer constructing `Vault` directly, rather than dead code.
- Not about managed mode, where the webhook behaves as documented.

Refs #530, #1126.

pvliesdonk:
## Closes / Refs

Closes #894, closes #1128, closes #1111, closes #1110, closes #1123.

## What & why

The five `bug`-labelled issues on the **v4.1** milestone, plus a `uv.lock` refresh.

**#894 — the rename auto-commit swept unrelated changes.** The rename branch of `_stage_and_commit` staged with a pathspec-less `git add -u`, so every tracked modification in the repository landed in the rename's commit: a concurrent Obsidian edit the server never saw was committed and pushed. `-u` also recorded no deletion for an old path git had never tracked, leaving the commit with only the new path.

A rename has two sides but `WriteCallback` carries one path, so the old path is now threaded through. Because `WriteCallback` is a public type a downstream consumer implements, the extra argument is an **opt-in rather than a signature change**: a callback sets `accepts_old_path` (`types.ACCEPTS_OLD_PATH_ATTR`), and both dispatch seams — `WriteCallbackDispatcher` and `DocumentManager` — read it once and pass `old_path=` only to callbacks that advertise it. A three-argument callback is unaffected at type-check time and at run time. `GitWriteStrategy` opts in and stages `git add -A -- <old> <new>`; an untracked old path is dropped from the pathspec, because `git add` fails the whole invocation on a pathspec matching nothing, which would leave the new path unstaged too.

**#1128 — a webhook delivery outside managed git mode ran git anyway.** `enable_pull` gated only the periodic loop, so `force_pull` ran the pipeline in unmanaged / commit-only mode: `git fetch origin` against a remoteless checkout (answering the retryable-looking `fetch_failed`, so every push burned GitHub's full retry budget on a 503), and an unhandled `CalledProcessError` out of `_head_sha` against a vault that is not a git repository at all. `force_pull` now returns `pull_disabled` before running any git command. The handler answers 200 for it — a configuration state, not a failure — wraps `force_pull` so nothing escapes as a 500, and the server warns once at startup when the secret is set with no remote.

Both open questions in the issue are answered: `git_sync` cannot reach these outcomes (`_resolve_managed_strategy` refuses outside managed mode, and managed mode always has `enable_pull=True`), and the `pull_result is None` branch is not dead code — a library consumer constructing `Vault(git_strategy=None)` reaches it.

**#1111 — a blank semantic query reached the embedding provider.** An empty or whitespace-only query was embedded verbatim as `[""]`, which every OpenAI-compatible endpoint rejects with a raw HTTP 400. The guard sits at the manager boundary so hybrid takes the same path as semantic, with a backstop in `VectorIndex.search` for direct library consumers. It runs *after* `_require_vectors()`, so an unconfigured provider still raises rather than silently answering `[]`. Took the issue's option 1 (return `[]`), matching the adjacent `limit <= 0` guard.

**#1110 — flaky `test_vault_search_respects_limit`.** Added the `wait_for_mcp_writer_drain(client)` its neighbours in `TestBrowserDataTools` already call.

**#1123 — the changelog lost every breaking-change block.** The 2026-08-16 reconstruction generated pre-4.0 sections with only Features / Bug Fixes / Chores headings, flattening sixteen `!` markers in the 1.28.0 → 3.0.0 range and dropping the operator migration instructions that had shipped in the tagged changelogs — so 3.0.0, a major release, presented an empty section and `grep READY_TIMEOUT_S CHANGELOG.md` returned nothing. Restored 3.0.0's block from the `!` commits in its range plus prose recovered from `v3.2.0-rc.7:CHANGELOG.md`.

The five sections still empty (`3.2.0-rc.6`, `3.1.0`, `1.27.0`, `1.25.0`, `1.23.1`) are **promotion-only ranges** — their whole diff is the release commit and the work sits in the `-rc` sections beneath them, so there is nothing lost to restore. The preamble now names that shape, and a test fails if an empty section is ever left unexplained. This is a one-time repair of historical sections; knope only ever writes new sections below the flag, so it does not fight the tooling.

**`uv.lock` refresh** — a single `uv lock --upgrade`, which subsumes #1027 (openai 3.3.1) and #1021 (ruff 0.16.4) plus transitive bumps, and drops `distro`. Deliberately unmoved: numpy (`constraint-dependencies`, #727), pygments (`override-dependencies`), mypy (#727).

- **Auditing the remaining historical `!` markers**: #1077. This PR restores 3.0.0's block; the systematic pre-3.1.0 sweep is that issue's scope.
- **Backfilling the other five empty changelog sections with content**: they are promotion-only ranges with nothing to backfill. Documented in the preamble instead.
- **Refusing at configuration time when a webhook secret is set with no remote**: #1128 offered that as an alternative. A startup failure for a config that boots today is a bigger operator-facing change than the defect warrants; the endpoint stays mounted, answers 200, and warns.
- **Lifting the numpy / pygments / mypy pins**: #727 and the `override-dependencies` comment own those.
- **Touching the other open dependency PRs.** Triage in the comment below — several are no-ops or already-applied and want closing, which is the maintainer's call.

## Local review

- [x] No commit carries `!`. `WriteCallback`'s three-argument shape is unchanged and a downstream callback keeps working untouched; `PullResult.reason` gained a value, which is additive.

## Docs impact

- [ ] `README.md` — no user-facing surface changed that it documents.
- [x] `docs/` site pages — `docs/configuration.md` and `docs/guides/git-integration.md` (webhook outside managed mode; the guide's existing "Managed mode only" warning described the bug as a known limitation and now describes the fixed behaviour), `docs/tools/index.md` (blank-query result).
- [x] `docs/design/` — per-operation staging guarantee (#894), `enable_pull` gating every pull (#1128), the read-side blank-input guard alongside the existing #1087 section (#1111).
- [x] Inline docstrings — `_stage_and_commit`, new `_stage_rename` / `_is_tracked` / `_fire_rename`, `force_pull`, `make_webhook_handler`, `WriteCallbackDispatcher.__init__` / `fire`, `PullResult.reason`.

## Verification

| Gate | Result |
|---|---|
| `uv run pytest -q` | 3498 passed, 15 skipped |
| `uv run mypy src/ tests/` | no issues, 192 files |
| Patch coverage (`diff-cover` vs `origin/main`) | **97%** (2 missed lines are `Protocol` `...` bodies) |
| `bash scripts/structural_gate.sh` | 100%, 0 violations on 983 diff lines |
| `uv run pre-commit run vale --all-files` | passed |

The test suite was also run in full on the upgraded lock resolution before it was committed.

---
_Generated by [Claude Code](https://claude.ai/code/session_0149HEqn5kPBqS8RXJnDgiqX)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,25 @@
 # CHANGELOG
 
+Machine-generated audit trail. Sections from `4.0.0-rc.1` onward are written
+by knope into each release PR from `feat` / `fix` / `!` commit subjects —
+never hand-edit those, or the flag line below.
+
+Two things to know when reading further down:
+
+- **Sections at and below `3.2.0-rc.7` were reconstructed** from commit
+  subjects in #1067, replacing a hand-maintained Keep a Changelog file whose
+  entries all lived under a single `## [Unreleased]` heading. The
+  reconstruction emitted no `Breaking Changes` heading, so pre-4.0 `!`
+  markers were flattened; #1123 restored the `3.0.0` block by hand, and
+  #1077 tracks auditing the rest of the historical markers.
+- **A stable release section can be empty** when its whole range is the
+  promotion commit — the work sits in the `-rc` sections directly beneath
+  it. `3.2.0-rc.6`, `3.1.0`, `1.27.0`, `1.25.0`, and `1.23.1` are all this
+  shape.
+
+Narrative, rationale, and upgrade guidance live in
+[`docs/releases/`](docs/releases/), not here.
+
 <!-- version list -->
 
 ## 4.0.0 (2026-08-25)
@@ -1341,6 +1361,68 @@
 
 
 ## 3.0.0 (2026-06-17)
+
+### Breaking Changes
+
+Restored by hand in #1123. The 2026-08-16 reconstruction (#1067) generated
+this file's pre-4.0 sections from commit subjects using only Features / Bug
+Fixes / Chores headings, so every `!` marker in the 1.28.0 → 3.0.0 range was
+flattened away and the hand-written migration instructions that shipped in
+the tagged changelogs were lost with them. The entries below are the sixteen
+`!`-marked commits in that range, with the operator-facing detail recovered
+from `git show v3.2.0-rc.7:CHANGELOG.md`.
+
+**Readiness vocabulary (#538, #542, #554)**
+
+- `Collection.get_index_status` `status` value renamed from `"ready"` to
+  `"queryable"`. MCP clients pattern-matching on the old value silently
+  treat the new value as unknown until updated.
+- `Collection.get_index_status` priority order flipped: a built index with a
+  captured background error from a prior attempt now reports
+  `status="queryable"` (with the diagnostic in `error`), not
+  `status="failed"`. `"failed"` now means "preconditions do not hold AND a
+  captured error exists" — the index is not readable.
+- `Collection.is_index_ready()` renamed to `Collection.is_queryable()`.
+- `Collection.wait_for_index_ready()` renamed to
+  `Collection.wait_until_queryable()`.
+- `Collection._require_index_ready()` (private) renamed to
+  `Collection._require_built()`.
+- MCP decorator `needs_index_ready` renamed to `needs_queryable`; module
+  `_server_readiness.py` renamed to `_server_queryable.py`.
+- Public exception `IndexNotReadyError` renamed to `IndexUnavailableError`,
+  which gained a required `reason` field typed by a new
+  `IndexUnavailableReason` literal.
+- Environment variable `MARKDOWN_VAULT_MCP_READY_TIMEOUT_S` renamed to
+  `MARKDOWN_VAULT_MCP_BUILD_TIMEOUT_S`. **Running deployments that set the
+  old variable silently fall back to the 60-second default after upgrading;
+  update operator configs (compose files, systemd units, `.env` files)
+  accordingly.**
+
+External consumers that imported `IndexNotReadyError`, called
+`is_index_ready()` / `wait_for_index_ready()`, or set the old env var must
+rename their references. No deprecation shims ship.
+
+**`Collection` renamed to `Vault` (#629, #630)**
+
+- The public class is `Vault`. Downstream code importing `Collection` must
+  rename its references.
+
+**Field-collapsed search results (#469, #471)**
+
+- `search` returns one entry per file with up to `chunks_per_file` nested
+  sections (`GroupedResult`) instead of one entry per chunk. Callers reading
+  a flat chunk list must iterate `sections`.
+- `get_similar` and `get_context.similar` return `GroupedResult` as well;
+  the `SimilarItem` type is removed.
+- The MCP tool surface gained `chunks_per_file` and the `GroupedResult`
+  shape, and the Vault Explorer SPA was adapted to it.
+
+**Attachment size default (#442)**
+
+- `MARKDOWN_VAULT_MCP_MAX_ATTACHMENT_SIZE_MB` default lowered from `10` to
+  `1` — a 10 MB base64-encoded attachment blows up most LLM contexts. Any
+  deployment relying on the old ceiling must set
+  `MARKDOWN_VAULT_MCP_MAX_ATTACHMENT_SIZE_MB=10` explicitly.
 
 
 ## 2.0.0-rc.5 (2026-06-17)
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -300,6 +300,8 @@ The endpoint is only available on HTTP/SSE transports. To set up:
 2. In your GitHub repository, add a webhook pointing at `https://<your-host>/github-webhook` with content type `application/json` and the same secret.
 3. Select the `push` event (other events are acknowledged with 200 and ignored).
 
+Set the secret only on a deployment that owns its remote (`GIT_REPO_URL`). Elsewhere the endpoint mounts but is inert. With no remote to pull from, every delivery is acknowledged with 200 and logged as a warning. The server logs the same warning once at startup.
+
 The periodic pull loop (`GIT_PULL_INTERVAL_S`) remains active as a belt-and-suspenders fallback for missed webhook deliveries. For the same setup in context, alongside the other sync mechanisms and what each delivery outcome means, see the [git integration guide](guides/git-integration.md#push-triggered-pull-github-webhook).
 
 ## File Watcher
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -848,6 +848,19 @@ enrichment behaviour is unchanged. A sidecar written by a provider that
 did accept empty inputs converges after one re-embed of the affected
 documents.
 
+The read side has the same hazard (#1111): `search(query, mode="semantic")`
+embeds the query verbatim, so a blank or whitespace-only query would reach
+the provider as `[""]` and return a vendor-shaped HTTP 400 rather than a
+statement about the caller's own input. `SearchManager.search()` applies
+`is_embeddable()` to the query on both the semantic and the hybrid channel
+— at the manager boundary rather than only inside `VectorIndex`, because
+hybrid reaches the provider through its own vector-channel call site — and
+returns `[]`, matching the posture of the adjacent `limit <= 0` guard.
+`VectorIndex.search()` carries the same guard as a backstop for a library
+consumer calling it directly. The guard runs *after* `_require_vectors()`,
+so a server with no embedding provider still raises
+`EmbeddingsNotConfiguredError` instead of silently answering `[]`.
+
 ### Error Handling
 
 Two-layer model:
@@ -2206,6 +2219,30 @@ unhandled exceptions are logged at `ERROR` but not propagated to the caller.
 Default: `None` (no callback). Built-in option: `GitWriteStrategy` (or the
 legacy factory `git_write_strategy(token=...)`) that auto-commits and pushes.
 
+**Per-operation staging guarantee (#894).** One MCP write operation produces
+one auto-commit containing *only* the paths that operation touched. A rename
+is the case where this is not free: it has two sides — the vanished old path
+and the new one — while the callback signature carries a single path, so the
+git strategy used to stage the old side with a pathspec-less `git add -u`.
+That stages every tracked modification in the repository, so a concurrent
+edit the server never saw (Obsidian writing straight into the vault) was
+swept into the rename's commit and pushed, and an old path git had never
+tracked had its deletion recorded not at all.
+
+The rename dispatch therefore carries the old path as well. Because
+`WriteCallback` is a public type a downstream consumer implements, the extra
+argument is an **opt-in rather than a signature change**: a callback sets the
+`accepts_old_path` attribute (`types.ACCEPTS_OLD_PATH_ATTR`) to `True`, and
+`WriteCallbackDispatcher` reads it once at construction and passes
+`old_path=` only to callbacks that advertise it. A three-argument
+third-party callback is unaffected, at type-check time and at run time.
+`GitWriteStrategy` opts in and stages with
+`git add -A -- <old_path> <new_path>`: `-A` (not `-u`) so an untracked old
+path is covered, and an explicit pathspec so nothing else in the working tree
+is. An old path git does not track is dropped from the pathspec rather than
+passed — `git add` fails the whole invocation on a pathspec matching nothing,
+which would leave the new path unstaged too.
+
 **Deprecation note**: `git_write_strategy()` factory function is preserved for
 backward compatibility. Prefer constructing `GitWriteStrategy` directly for
 access to `flush()` and `close()` methods.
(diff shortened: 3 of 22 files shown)


pvliesdonk:
## Dependency-PR sweep

Twelve dependency PRs are open. **Every one is stale against `main`** — all are based on `e9e2b11` while `main` is now `9708c3f` (the 4.0.0 release), and the ruleset's strict required checks block a behind-base branch. That is why Renovate's `automerge: true` for patch/minor has not fired: #1150 is fully green (`CI Success` ✅) and still sitting there.

`uv lock --upgrade` on current `main` confirms the "single uv.lock upgrade" hunch — it resolves openai 3.3.1, ruff 0.16.4, pymdown-extensions 11.0.2, mcp 1.29.1, and ~12 transitive bumps in one go. That refresh is in this PR (full suite green on it), which subsumes several of the rows below.

### Renovate (pep621)

| PR | Finding |
|---|---|
| **#1027** openai → 3.3.1 | Subsumed by this PR's lock refresh. |
| **#1021** ruff → 0.16.4 | Subsumed by this PR's lock refresh. |
| **#1081** numpy → v2.5.2 | **A no-op.** Its diff does not touch numpy at all — the only change is an `importlib-metadata` marker. `constraint-dependencies = ["numpy<2.5.0"]` (#727) blocks the bump, so Renovate produced a PR whose title promises something the diff cannot deliver. Close it; the real blocker is #727. |
| **#1150** pygments → `>=2,<2.22` | **Half a change.** It raises the `docs` extra's cap but leaves `[tool.uv] override-dependencies = ["pygments>=2,<2.21"]` untouched, so the resolution stays at pygments 2.20.0 either way. Separately worth a look: that override's comment says *"Pygments 2.20.0 crashes on `filename=None` in HtmlFormatter"*, but `<2.21` **admits** 2.20.0 — the pin does not exclude the version it names, and the lock is on 2.20.0 today. If 2.21.0 carries the fix, the move is to raise both pins together; if not, the cap wants to be `<2.20`. Either way this is its own issue, not a Renovate merge. |
| **#1023** mypy → v2 | Deferred by `mypy>=1.0,<2` with "#727" on the line. Hold or close pending #727. |

### Dependabot (github_actions + uv)

| PR | Finding |
|---|---|
| **#977** pymdown-extensions 10.21.3 → 11.0.1 | Already on `main` at 11.0.1; this PR's refresh takes it to 11.0.2. Superseded. |
| **#951** cryptography 49.0.0 → 50.0.0 | `uv.lock` on `main` already resolves cryptography 50.0.0. Superseded. |
| **#947** gh-action-pypi-publish 1.14.0 → 1.14.2 | Already applied — `release.yml:436` is SHA-pinned to `dc37677…  # v1.14.2`. Superseded. |
| **#895** python-semantic-release 10.5.3 → 10.6.1 | **Bumps a tool the repo no longer uses.** PSR was replaced by knope; the only remaining references are the guard steps in `release.yml` / `release-prepare.yml` that *refuse while PSR is still the release path*. There is no PSR action to bump. Close. |
| **#948** codeql-action 4 → 4.37.3 | Real choice, not a no-op: `codeql.yml` uses floating `@v4`. Pinning to an exact patch is a posture decision (and would then need Renovate/Dependabot to keep moving it). Maintainer's call. |
| **#944** setup-uv 8.2.0 → 9.0.0 | Major, and the pin lives in `.github/actions/setup-project-env` and `build-plugin-zip` — **template-owned**. It should arrive via `copier update` from `fastmcp-server-template`, not a Dependabot bump landed here that the next template render would fight. |
| **#758** mypy `<2` → `<3` | Same #727 deferral as #1023, and duplicates it across managers. Close one. |

### Two things worth their own issues

1. **The pygments override contradicts its own comment** (`<2.21` admits the 2.20.0 it says is broken). I did not change it here — this PR is scoped to the v4.1 bugs — but it is either an ineffective guard or a stale comment.
2. **Renovate and Dependabot overlap on `uv`.** `renovate.json` sets `enabledManagers: ["pep621"]`, but Dependabot also files uv-lock PRs (#977, #951), which is how the same dependency gets two PRs. Picking one owner per ecosystem would stop the duplicates.

I have not merged or closed anything — that is the maintainer's call, and the stale-base problem resolves itself once Renovate rebases after this PR lands.

---

PR #1186, opened 2026-08-28
Rescue malformed FTS5 queries and add a configurable default search mode

mikebronner:
## Summary

Two independent changes to how a search request is resolved. **Both default to exact no-ops** — an existing deployment behaves identically after this PR.

### 1. Hyphenated queries silently returned nothing

`search()` passes the caller's string straight to FTS5's `MATCH`, where it is parsed as a query *expression*. In that grammar `-`, `:`, and `*` are operators, not characters — so a natural-language query naming a hyphenated project is rejected outright:

```
sqlite> SELECT count(*) FROM notes_fts WHERE notes_fts MATCH 'memory-lint orphan';
Error: no such column: lint
```

`fts_index.py` caught that `OperationalError`, logged it at `debug`, and returned `[]`. A malformed query and a genuine no-match were indistinguishable, so any vault using kebab-case names got a silent empty result on ordinary queries. Measured on a 3,450-document vault: `memory-lint orphan`, `dev-team orchestration`, and `queue: after_commit` all returned zero; `"memory-lint" orphan` returned 48.

A rejected query is now retried **once** with each whitespace-separated token double-quoted, demoting them to string literals matched verbatim. A query that parses cleanly never reaches the retry, so deliberate FTS5 syntax keeps its meaning — verified: `alpha NOT gamma` still excludes, `snapshot*` still prefix-matches, `"kpi snapshot"` still phrase-matches. Operational faults (locked DB, disk I/O) are classified *before* the retry and propagate immediately rather than being retried.

### 2. `MARKDOWN_VAULT_MCP_DEFAULT_SEARCH_MODE`

New `SearchConfig.default_mode`, **defaulting to `keyword`** so nothing changes for existing users. `mode` on `search()` becomes optional; omitting it resolves the configured default. An operator who sets `hybrid` gets fused results on every unqualified search.

Because hybrid/semantic need embeddings, a *configured* default degrades to `keyword` when none are present — enabling it cannot break an embeddings-less vault. An **explicit** `mode="semantic"`/`"hybrid"` still raises `EmbeddingsNotConfiguredError`; a caller who asked for semantic search is never handed keyword results under a semantic label. An unknown configured value is rejected at config load.

## Test plan

- `uv run pytest` — **2,703 passed, 2 skipped**
- `ruff check --fix` → `ruff format` → `ruff format --check` — clean
- `diff-quality --violations=ruff.check --extend-select=C901,PLR0911,PLR0912,PLR0913,PLR0915,S --fail-under=100` — **100%**
- Patch coverage on added source lines — **100%** (54/54)
- `pre-commit run --all-files` — all hooks pass

Every new test was mutation-checked. Disabling the retry kills 4; making it unconditional kills the power-syntax guard; breaking the quote-escaping kills 2; dropping either propagate-guard kills 1 each; removing the mode degrade kills 1; degrading explicit modes kills 5.

## Notes for review

- The retry lives inside an `except` block, which is genuinely non-obvious control flow. It is there so that valid queries are never rewritten; `_execute_match`'s docstring and a new `docs/design/design.md` section carry the reasoning.
- The classifier `_is_malformed_query_error` preserves the original four substring checks exactly rather than widening them.
- Docs updated in the same commit: `README.md`, `docs/configuration.md`, `docs/design/design.md`, and the config wizard spec (`docs/javascripts/config-wizard/wizard-spec.json`) so the drift gates stay green.


mikebronner:
Closing this — it duplicates work already merged here.

I diagnosed the FTS5 query-rejection bug against a fork checkout that turned out to be 149 commits behind `main`, without checking upstream first. #866 already fixes it, with substantially this implementation (`_quote_fts_query` / `_is_fts5_query_error`, retry-once-quoted, classify-before-retry on both attempts) — and logs the double-failure at WARNING rather than debug, which is the better call.

Apologies for the noise. Thanks for the excellent server.

PR #1189, opened 2026-08-28
feat: ✨ Resolve the default search mode from what the vault can serve.

### LINKED ISSUE #1200: [Bug]: search defaults to keyword even where hybrid is configured and the server recommends it
### What happened?

A `search` call that omits `mode` runs keyword-only BM25, on every vault, including one with embeddings built and ready. The server's own instructions tell callers the opposite:

> Use 'search' (mode='hybrid' preferred when available) to find documents

(`_instructions.py:152`)

A recommendation in the instructions only reaches callers that read and follow it. Agent callers routinely omit `mode` and land on keyword, so an operator who configured an embedding provider, paid to build the sidecar, and keeps it current gets keyword results anyway unless every call site opts in.

### Observed

At `21f8f80`:

- `SearchManager.search` declares `mode: Literal["keyword", "semantic", "hybrid"] = "keyword"` (`managers/search.py:636`).
- The `search` tool declares the same default and documents it: "Default mode is `keyword` (FTS5/BM25)" (`_server_tools/reader.py:40,50`).
- Nothing consults whether vectors are available when the caller says nothing. `_require_vectors` (`managers/search.py:240`) is only reached once a semantic or hybrid mode has already been chosen.

The retrieval cost is not marginal. Measured by @mikebronner on a 3,476-document vault with embeddings configured, in #1189 — results returned per query:

| Query | keyword | hybrid |
|---|---|---|
| "what did we decide about the tru-data KPI snapshot performance problem" | 1 | 5 |
| "why do vault searches return nothing" | 1 | 5 |
| "how should agents handle a fork of someone else's repo" | **0** | 5 |
| "what is the rule about committing without approval" | **0** | 5 |
| `memory-lint orphan` (term query) | 5 | 5 |

Term queries are unaffected — BM25 handles those well. Conversational queries, which is what an agent actually sends, return **nothing at all** in two of four cases.

### Expected

When the caller does not specify a mode, the server picks the best mode it can actually serve: hybrid where vectors are available, keyword where they are not. Doing the work of configuring and maintaining an embedding index should not also require every call site to remember to ask for it.

### Suspected cause

`[verified: read managers/search.py:631-641, _server_tools/reader.py:40-62, and _instructions.py:152 at 21f8f80]` The default is a hardcoded literal on the signature rather than a resolution against what the vault can serve. The instructions were updated to recommend hybrid without the default following.

### Scope and decisions already taken

This is the variant of #1189 discussed [on that PR](https://github.com/pvliesdonk/markdown-vault-mcp/pull/1189#issuecomment-5460862638). #1189 adds `MARKDOWN_VAULT_MCP_DEFAULT_SEARCH_MODE` defaulting to `keyword`, so the better behaviour is opt-in. The decision recorded there is to make the resolved default hybrid-when-available instead, rather than shipping a knob that starts in the wrong position.

Carry these over from #1189, which got them right:

- **Degrade, never fail.** With no embeddings configured, the resolved default is `keyword`. Enabling this cannot make a vault unsearchable.
- **An explicit mode is never silently downgraded.** `mode="semantic"` or `mode="hybrid"` still raises `EmbeddingsNotConfiguredError` when vectors are absent. A caller who asked for semantic results is never handed keyword results under a semantic label.
- **One predicate for both paths.** The raise path and the resolve path must read the same availability check so they cannot drift.

There is no "semantic only" case to handle: FTS is always present, so availability is binary — vectors or not.

### Not breaking

Assessed against 4.0.0, per the two-part test in `AGENTS.md`:

- **No operator surface changes.** Nothing must be set to upgrade into this.
- **The MCP surface keeps its shape**, and the previous behaviour stays reachable: `mode="keyword"` is explicit, documented, and unchanged. A semantics change with the old behaviour still reachable is not breaking.

So `feat`, not `feat!` — though it is honestly a `fix`, since the defect is that the default contradicts the server's own advice.

### Open questions

- **Does a configuration knob survive at all?** Once the default resolves correctly, the case for `DEFAULT_SEARCH_MODE` is an operator who wants to force keyword globally. Worth deciding deliberately rather than inheriting it from #1189: a knob that exists only to opt *out* of the good default earns its config-surface cost differently than one that opts in.
- **The tool description and the instructions both need rewording.** `reader.py:50` states the keyword default as a fact, and `_instructions.py:152` tells callers to pass `mode='hybrid'` — advice that becomes redundant once the default does it.
- `[unverified]` Whether hybrid's latency on a large vault makes an unqualified search noticeably slower. #1189 measured result quality, not time.

### Credit

The problem statement, the benchmark, and the safety contract above are @mikebronner's work in #1189. This issue exists to carry that into the shape agreed on the PR; the contribution should be credited when it lands.

Refs #1189.

mikebronner:
Fixes #1200

## What & why

A `search` call that omits `mode` now runs the best mode the vault can serve: hybrid where embeddings are configured, keyword where they are not. Before this, it ran keyword-only BM25 on every vault, including one with a vector index built and current, while the server's own instructions told callers to prefer hybrid.

`MARKDOWN_VAULT_MCP_DEFAULT_SEARCH_MODE` (`SearchConfig.default_mode`) carries the setting and ships as `auto`. An operator who wants unqualified searches to stay off the embedding provider pins `keyword`.

### Why the default, not a recommendation

The server instructions already recommended `mode='hybrid'`. A recommendation only reaches callers that read and follow it, and agent callers routinely omit `mode`. Measured on a 3,548-document vault with embeddings configured — results returned per query:

| Query | keyword | hybrid |
|---|---|---|
| "what did we decide about the tru-data KPI snapshot performance problem" | 1 | 5 |
| "why do vault searches return nothing" | 1 | 5 |
| "how should agents handle a fork of someone else's repo" | **0** | 5 |
| "what is the rule about committing without approval" | **0** | 5 |
| `memory-lint orphan` (term query) | 5 | 5 |

Term queries are unaffected. BM25 handles those well. Conversational queries return nothing at all in two of four cases.

### What it costs

Latency, measured on the same vault with a local `fastembed` provider, warm model and warm sidecar, median of 5 runs per query: keyword 9.2ms, hybrid 47.7ms. **+38.6ms median.** Sub-50ms absolute, against an LLM turn measured in seconds.

That number does not transfer to a metered provider. Every unqualified search embeds the query, so an OpenAI or Voyage operator pays a network round trip and an API call per search. That path is unmeasured here, because no key was available. Pinning `MARKDOWN_VAULT_MCP_DEFAULT_SEARCH_MODE=keyword` is the opt-out, and it is the reason the setting survives at all.

Two smaller costs, stated so nobody has to find them:

- **Result sets change on upgrade with no config touched.** Quality improves, which is the point. But an operator who tuned prompts against BM25 ranking, or who snapshot-tests search output, gets different results without acting. Worth a release note.
- **Hybrid is not strictly better.** RRF can dilute a crisp BM25 hit by fusing in chunks that are semantically near and factually wrong. `mode="keyword"` remains the right answer for filenames and FTS5 operators, and the tool description now says so.

## Safety contract

- **Degrade, never fail.** Any configured default that needs embeddings resolves to `keyword` when the vault has none. No setting can make a vault unsearchable.
- **An explicit mode is never silently downgraded.** `mode="semantic"` and `mode="hybrid"` still raise `EmbeddingsNotConfiguredError` without vectors. A caller who asked for semantic results is never handed keyword results under a semantic label.
- **One predicate for both paths.** `_require_vectors` and `_resolve_mode` both read `_vectors_available`, so the raise path and the resolve path cannot drift.
- **An unknown configured value is rejected at config load**, not at the first search.

- **No boot-time warning when a pinned `hybrid`/`semantic` default degrades.** Today it logs at DEBUG only. An operator who pins a mode the vault cannot serve deserves louder feedback. Not filed yet, and I will open it on request.
- **Does not simplify the callers that hand-roll the fallback.** `static/prompts/research.md` and `docs/guides/para.md` still do try-hybrid-then-retry-keyword by hand. That workaround is redundant now, but rewriting shipped prompts is a separate topic. Not filed yet.
- **Does not measure the remote-provider latency or cost path.** No key available here.
- **Does not add `auto` as a caller-passable mode.** Omitting `mode` already does the same thing, so a third spelling would add surface without adding capability.

## Local review

- [x] Ran a local code-review pass on the cumulative diff before pushing.
- [x] No commit carries `!`. Assessed against 4.0.0: no operator surface must change to upgrade into this, because the new variable has a working default; and the previous behavior stays reachable through an explicit `mode="keyword"`.

## Docs impact

- [x] `README.md` — regenerated
- [x] `docs/` site pages — `docs/tools/index.md` mode table and mode-choice tip; wizard spec regenerated
- [ ] `docs/design/` — no design-doc claim changed
- [x] Inline docstrings — `search` tool description, `_resolve_mode`, and the `_instructions.py` server instructions

Config-surface artifacts (`README.md`, `server.json`, `.env.example`, `packaging/env.example`, `examples/*.env`, the wizard spec) are regenerated by `scripts/gen_config_surface.py` from the field metadata, not hand-edited.

## Test plan

- Full suite: **3,703 passed, 2 skipped.** One failure and two errors are pre-existing. `test_git.py::TestRenameCommitEndToEnd::test_vault_rename_does_not_commit_a_concurrent_edit` and the two `test_release_flow_contract.py` errors reproduce identically on a clean `upstream/main` worktree at `21f8f80`, so they do not come from this diff.
- `mypy src/ tests/` clean · `ruff check` + `ruff format` clean · `pre-commit run --all-files` all hooks pass
- `scripts/structural_gate.sh` against the merge base: **100%**, 0 violations across 277 lines
- Patch coverage on added source lines: **100%** (21/21), measured against the merge base

### Mutation checks

Every new test was mutation-checked: break one behavior in the source, confirm a named test goes red, restore.

| Mutation | Killed by |
|---|---|
| Drop the `auto`-to-`hybrid` branch in `_resolve_mode` | `test_auto_picks_hybrid_with_embeddings` |
| Revert `SearchConfig.default_mode` to `keyword` | `test_dataclass_default_is_auto_without_env` |
| Revert `ProjectConfig.default_search_mode` to `keyword` | `test_dataclass_default_is_auto_without_env` |
| Revert the `from_env` fallback to `keyword` | `test_defaults_to_auto`, `test_empty_value_falls_back_to_auto` |
| Drop `auto` from `_SEARCH_MODES` | `test_accepts_each_valid_mode[auto]`, `test_defaults_to_auto`, `test_empty_value_falls_back_to_auto` |
| Disable the no-embeddings degrade | `test_auto_picks_keyword_without_embeddings`, `test_pinned_hybrid_degrades_without_embeddings`, `test_pinned_semantic_also_degrades` |
| Make `_resolve_mode` ignore the pinned default | `test_pinned_keyword_suppresses_hybrid_with_embeddings` |
| Disable `default_mode` config validation | `test_invalid_mode_raises` |

Two of these survived the first pass, and that is why the table is here. Reverting either dataclass default left the suite green, because `from_env` always passes an explicit value, so the field defaults were never asserted on. Line coverage did not catch it either: a dataclass field executes at class-definition time and counts as covered. `ProjectConfig` and `SearchConfig` are importable library surface, so a consumer who builds one in code would have silently kept `keyword`. `test_dataclass_default_is_auto_without_env` closes that, and both mutations now die.

## Notes for review

- `mode` is `Literal[...] | None` through `ReaderFacet.search` and the `search` tool. `None` means "use the configured default". Passing an explicit mode behaves exactly as before.
- `auto` is a config value only, not a mode a caller can pass.
- The branch merged `main` twice while this was open, and the server instructions were restructured under it both times: #1188 moved them from string concatenation to a `snippets` list, then #1192 wrapped each fragment in a `Snippet(text, priority, tools)` record. Both conflicted with the wording change here. I kept the new structure each time and reapplied the wording on top, so the `search` guidance is one `Snippet` with the same `WORKFLOWS` priority and tools tuple as its neighbors.


pvliesdonk:
Thanks for the contribution. 

I agree with the problem, and would consider it a bug.

But should the fix not just be that default is hybrid when available, and either keyword/semantic when not both are available?

Why do all the work of setting up a hybrid search, then not use it? Those few times you actually need fts5 the model will likely know to change from default. 

Willing to leave that to you or let my Claude pick it up.

mikebronner:
@pvliesdonk Thanks for the feedback and the creation of the issue that had even more details. I went ahead and made the suggested updates. Below are my findings. Please let me know if there is anything I should change or address.

Thanks!
~Mike

---

Done — the default now resolves from what the vault can serve. `MARKDOWN_VAULT_MCP_DEFAULT_SEARCH_MODE` ships as `auto`: hybrid where embeddings are configured, keyword where they are not. An explicit `mode=` is unchanged, and any configured default still degrades rather than raising. I rewrote `_instructions.py`, the `search` tool description, and `docs/tools/index.md`, because all three stated the old default as a fact.

Two answers from #1200. Costs first.

## The costs

**Latency: +39ms median.** You marked this `[unverified]`, so I measured it. Same vault as the quality table, 3,548 documents indexed, local `fastembed` provider, warm model and warm sidecar, median of 5 runs per query at `limit=5`.

Results returned and time taken, from the same run, so the cost and the benefit are measured on the same queries:

| Query | results kw | results hybrid | keyword | hybrid | delta |
|---|---|---|---|---|---|
| "what did we decide about the tru-data KPI snapshot performance problem" | 1 | 5 | 8.3ms | 68.5ms | +60.2ms |
| "why do vault searches return nothing" | 1 | 5 | 10.1ms | 46.3ms | +36.2ms |
| "how should agents handle a fork of someone else's repo" | **0** | 5 | 2.5ms | 28.3ms | +25.8ms |
| "what is the rule about committing without approval" | **0** | 5 | 3.0ms | 27.6ms | +24.5ms |
| `memory-lint orphan` (term query) | 5 | 5 | 10.7ms | 59.8ms | +49.1ms |
| `holmes review learnings` (term query) | 5 | 5 | 17.1ms | 49.2ms | +32.0ms |
| **median** | | | **9.2ms** | **47.7ms** | **+38.6ms** |

`limit=5`, so 5 means the cap was reached. Read the two result columns together with the two time columns: the four conversational queries are where keyword is cheap because it is finding little or nothing, and two of them return nothing at all. The two term queries return a full page either way, and there hybrid buys ranking rather than recall.

Sub-50ms absolute, against an LLM turn measured in seconds. I consider that free.

**That is the local-provider number, and it is the wrong number for a metered provider.** Every unqualified search now embeds the query. An OpenAI or Voyage operator pays a network round trip and an API call on every search that omits `mode`. I did not measure that path and I have no key to measure it with.

**Result sets change on upgrade with no config touched.** Quality improves, which is the point. But an operator who tuned prompts against BM25 ranking, or who snapshot-tests search output, gets different results after the upgrade without acting. Not breaking under the `AGENTS.md` two-part test. Worth a release note.

**Hybrid is not strictly better.** The benchmark returns 5/5 both ways on term queries, but equal counts are not equal ranking. RRF can dilute a crisp BM25 hit by fusing in chunks that are semantically near and factually wrong. `mode="keyword"` remains the right answer for filenames and FTS5 operators. The tool description now says so.

**The degrade is quiet.** Pin `hybrid` on a vault with no embeddings and you get keyword plus one DEBUG line. A boot-time warning would serve the operator better. I left it out to keep this diff on one topic.

## Does the knob survive?

I kept it, and the cost above is the entire argument.

Pin `keyword`, and unqualified searches stay off the embedding provider. Delete the knob, and a metered-provider operator has one option left: pass `mode="keyword"` at every call site. That is the failure this issue describes, inverted.

The knob did change character. It was a way to opt *in* to the better default. It is now a way to opt *out* of a per-search bill. That is a narrower job, and I think it earns its config-surface cost, because what it avoids is money rather than preference. Your repo, your call. Removing it is a one-line change and I will make it if you disagree.

## Also

- The title still says `feat`, following your reading in #1200. If you want the `fix` your `bug` label implies, feel free to retitle it.
- `src/markdown_vault_mcp/static/prompts/research.md` (step 1) and `docs/guides/para.md` (line 257) still hand-roll try-hybrid-then-fall-back-to-keyword. That workaround is redundant now. I left both alone to keep this diff on one topic, I'm happy to make that change as well.

## Merge note

The branch merged `main` twice while this was open, and the server instructions were restructured under it both times. #1188 moved them from string concatenation to a `snippets` list. #1192 then wrapped each fragment in a `Snippet(text, priority, tools)` record for the template v6.0.0 and pvl-core 5 adoption. Both conflicted with the wording change here.

I kept the new structure each time and reapplied the wording on top. The `search` guidance is now one `Snippet` at `WORKFLOWS` priority, carrying the same `("search", "read", "list_documents", "stats")` tools tuple the fragment had before, so the allow/deny filtering treats it exactly as it did.

## Test state

3,703 pass, on the branch after the second merge. `test_git.py::TestRenameCommitEndToEnd::test_vault_rename_does_not_commit_a_concurrent_edit` and the two `test_release_flow_contract.py` errors fail identically on a clean `upstream/main` worktree at `21f8f80`, so they do not come from this diff. `mypy` and `ruff` are clean, and every `pre-commit` hook passes. `scripts/structural_gate.sh` against the merge base reports 0 violations across 277 lines. Patch coverage on added source lines is 21 of 21.

Every new test is mutation-checked. Eight mutations, each killed by a named test: dropping the auto-to-hybrid branch, reverting either dataclass default, reverting the `from_env` fallback, removing `auto` from the accepted set, dropping the no-embeddings degrade, ignoring the pinned default, and dropping config validation.

pvliesdonk:
Awesome, looks good
@claude review

PR #1196, opened 2026-08-29
fix(okf): give generated reserved files the frontmatter the index gate requires

### LINKED ISSUE #1174: [Bug]: okf_seed_log writes log.md with no frontmatter, so REQUIRED_FIELDS excludes it from the index
### What happened?

Expected the `log.md` seeded by `okf_seed_log` to be searchable in a vault that sets `MARKDOWN_VAULT_MCP_REQUIRED_FIELDS`; the generated file carries no frontmatter at all, so the indexer skips it.

### Observed

Reading the source at `134b1c1` (main):

- `build_log_markdown` (`src/markdown_vault_mcp/okf.py:700`) documents its return value as "The markdown body (no frontmatter)" and builds the file as `"# Log\n\n" + ...`.
- `OkfMigrationManager.seed_log` (`src/markdown_vault_mcp/managers/okf_migrate.py:221`) writes that body with `self._doc_mgr.write(log_path, body)` — no `frontmatter=` argument.
- A document missing a configured required field is a deterministic `missing_frontmatter` skip (`src/markdown_vault_mcp/scanner.py:1181`; the category is in `SKIP_CATEGORIES`, `src/markdown_vault_mcp/types.py:17`).

On a vault configured with, for example, `REQUIRED_FIELDS=title`, the effect is that the bundle's own change history is absent from `search` and `list_documents`. It remains reachable through `read`, which goes to disk rather than the index.

Each of the two behaviours is individually documented and deliberate. What this issue records is the pairing: the server generates a file that its own configured index gate then rejects.

### Expected

A reserved file that the server generates itself stays visible to the server's own enumeration surfaces under a configuration the server accepts at startup.

### Reproduction

Not executed. Derived by reading the files named above at `134b1c1`. I did not run a live vault with `REQUIRED_FIELDS` set and observe the skip, so the interaction is read off the code rather than seen.

### Suspected cause

- `[verified: read `okf.py:700-747` and `okf_migrate.py:221-232` at 134b1c1]` The generated body carries no frontmatter, and `write()` is called without a `frontmatter` argument.
- `[unverified]` That a vault with a required field then classifies the file as a `missing_frontmatter` skip at index time. I have not verified the cause.

### Open questions / scope

- The sibling generator `okf_generate_index` shows the same shape; filed separately as #1175.
- Not about whether `REQUIRED_FIELDS` should exclude notes at all — that behaviour is documented and intended.
- Whether reserved OKF files should be subject to the required-field gate at all is undecided here; this issue only records that today they are.

### LINKED ISSUE #1175: [Bug]: okf_generate_index writes a frontmatter-less index.md when none exists, which REQUIRED_FIELDS then excludes
### What happened?

Expected `okf_generate_index` to produce an `index.md` that is itself indexed. When the target folder has no pre-existing `index.md`, the generated file carries no frontmatter, so a vault setting `MARKDOWN_VAULT_MCP_REQUIRED_FIELDS` skips it.

### Observed

Reading the source at `134b1c1` (main):

- `build_index_markdown` (`src/markdown_vault_mcp/okf.py:700`) returns "The markdown body (no frontmatter)".
- `OkfMigrationManager.generate_index` (`src/markdown_vault_mcp/managers/okf_migrate.py:179-188`) reads any existing file first and passes `frontmatter=existing_fm if preserved else None`. Preservation is conditional on a file already being there with frontmatter; nothing supplies frontmatter when it is not.
- A document missing a configured required field is a `missing_frontmatter` skip (`src/markdown_vault_mcp/scanner.py:1181`).

There is a second-order effect on the listings themselves. `generate_index` builds its entries from `self._search_mgr.list(folder=...)` (`okf_migrate.py:155`), which reads the index, and a sub-folder pointer is synthesised only from entries that appear in that listing (`okf_migrate.py:159-162`). So when a sub-folder's only file is an `index.md` that was skipped for missing frontmatter, the parent's generated listing contains no pointer into that sub-folder — the folder becomes unreachable by navigation from the bundle root.

The frontmatter-preserving behaviour is deliberate and documented (the root `index.md`'s `okf_version` declaration must survive regeneration). This issue is about the case where there is nothing to preserve.

### Expected

Running `okf_generate_index` on a folder produces a listing that is itself indexed under a configuration the server accepts at startup, and does not remove that folder from its parent's listing.

### Reproduction

Not executed. Derived by reading the files named above at `134b1c1`. I did not run the tool against a live vault with `REQUIRED_FIELDS` set.

### Suspected cause

- `[verified: read `okf_migrate.py:123-191` and `okf.py:700-720` at 134b1c1]` The generated body carries no frontmatter, and the `frontmatter=` argument is `None` whenever no existing file supplied one.
- `[unverified]` The downstream index skip and the missing sub-folder pointer that follows from it. I have not verified the cause.

### Open questions / scope

- Related observation on the sibling generator, filed separately as #1174.
- Not about the frontmatter-preservation behaviour when a file does exist — that path is intended.

pvliesdonk:
## Closes / Refs

Closes #1174, closes #1175.

## What & why

`index.md` and `log.md` are written by the server, but `required_frontmatter` is configured by the operator. A body-only reserved file is a deterministic `missing_frontmatter` skip (`scanner.py:1240`), so on a vault that sets one the server generated files its own indexer then rejected: the bundle's navigation and change history vanished from `search` and `list_documents` while staying readable from disk.

**Both issues scope this to the one-shot migration tools; the same two call sites also run on the live write path.** With `OKF_WRITE` on, `ConventionMaintainer` creates and rewrites `log.md` (`_okf_convention.py:137`) and regenerates `index.md` (`:146`) after every content write, so the gap recurred on every save rather than once at migration. `_append_log` was worse than merely missing: it is a read-modify-write and `read()` returns the body without frontmatter, so it **stripped** frontmatter an operator had seeded by hand — undoing exactly the workaround #1177 sets out to document.

`ReservedFrontmatterPolicy` (`okf.py`) is now the single owner of what those files carry, shared by both generators and the maintainer:

- **Conditional.** With no required fields configured the policy returns `None` and the generators emit the body-only files existing bundles already have on disk — an unaffected vault sees no churn and, on a git vault, no extra commits. The 244 pre-existing OKF tests pass untouched, which is the evidence for that.
- **Existing keys win**, so a hand-authored title and the root `okf_version` declaration survive regeneration.
- **`title_field` gets the H1** the body builders emit; any other required field is seeded as `null`, since the gate tests presence rather than value and the server has nothing truthful to put there. A null value produces no `document_tags` row (`fts_index.py:737`), so this cannot pollute `list_tags` or the `status` filter.
- **`okf_version` is never synthesized** — the audit flags it outside the bundle root as `misplaced` (`okf.py:_audit_file`), which would trade one defect for another. This settles the open question raised in the #1175 triage comment: a nested folder's generated index must not carry it.

Also fixes the second-order effect in #1175: `generate_index` synthesises sub-folder pointers from index entries, so a folder whose only indexable file was its own skipped `index.md` disappeared from its parent's listing.

Verified end to end on a real vault (`OKF_WRITE=true`, `REQUIRED_FIELDS=title,type`): both reserved files land well-formed, are indexed, and `okf_validate` reports zero findings against them — no `misplaced_okf_version`, no `log_heading_shape`.

- **Exempt reserved files from the required-frontmatter gate in the scanner.** Both issues leave this open. It would change which documents the index contains for every vault, needs an `INDEX_SEMANTICS_VERSION` bump, and alters documented `REQUIRED_FIELDS` behaviour for a problem solvable inside the OKF generators. Not filed — the issues already record it as undecided.
- **Add a `frontmatter_seeded` field to `OkfIndexResult`.** `frontmatter_preserved: false` alongside a file that now has frontmatter is a wrinkle; it is answered with a precise `Attributes` docstring and a tool-reference sentence rather than a new result field.
- **Migrate reserved files already on disk.** Existing vaults keep their frontmatter-less `log.md` / `index.md` until something rewrites them. Under `OKF_WRITE` the next write into a folder fixes it; otherwise re-run the generators. Worth a line on the 4.1 release-notes page.
- **Touch #1176 / #1177 / #1178**, the other v4.1-and-later OKF issues.

## Local review

- [x] Ran a local code-review pass on the cumulative diff before opening. Report below.
- [x] No commit carries `!`. Both constructors gained a defaulted keyword argument (additive); the new `okf.py` symbols are additive; no operator surface changes — the fix needs no action to upgrade into. Assessed against 4.0.0.
- [x] No `INDEX_SEMANTICS_VERSION` bump: this changes the bytes written to reserved files, not how stored rows are derived from a note's bytes.

### Self-review 17adb92..b678261

Charters: rules, diff, comments, history, conformance. Coverage: full; past-PR charter skipped (no prior review history on these files).

No blocker or important findings survived refutation. Four candidates struck, with what settled each:

- *`type: null` reads as junk in a user's file* — satisfies the presence-only gate (`scanner.py:1240`) and produces no tag row (`fts_index.py:737`, `if value is None: continue`, confirmed by inspecting a generated vault). The alternative reintroduces the defect.
- *An omitted policy argument silently restores the bug* — the default exists so a downstream consumer constructing `OkfMigrationManager` directly is not broken; the vault wiring is covered by integration tests that go through `vault.writer`, which fail if it is dropped.
- *Extra churn on unconfigured vaults* — `build()` returns `None` there, so the written bytes are identical; the pre-existing OKF suite passes unchanged.
- *Contradicted adjacent commitments* — `build_index_markdown` / `build_log_markdown` still return "the markdown body (no frontmatter)", which stays true: frontmatter is applied at the write, not in the builder.

Every new test was confirmed to fail against a neutered policy before being kept, so all nine are genuine regression tests rather than restatements of current behaviour.

## Docs impact

- [ ] `README.md` — does not document the OKF tools or their behaviour.
- [x] `docs/` site pages — `docs/guides/okf.md` (new "Reserved files on a vault with required fields" section), `docs/tools/index.md` (both generator entries + what `frontmatter_preserved` actually reports).
- [x] `docs/design/` — `docs/design/okf.md` §7 gains "Generated reserved files and the index gate": the defect, the policy's four rules, and why the maintainer needs it for the second reason (frontmatter carried across a read-modify-write).
- [x] Inline docstrings — `ReservedFrontmatterPolicy` and `build`, `OkfIndexResult`, `OkfMigrationManager.__init__` / `generate_index` / `seed_log`, `ConventionMaintainer.__init__` / `_append_log`, plus the `okf_migrate` module docstring.

## Verification

| Gate | Result |
|---|---|
| `uv run pytest -q` | 3686 passed, 4 skipped |
| Patch coverage (`diff-cover` vs `origin/main`) | **100%** (0 of 23 lines missing) |
| `bash scripts/structural_gate.sh` | 100%, 0 violations on 419 diff lines |
| `uv run pre-commit run --all-files` | all hooks pass |

One note on the suite: `test_release_tags_are_visible_when_the_changelog_records_releases` fails on a tag-less checkout, which is what this session started with. `git fetch --tags` clears it; the run above is post-fetch. CI fetches tags, so it does not arise there.

`pyproject.toml` gains a `PLR0913` per-file ignore for `managers/okf_migrate.py`, matching the existing entries for `_okf_convention.py` and `git/query.py`: the constructor is six injected collaborators and no logic, and `PLR0913` is gate-only, so an inline `# noqa` would be stripped as unused (RUF100).

---

### DIFF
--- a/docs/design/okf.md
+++ b/docs/design/okf.md
@@ -411,6 +411,43 @@ not ongoing enforcement) but are write tools and respect read-only mode. Each is
 registered per the Tool Registration Checklist, with `destructiveHint` /
 `idempotentHint` reflecting that they mutate the vault.
 
+### Generated reserved files and the index gate
+
+The generated `index.md` / `log.md` are subject to the same
+`required_frontmatter` gate as any other note, so a body-only reserved file is
+a deterministic `missing_frontmatter` skip on a vault that configures one — the
+server generating a file its own indexer then rejects (#1174, #1175). Two
+second-order effects follow: the bundle's change history disappears from
+`search` and `list_documents` while staying readable from disk, and a folder
+whose only indexed file would have been its generated `index.md` loses its
+pointer in the parent listing, because `generate_index` synthesises sub-folder
+pointers from index entries.
+
+`ReservedFrontmatterPolicy` (`okf.py`) is the single owner of what those files
+carry. Both generators and the `OKF_WRITE` maintainer write through it:
+
+- **Conditional, not unconditional.** With no required fields configured a
+  reserved file needs no frontmatter to be indexed, and the policy returns
+  `None` so the generators keep emitting the body-only files existing bundles
+  already have on disk. Closing the defect costs an unaffected vault no churn
+  — and, on a git vault, no commits.
+- **Existing keys win.** A hand-authored title and the root `index.md`'s
+  `okf_version` declaration survive regeneration; only genuinely absent
+  required fields are added.
+- **`title_field` gets the derived title** (the same H1 the body builders
+  emit); any other required field is seeded as `null`, since the gate tests
+  presence rather than value and the server has nothing truthful to put there.
+- **`okf_version` is never synthesized.** The audit flags it on any file but
+  the bundle root as `misplaced`, so seeding it into a folder's index would
+  trade one defect for another.
+
+The `OKF_WRITE` maintainer needs this for a second reason: `_append_log` is a
+read-modify-write and `DocumentManager.read` returns the body without
+frontmatter, so the log's frontmatter has to be carried across the rewrite
+explicitly. Since the maintainer runs after *every* content write, the
+alternative is not a one-time gap but frontmatter stripped again after each
+save — including frontmatter an operator seeded by hand.
+
 ### Export
 
 Export ships **not as a bespoke tool but as an overloaded download ref** on
--- a/docs/guides/okf.md
+++ b/docs/guides/okf.md
@@ -85,6 +85,12 @@ Three tools handle the changes you should not make by hand. They are write tools
 - **`okf_generate_index`** writes a folder's `index.md` as a listing of the notes directly in that folder, plus a pointer into each subfolder's own `index.md`. It draws the description from each note's frontmatter and lists one level at a time, so a nested bundle stays navigable rather than flattening into one long page. It preserves existing frontmatter, so regenerating the root `index.md` keeps your `okf_version` declaration.
 - **`okf_seed_log`** creates a `log.md` change history from the vault's git commits, newest first. The `folder` argument both places the log and scopes its content: seeding a folder includes only the commits that touched that subtree, while seeding the bundle root includes the whole vault's history. It refuses to overwrite an existing `log.md`, so a hand-maintained history is safe.
 
+### Reserved files on a vault with required fields
+
+If you set `MARKDOWN_VAULT_MCP_REQUIRED_FIELDS`, documents missing any listed field are excluded from the index. The generated `index.md` and `log.md` are ordinary documents to the indexer, so the generators give them the fields your vault requires. Your title field is filled in from the file's heading; any other required field is written empty, for you to complete. Without this the bundle's own listing and change history would be absent from `search` and `list_documents` while still opening through `read`.
+
+Anything already in the file wins, so a title you wrote yourself and the root `index.md`'s `okf_version` declaration are left alone. On a vault that sets no required fields nothing changes: the reserved files carry no frontmatter, exactly as before.
+
 ## The enforced write layer
 
 The steps above keep a bundle conformant by convention. The enforced write layer makes the server keep two of those fields correct on its own. Turn it on by setting `MARKDOWN_VAULT_MCP_OKF_WRITE=true`. It requires `OKF_MODE` to be `auto` or `on`; pairing it with `OKF_MODE=off` is a configuration error, since there is nothing to enforce. When the flag is off the write path is untouched, so an ordinary vault behaves exactly as before.
--- a/docs/tools/index.md
+++ b/docs/tools/index.md
@@ -482,17 +482,17 @@ Rewrite `[[wikilinks]]` as the bundle-root-absolute markdown links OKF recommend
 
 #### `okf_generate_index`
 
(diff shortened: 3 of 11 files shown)


pvliesdonk:
@claude review

PR #1199, opened 2026-08-29
fix(apps): declare the font origins the SPA actually loads

### LINKED ISSUE #1181: MCP App resource declares an empty resourceDomains CSP while the SPA loads Google Fonts over HTTPS
## What happened?

The MCP Apps resource `ui://markdown_vault_mcp/app.html` is served with `_meta.ui.csp.resourceDomains: []`, but the SPA it serves loads a stylesheet from `https://fonts.googleapis.com`. An MCP host that builds its iframe policy from the declared `resourceDomains` therefore has no origin permitting that stylesheet or the font files it pulls from `https://fonts.gstatic.com`.

## Observed

`resources/read` on `ui://markdown_vault_mcp/app.html` returns `mimeType: text/html;profile=mcp-app` with:

```json
{"ui": {"csp": {"resourceDomains": []}}}
```

The returned HTML contains:

```html
<link rel="preconnect" href="https://fonts.googleapis.com"/>
<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin/>
<link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Newsreader:…&amp;family=Public+Sans:…&amp;family=IBM+Plex+Mono:…&amp;display=swap"/>
```

`_CDN_RESOURCE_DOMAINS` is `[]` in `src/markdown_vault_mcp/_vault_apps.py:27`, and `register_apps` passes it straight into `AppConfig(csp=ResourceCSP(resource_domains=_CDN_RESOURCE_DOMAINS))` (`src/markdown_vault_mcp/_server_apps.py:179`).

Version: 3.4.7 (`serverInfo.version` from the `initialize` response).

## Expected

Either the declared `resourceDomains` names every origin the served HTML fetches from, or the served HTML fetches from none.

## Reproduction

Against a server started with any source dir:

```python
# initialize, then:
# resources/read {"uri": "ui://markdown_vault_mcp/app.html"}
```

Inspect `contents[0]._meta` (empty `resourceDomains`) and grep `contents[0].text` for `fonts.googleapis.com` (a `<link rel="stylesheet">` is present). The same mismatch is visible without a server by comparing `_vault_apps.py:27` against `src/markdown_vault_mcp/static/app.html`.

## Suspected cause

[verified: read `_vault_apps.py:27` and `_server_apps.py:179`, and inspected a live `resources/read` response] The declared list is the empty `_CDN_RESOURCE_DOMAINS` constant, and the vendored SPA references Google Fonts by URL rather than embedding them. What I have **not** verified is the user-visible effect: I did not observe a host actually blocking the stylesheet, so whether this degrades to fallback fonts, or fails more visibly, is unknown. I have not verified the cause of any rendering symptom, only the declaration/content mismatch itself.

## Open questions / scope

- Which hosts enforce the declared policy strictly, and what a blocked font request looks like to the user, is unmeasured.
- [unverified]: other vendored assets in `static/app.html` may reference remote origins that the same empty declaration does not cover; I only checked the font links.
- Whether `preconnect` link tags should be present at all when the fonts are meant to be self-contained is a separate question from the CSP declaration.

pvliesdonk:
## Closes / Refs

Closes #1181. Refs #1198 (the vendoring alternative, filed while weighing this fix).

## What & why

The app-shell resource advertised `resourceDomains: []` while the served SPA loads a stylesheet from `fonts.googleapis.com` and the font files it references from `fonts.gstatic.com`. A host that builds its iframe Content Security Policy from the declaration therefore had no origin permitting either request — and a blocked font produces no error, just fallback type, so the mismatch was invisible from both sides.

The empty list was a stale truth rather than an oversight: every script dependency really is vendored inline, and the comment above the constant said so. The three typefaces are the exception the vendoring never covered, so the comment is corrected here too — it was asserting something the font links falsify.

I verified these are the only external origins the page requests, by extracting resource-loading attributes and CSS `url()` references from the served HTML: exactly the two font hosts, no CSS references, no anchors. The `github.com` and `visjs.github.io` strings elsewhere in the file are vendored-library attribution comments, not requests.

### Why declare rather than vendor

`resourceDomains` is the narrower fix and matches what the issue is actually about: the declaration disagreed with the content. Vendoring the faces instead would make the app genuinely self-contained and let the `preconnect` tags go, but it adds three base64 families to an already ~630KB file and reaches into the SPA build pipeline. That is #1198, with the two constraints I hit while weighing it: the size needs measuring before committing to the approach, and the OFL requires the license text to travel with the binaries.

### The regression test

Drift between the declaration and the content is what caused this, so the test binds the two **in both directions** and asserts set equality rather than containment:

- a newly added external asset fails until it is granted;
- a stale grant fails once its asset is vendored away — which is exactly the state #1198 would leave behind.

It matches only resource-loading attributes (`href`/`src` on `link`/`script`/`img`) plus CSS `url()`, so attribution comments don't demand phantom grants. Both new tests were confirmed to fail against the old empty list before being kept.

- **Vendor the fonts** — #1198.
- **Remove the `preconnect` tags.** They are correct while the fonts are remote; they go with the vendoring, not with this.
- **Touch #1176.** I picked it up alongside this one and stopped: the `status` collision is documented in four places, all present at `134b1c1` (the commit that issue reports reading), and the shipped PARA templates keeping `status` is a decision recorded in #1080 rather than an oversight. Evidence and a recommendation to close are [on the issue](https://github.com/pvliesdonk/markdown-vault-mcp/issues/1176#issuecomment-5462307826); re-keying the templates would reverse that decision, so it is not folded in here.

## Local review

- [x] No commit carries `!`. The MCP resource's `_meta` gains two origins — an additive change to a declaration the client re-reads on connect, with no operator or library surface touched. Assessed against 4.0.0.

## Docs impact

- [ ] `README.md` — does not document the app's CSP.
- [x] `docs/` site pages — `docs/guides/mcp-apps.md`, where the fonts are already described as the app's one runtime network dependency; it now also says the two origins are named in the resource declaration and what a strict host would otherwise do.
- [ ] `docs/design/` — `design.md` names `_CDN_RESOURCE_DOMAINS` in the module layout only, and makes no claim about its contents.
- [x] Inline docstrings — the constant's comment, which previously asserted no external domains were needed.

## Verification

| Gate | Result |
|---|---|
| `uv run pytest -q` | 3688 passed, 4 skipped (baseline 3686, plus the two new tests) |
| `bash scripts/structural_gate.sh` | 100%, 0 violations on 57 diff lines |
| `uv run pre-commit run vale --all-files` | passed |

---

### DIFF
--- a/docs/guides/mcp-apps.md
+++ b/docs/guides/mcp-apps.md
@@ -109,7 +109,10 @@ code, property keys, and tags. (As each view is restyled, these families also
 reach the remaining mono metadata such as paths and scores.) The fonts are the
 one runtime network dependency the app keeps (the vendored libraries are embedded);
 if they fail to load, the app falls back to system fonts with no loss of
-function.
+function. Because a host may build the app frame's Content Security Policy from
+the origins the resource declares, `fonts.googleapis.com` and `fonts.gstatic.com`
+are named in that declaration. A host that enforces the policy strictly would
+otherwise drop the request silently and show the fallback fonts.
 
 ### Source layout and build
 
--- a/src/markdown_vault_mcp/_vault_apps.py
+++ b/src/markdown_vault_mcp/_vault_apps.py
@@ -22,9 +22,18 @@ from markdown_vault_mcp.config import _ENV_PREFIX
 if TYPE_CHECKING:
     from markdown_vault_mcp.types import GraphView
 
-# All SPA dependencies are vendored inline (see scripts/vendor_spa.py).
-# No external CDN domains needed at runtime.
-_CDN_RESOURCE_DOMAINS: list[str] = []
+#: Origins the served SPA loads sub-resources from, declared so a host that
+#: builds its iframe CSP from ``resourceDomains`` permits them (#1181). Every
+#: script dependency is vendored inline (see scripts/vendor_spa.py), but the
+#: three typefaces are still fetched: the stylesheet from ``fonts.googleapis``
+#: and the font files it references from ``fonts.gstatic``. Keep this list and
+#: what ``static/app.html`` actually requests in step — a declaration that
+#: omits an origin fails silently, with the host simply dropping the request.
+#: ``tests/test_apps_vendoring.py`` asserts the two agree in both directions.
+_CDN_RESOURCE_DOMAINS: list[str] = [
+    "https://fonts.googleapis.com",
+    "https://fonts.gstatic.com",
+]
 
 
 def _compute_claude_app_domain() -> str | None:
--- a/tests/test_apps_vendoring.py
+++ b/tests/test_apps_vendoring.py
@@ -2,12 +2,14 @@
 
 from __future__ import annotations
 
+import re
 import shutil
 import subprocess
 import sys
 from pathlib import Path
 
 import markdown_vault_mcp._server_apps as apps
+import markdown_vault_mcp._vault_apps as vault_apps
 
 
 def test_app_html_is_served_with_tools_rewritten() -> None:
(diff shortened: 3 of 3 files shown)


pvliesdonk:
@claude review

PR #1202, opened 2026-08-29
test(apps): stop the extracted origin at a query or fragment

pvliesdonk:
## Closes / Refs

Refs #1181, refs #1199. No issue: a four-line test-only correction carried over from a review on #1199, within the typo/trivial exception in `AGENTS.md`'s PR discipline.

## What & why

The `@claude` review on #1199 raised one non-blocking observation, I fixed it, and the fix missed the merge by seconds. This carries it over.

`_spa_external_resource_origins` trims a matched URL down to its origin with `re.match(r"https?://[^/]+", url)`, stopping at the first `/`. A URL whose query or fragment follows the host directly has no `/` before it, so:

| URL | extracted "origin" before | after |
|---|---|---|
| `https://host?x=y` | `https://host?x=y` | `https://host` |
| `https://host#frag` | `https://host#frag` | `https://host` |
| `https://fonts.googleapis.com/css2?family=X` | `https://fonts.googleapis.com` | `https://fonts.googleapis.com` |

The extracted value would never equal a declared `https://host`, so the test would fail for the wrong reason — a spurious mismatch rather than a missed one, which is the worse failure mode for a guard whose whole job is to be trusted.

**Unreachable today**, as the reviewer noted: the single external URL in the SPA has a path before its query, so the extraction is byte-identical either way (third row above). The next external asset added is exactly when it would spring, which is why it seemed worth the four lines rather than leaving a trap for #1198.

- **Change any production code.** `_vault_apps.py` and the declared origins are untouched; #1199 shipped those.
- **Broaden the extraction.** It still matches only resource-loading attributes and CSS `url()`, so attribution-comment hosts stay excluded — the reviewer independently confirmed that set (`github.com`, `visjs.github.io`, `almende.com`, `naver.github.io`, `opensource.org`, `apache.org`, `w3.org`).

## Local review

- [x] No commit carries `!`. Test-only; no operator or library surface.

## Docs impact

- [ ] `README.md` — not affected.
- [ ] `docs/` site pages — not affected.
- [ ] `docs/design/` — not affected.
- [x] Inline docstrings — a three-line comment explaining why the character class excludes `?` and `#`.

## Verification

| Gate | Result |
|---|---|
| `uv run pytest tests/test_apps_vendoring.py -q` | 7 passed |
| `bash scripts/structural_gate.sh` | 100%, 0 violations on 4 diff lines |
| Behaviour on the reviewer's two shapes | `https://host?x=y` and `https://host#frag` now extract to `https://host` / `https://other` |

The merge commit is `main` merged into the branch to resolve the squash-merge divergence (the branch's pre-merge commits and `4e3b23b` carry the same tree under different history); the conflict was this one line, resolved to the hardened form.

---

### DIFF
--- a/docs/guides/mcp-apps.md
+++ b/docs/guides/mcp-apps.md
@@ -109,7 +109,10 @@ code, property keys, and tags. (As each view is restyled, these families also
 reach the remaining mono metadata such as paths and scores.) The fonts are the
 one runtime network dependency the app keeps (the vendored libraries are embedded);
 if they fail to load, the app falls back to system fonts with no loss of
-function.
+function. Because a host may build the app frame's Content Security Policy from
+the origins the resource declares, `fonts.googleapis.com` and `fonts.gstatic.com`
+are named in that declaration. A host that enforces the policy strictly would
+otherwise drop the request silently and show the fallback fonts.
 
 ### Source layout and build
 
--- a/src/markdown_vault_mcp/_vault_apps.py
+++ b/src/markdown_vault_mcp/_vault_apps.py
@@ -22,9 +22,18 @@ from markdown_vault_mcp.config import _ENV_PREFIX
 if TYPE_CHECKING:
     from markdown_vault_mcp.types import GraphView
 
-# All SPA dependencies are vendored inline (see scripts/vendor_spa.py).
-# No external CDN domains needed at runtime.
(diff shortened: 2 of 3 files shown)


pvliesdonk:
@claude review

PR #1203, opened 2026-08-29
chore(deps): update all non-major dependencies

pvliesdonk:
This PR contains the following updates:

| Package | Change | [Age](https://docs.renovatebot.com/merge-confidence/) | [Confidence](https://docs.renovatebot.com/merge-confidence/) |
|---|---|---|---|
| [openai](https://redirect.github.com/openai/openai-python) | `3.3.1` → `3.6.0` | age | confidence |
| [ruff](https://docs.astral.sh/ruff) ([source](https://redirect.github.com/astral-sh/ruff), [changelog](https://redirect.github.com/astral-sh/ruff/blob/main/CHANGELOG.md)) | `0.16.4` → `0.16.5` | age | confidence |
| [typer](https://redirect.github.com/fastapi/typer) ([changelog](https://typer.tiangolo.com/release-notes/)) | `0.27.1` → `0.27.2` | age | confidence |

---

### Release Notes

<details>
<summary>openai/openai-python (openai)</summary>

### [`v3.6.0`](https://redirect.github.com/openai/openai-python/blob/HEAD/CHANGELOG.md#360-2026-08-27)

[Compare Source](https://redirect.github.com/openai/openai-python/compare/v3.5.0...v3.6.0)

##### Features

- **api:** add compute\_units to Responses and Chat Completions usage ([#&#8203;3749](https://redirect.github.com/openai/openai-python/issues/3749)) ([52421d1](https://redirect.github.com/openai/openai-python/commit/52421d197e96b4480fcc16e78345578996452c79))

##### Bug Fixes

- **auth:** harden X.509 workload identity integration ([#&#8203;3740](https://redirect.github.com/openai/openai-python/issues/3740)) ([fc3ad6c](https://redirect.github.com/openai/openai-python/commit/fc3ad6c55a1a250e16707396e86d7373f690a0fc))

##### Chores

- **deps-dev:** bump [@&#8203;stdy/cli](https://redirect.github.com/stdy/cli) from 0.22.1 to 0.22.2 ([#&#8203;3719](https://redirect.github.com/openai/openai-python/issues/3719)) ([4f5598c](https://redirect.github.com/openai/openai-python/commit/4f5598c8d822fcc281bb8862e424fa6ead310ca8))
[...]

PR #1217, opened 2026-08-30
chore(deps): hold numpy below 2.5 in [tool.uv] constraints, the one bound Renovate cannot rewrite

### LINKED ISSUE #1216: [Bug]: Renovate's update-lockfile fallback rewrites the numpy cap to >=2.5, making the project unresolvable on the 3.11 split
### What happened?

Expected Renovate's rebase of #1203 (after the `numpy<2.5` cap from #1215 merged) to drop the numpy bump and go green; instead every syncing CI job now fails at `uv sync` in seconds.

### Observed

On #1203's rebased heads (`7f4ed18`, `e9de2c96`; runs 33307570917 and 33308033174), the install step fails:

```
× No solution found when resolving dependencies for split (markers: python_full_version == '3.11.*'):
╰─▶ Because the requested Python version (>=3.11) does not satisfy Python>=3.12 and numpy>=2.5.0 depends on Python>=3.12,
    we can conclude that numpy>=2.5.0 cannot be used.
    And because markdown-vault-mcp[all] depends on numpy>=2.5 and your project requires markdown-vault-mcp[all],
    we can conclude that your project's requirements are unsatisfiable.
```

The rebased branch's `pyproject.toml` carries `numpy>=2.5` in the extras — Renovate replaced the `>=1.20,<2.5` ranges introduced by #1215, and its commit subject changed from `chore(deps)` to `fix(deps)` accordingly (a range edit, not a lock-only edit).

### Expected

The cap should hold through Renovate re-locks: the grouped non-major PR should carry only in-range lockfile bumps (openai, ruff, typer), leaving numpy at 2.4.6.

### Reproduction

Not reproduced locally; observed on the two rebased CI runs linked above, against base `28c6b54` (which carries the cap and is green on main).

### Suspected cause

`[verified: read the failing job log, the rebased branch's resolver error naming `numpy>=2.5`, and Renovate's rangeStrategy documentation]` — `renovate.json` sets `rangeStrategy: "update-lockfile"`, which updates only the lockfile **for in-range updates** but falls back to replacing the range when the newest version is out-of-range. numpy 2.5.2 is out-of-range under the new `<2.5` cap, so Renovate rewrote the range in all three extras instead of skipping the update. The cap being on the `[project]` side (deliberate in #1215, to keep one bound per package per `scripts/check_pins.py`) is exactly what made it visible to Renovate's range rewriting.

### Open questions / scope

- `[unverified]`: whether `lockFileMaintenance` (Monday cron) would reintroduce the fork some other way once numpy is excluded from ordinary updates.
- Not about the cap itself (#1214, settled) nor the long-term floor raise (pvliesdonk/fastmcp-server-template#547).

pvliesdonk:
## Closes / Refs

Closes #1216. Refs #1203, #1214, #1215, pvliesdonk/fastmcp-server-template#547.

> Reworked after review feedback: the first draft countered Renovate's range rewriting with an `allowedVersions` rule in `renovate.json` — a second pin location, in a template-owned file with no domain seam. This version keeps exactly one bound, total.

## What & why

On rebasing #1203 over the merged extras cap (#1215), Renovate's `rangeStrategy: "update-lockfile"` hit its out-of-range fallback and **rewrote** `numpy>=1.20,<2.5` to `numpy>=2.5`, making the project unresolvable on the 3.11 split (#1216). Any bound in `[project]` is exposed to that rewriting. So:

- The extras go back to **bare `numpy`** (the `>=1.20` floor was practically vacuous — every resolver lands on 2.x).
- The single ceiling moves to `[tool.uv] constraint-dependencies = ["numpy<2.5"]` — the table Renovate's pep621 manager does not parse, so the #1216 failure mode is structurally impossible, while uv honors the constraint in every resolution: ordinary lock updates *and* the Monday `lockFileMaintenance` full re-lock. This is the arrangement that held for months before #1168 removed it.
- `renovate.json` stays pristine template content — no domain edit, no copier-update conflict.
- The pin's `# until:` points at the 3.12 floor raise (fastmcp-server-template#547), and `scripts/check_pins.py` **fails CI the moment that issue closes** — the pin mechanically cannot outlive its reason, and the one-bound-per-package rule guards against a second location reappearing.

Once this merges and Renovate re-runs, it rebuilds #1203's immortal grouped PR: the numpy bump no-ops against the constraint, leaving the in-range lock bumps (openai, ruff, typer); that PR has automerge.

- Touch `renovate.json` (first draft reverted; the template-seam gap it exposed is reported on fastmcp-server-template#258).
- Touch #1203 — Renovate rebuilds it after this merges.
- Revisit the floor plan (fastmcp-server-template#547; parked by maintainer decision).

## Local review

- [x] Reworked from first-draft review feedback (see note above); re-ran the pass on the new diff.
- [x] No `!`: dropping the vacuous `>=1.20` extras floor and adding a resolver constraint change nothing an operator or library consumer must do against the last stable release.

## Docs impact

- [ ] `README.md` — no version prose there.
- [ ] `docs/` site pages — none name numpy bounds.
- [x] `docs/design/` — packaging snippet updated to the bare-extras + `[tool.uv]` constraint shape.
- [ ] Inline docstrings — no code changed.

## Verification

| Check | Result |
|---|---|
| `uv lock` | lock carries `[manifest] constraints = numpy<2.5`; extras `requires-dist` bare |
| `uv lock --upgrade --dry-run` | openai/ruff/typer move; numpy does not appear |
| `scripts/check_pins.py --offline` | OK (1 pin, exit condition present; `until:` issue is open) |
| `uv run mypy src/ tests/` (all extras, numpy 2.4.6) | Success, 196 files |
| `uv run pytest -x -q` | 3717 passed, 18 skipped |
| `ruff check` / `format --check` | clean |
| `scripts/structural_gate.sh` | no Python source changes — skipped |

---

https://claude.ai/code/session_01DdoRbh5ZLJjZK5QVBHimvj

### DIFF
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -3182,6 +3182,7 @@ mcp = ["fastmcp>=3.0,<4"]
 embeddings-api = ["httpx>=0.25", "numpy"]
 embeddings = ["fastembed>=0.3", "numpy"]
 all = ["fastmcp>=3.0,<4", "httpx>=0.25", "fastembed>=0.3", "numpy"]
+dev = ["pytest>=7.0", "pytest-cov>=4.0", "ruff>=0.1", "mypy>=1.0"]
 
 [tool.uv]
 # The single numpy bound while requires-python floors at 3.11: numpy 2.5.x
@@ -3189,7 +3190,6 @@ all = ["fastmcp>=3.0,<4", "httpx>=0.25", "fastembed>=0.3", "numpy"]
 # (#727, #1214).  Held here, not in the extras, so Renovate cannot rewrite
 # it (#1216); check_pins.py enforces the exit condition.
 constraint-dependencies = ["numpy<2.5"]  # until: fastmcp-server-template#547
-dev = ["pytest>=7.0", "pytest-cov>=4.0", "ruff>=0.1", "mypy>=1.0"]
 
 [project.scripts]
 markdown-vault-mcp = "markdown_vault_mcp.cli:main"
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -3179,11 +3179,16 @@ dependencies = [
 
 [project.optional-dependencies]
 mcp = ["fastmcp>=3.0,<4"]
-# numpy <2.5 while requires-python floors at 3.11: numpy 2.5.x is 3.12-only
-# and its PEP 695 stubs abort mypy at python_version=3.11 (#727, #1214).
-embeddings-api = ["httpx>=0.25", "numpy>=1.20,<2.5"]
-embeddings = ["fastembed>=0.3", "numpy>=1.20,<2.5"]
-all = ["fastmcp>=3.0,<4", "httpx>=0.25", "fastembed>=0.3", "numpy>=1.20,<2.5"]
+embeddings-api = ["httpx>=0.25", "numpy"]
+embeddings = ["fastembed>=0.3", "numpy"]
+all = ["fastmcp>=3.0,<4", "httpx>=0.25", "fastembed>=0.3", "numpy"]
+
+[tool.uv]
+# The single numpy bound while requires-python floors at 3.11: numpy 2.5.x
+# is 3.12-only and its PEP 695 stubs abort mypy at python_version=3.11
+# (#727, #1214).  Held here, not in the extras, so Renovate cannot rewrite
+# it (#1216); check_pins.py enforces the exit condition.
+constraint-dependencies = ["numpy<2.5"]  # until: fastmcp-server-template#547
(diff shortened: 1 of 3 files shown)


pvliesdonk:
@claude review

pvliesdonk:
Review finding addressed in 765653e: `dev = [...]` moved back above `[tool.uv]` in the design-doc snippet, so the sample is valid TOML again. The noted pre-existing drift (snippet showing `dev` as an extra while the real file uses `[dependency-groups]`) is left as-is — it predates this PR and simplification is deliberate in that snippet.

---

PR #1221, opened 2026-08-30
refactor: consolidate the duplicated path-traversal guard into utils

### LINKED ISSUE #876: refactor: consolidate duplicated traversal-guard and frontmatter-strip helpers
**What:** two small behaviors are re-implemented in multiple modules instead of living in one shared helper.

**Where:**
- The vault traversal guard (`resolve()` + `is_relative_to()` + `"Path traversal detected"` message) now exists in ~6 places: `utils/__init__.py` (`validate_path`, `validate_history_path`), `managers/document.py` (`_validate_dir_path`), `_server_transfer.py` (×2), and `conventions.py` (`_normalize_folder`). Flagged by the #872 review pass.
- The `frontmatter.loads(text).content` with YAML-error fallback pattern exists inline in `scanner.py` (×2) and `conventions.py` (`_strip_frontmatter`).

**Why it compounds:** a hardening fix (symlink-escape handling, BOM handling like #673/#681, changed error message under test) must be applied to every copy; the copies are already drifting (`_validate_dir_path` additionally rejects root-equality).

**Suggested direction:** extract `utils.resolve_inside(path, source_dir)` and a shared `strip_frontmatter(raw) -> str`, then migrate call sites incrementally.

pvliesdonk:
## Closes / Refs

Closes #876. Refs #1161 (future-proofing epic)

## What & why

Extracts a single shared guard, `markdown_vault_mcp.utils.resolve_inside(path, base, *, original=None)`, implementing the resolve → containment-check → raise `ValueError("Path traversal detected: ...")` pattern, and migrates all nine duplicated sites to it (the duplication had grown since the issue was filed — nine sites, not six):

- `utils/__init__.py`: `validate_path`, `validate_history_path`, `validate_history_dir`
- `managers/document.py`: the attachment validator and `_validate_dir_path` — the latter keeps its additional vault-root rejection local, as that is site policy rather than containment
- `conventions.py`: `_normalize_folder`, using the helper's `original=` override so the error still names the caller's raw spelling
- `_transfer_sink.py`: `_validate_destination`, `_validate_source`, `_validate_bundle_scope`

Behaviour-preserving: every site raises the same exception class and message as before; the existing suite passes unchanged. A future hardening fix (symlink-escape handling, BOM handling, message change under test) now lands in one place.

- The frontmatter-strip half of #876 is stale and intentionally untouched: `scanner.py`'s two `frontmatter.loads(text).content` calls are bare, with skip-with-category error handling living elsewhere; only `conventions.py` has the YAML-fallback helper — there is no longer a shared pattern to extract.
- Near-miss variants stay as they are, deliberately: `okf_bundle.py` (different message, "Path escapes the vault"), `DocumentManager.read()`'s return-`None` guard, and the `is_relative_to` nesting checks in `document.py` / `_file_watcher.py` — different semantics, not the guard.

## Local review

- [x] Ran a local code-review pass on the cumulative diff before creating the PR (charters: rules, diff, comments, history; range `90f95f2..5f18f97`; no findings — verified message/class preservation at every migrated site, including the raw-spelling message in `conventions.py` and the strict-subtree policy in `_validate_dir_path`).
- [x] No commit carries `!` — behaviour-preserving internal refactor.

## Docs impact

- [ ] `README.md` — no operator-facing change
- [ ] `docs/` site pages — no tool-visible change
- [x] `docs/design/` — path-traversal-protection section now names `resolve_inside` as the single shared guard
- [x] Inline docstrings — Google-style docstring on the new helper; AGENTS.md tree line updated

New direct tests for the helper (`tests/test_utils_resolve_inside.py`) plus a test for the now-isolated root-equality branch; touched modules at 99–100% coverage, structural gate 100%.

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -13,7 +13,7 @@ The authoritative design specification lives at [`docs/design/design.md`](docs/d
 src/markdown_vault_mcp/
   __init__.py        -- minimal template-skeleton root (docstring + __version__); import from submodules, not the root (#665, #903)
   utils/
-    __init__.py        -- path-traversal guard (validate_path) + attachment/exclusion helpers re-exported for managers and facets
+    __init__.py        -- shared path-traversal guard (resolve_inside + validate_path variants) + attachment/exclusion helpers re-exported for managers and facets (#876)
     text.py            -- text normalization, position mapping, fuzzy matching
     links.py           -- link target computation and replacement
     serialization.py   -- toc_payload: TocEntry/SubtreeToc → JSON-able dicts
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -1407,6 +1407,15 @@ All public **write** methods accepting a `path` parameter call
 This applies to `write()`, `edit()`, `delete()`, `rename()`, and all
 attachment write operations.
 
+The resolve-then-contain-then-raise sequence itself lives in one shared
+helper, `utils.resolve_inside(path, base)` (#876): every raising validator —
+`validate_path`, `validate_history_path`, `validate_history_dir`, the
+attachment/folder validators in `DocumentManager`, the conventions folder
+normalizer, and the transfer-sink upload/download/bundle checks — delegates
+containment to it. Site-specific policy stays local: extension allowlists,
+`_validate_dir_path` additionally rejecting the vault root, and read paths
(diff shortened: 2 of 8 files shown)


pvliesdonk:
@claude review

PR #1224, opened 2026-08-30
feat(vault): add VaultSettings so Vault construction is settings-first

pvliesdonk:
## Closes / Refs

Closes #1158. Refs #1161 (future-proofing epic)

**Merge order note:** please merge #1160's PR (in flight) before this one — both touch `config_sections/_assembly.py`; this branch will take a base-merge afterwards.

## What & why

`Vault.__init__` took 37 keyword parameters — a config bus that `to_vault_kwargs` re-flattened from the section configs on every construction, making each new knob a five-place change and blocking a future `VaultRegistry`/`VaultClient`. This adds settings-first construction, **dual-mode and non-breaking** per the maintainer's decision:

- New `config_sections/vault_settings.py`: frozen `VaultSettings` — one field per config-derived constructor knob (31 fields), defaults deliberately mirroring Vault's *library* defaults (the `chunk_overlap_words=0` and `read_only=True` drifts vs the server tier are preserved and test-pinned); owns the construction-time derivations (`effective_indexed_fields`, `effective_exclude_patterns`, `effective_state_path`) and `from_project_config` (absorbing the historical renames and weight tuple→dict conversions).
- `_assembly.py`: `to_vault_instances(config)` → frozen `VaultInstances` (the five never-config-derived collaborators + the resolved git pull interval) and `to_vault_settings(config)`. `to_vault_kwargs` remains as a **deprecated delegating bridge** reproducing its historical dict shape exactly — `tests/test_config.py` passes byte-for-byte unchanged.
- `vault.py`: `settings: VaultSettings | None = None` added; all 37 legacy kwargs keep working (docstring-deprecated in one block, no runtime warnings — house style); mixing `settings=` with a non-default config-derived kwarg raises `ValueError` naming each conflict; the body is restructured into `_configure`/`_build_managers`/`_build_facets` with one wiring path serving both modes.
- `domain.Service.start` and the CLI construct settings-first; the two kwargs-dict "oracle" reads in domain are now typed reads on `VaultInstances`; the CLI's index-path override becomes `dataclasses.replace(settings, index_path=...)`.

A signature **drift-guard test** mechanically pins `VaultSettings` fields/defaults against `Vault.__init__` (count, names, defaults, and the two deliberate drifts), and an **equivalence test** builds one vault per mode and compares the resolved wiring — the dual mode cannot silently diverge.

## What this PR deliberately does NOT do

[...]

PR #1226, opened 2026-08-30
fix(git): resolve commit identity at the tool edge so OIDC claims survive the dispatcher thread

### LINKED ISSUE #1160: [Decay]: write-path identity is read from the MCP context in two places with two different rules (no Principal)
## What
The vault's write path reads "who is acting" directly from the MCP request context in two places, binding write semantics to FastMCP even though the core library is otherwise import-clean of it. [verified: grep for `fastmcp` imports under `src/` at 4f71ecc9]

## Where
- `src/markdown_vault_mcp/git/strategy.py` — `_extract_claim` / `_check_identity` read OIDC claims via `fastmcp.server.dependencies.get_access_token` for committer name/email.
- `src/markdown_vault_mcp/_okf_write.py` — `resolve_write_actor` reads `fastmcp_pvl_core.get_subject` for OKF provenance stamping.
- Credentials (`GIT_TOKEN`/`GIT_USERNAME`) and permissions (`read_only`) are resolved from config in yet other places; there is no single value carrying subject + credentials + committer identity.

The refactor-audit does not surface this — it is semantic coupling with no complexity signal (the skill's documented blind spot). It is the "identity is scattered" finding in the decoupling study (branch `docs/architectural-studies`, `docs/design/studies/decoupling-and-layering.md` §1.1 / §2.1) and the "identity propagation" hard problem in the service study (§3.4).

## Why it compounds
Any non-MCP driver of the vault (CLI, a service, tests exercising the write path) has no way to supply an actor except by faking the MCP context, and the two reads resolve identity by different rules (token claims vs. subject string) that can disagree. #893's `GitWriteStrategy` split has to decide where these reads go; doing it without a single identity value re-scatters them across the new collaborators. [unverified]: whether `get_access_token` is reachable outside a request (tests appear to patch it). I have not verified the cause.

## Suggested direction
A starting point, not a prescribed refactor: a `Principal` value (subject, credentials, committer name/email) resolved once by an `IdentitySource` at the MCP adapter edge and passed into the write path, with the two context reads becoming the MCP implementation of that source. Investigation first: enumerate every identity read and confirm the two current rules agree.

---

### LINKED ISSUE #1218: [Bug]: GIT_COMMIT_NAME_CLAIM / GIT_COMMIT_EMAIL_CLAIM never take effect on writes through the write-callback dispatcher
### What happened?

Expected: with `MARKDOWN_VAULT_MCP_GIT_COMMIT_NAME_CLAIM` / `GIT_COMMIT_EMAIL_CLAIM` configured (README.md:307, feature from #565), an authenticated write is committed with the OIDC claim values as git author. Actual: the commit author is always the static `commit_name`/`commit_email` fallback; the claims are never applied on the deployed write path.

### Observed

Found during the #1160 identity-read investigation (full write-up: https://github.com/pvliesdonk/markdown-vault-mcp/issues/1160#issuecomment-5468719544), at commit `3ea29e0` (v4.1):

- `GitWriteStrategy.__call__` resolves the claims per commit via `_extract_claim` (`src/markdown_vault_mcp/git/strategy.py:400,403`), which calls `fastmcp.server.dependencies.get_access_token` (`strategy.py:23`, `:58-74`).
- The strategy is not invoked in the request: `vault.py:490/523` route every write callback through `WriteCallbackDispatcher.fire`, which enqueues it and runs it on the dedicated `write-callback` daemon thread (`write_callback.py:129-155`).
- On that thread there is no request context: FastMCP's `get_access_token` catches the no-request case and returns `None` (`fastmcp/server/dependencies.py:467-506` in the pinned version), so `_extract_claim` returns `None` and both claims silently fall back to the static identity — on every write, for every authenticated user.
- The same write's OKF provenance IS stamped with the human subject, because that read happens in the tool handler and survives `asyncio.to_thread` via the `OkfWriteIntent` contextvar (`_okf_write.py:62-78`) — so the two identity records for one write disagree in production.

The existing tests don't catch it: `tests/test_git.py:5472-5557` exercise the claims by calling `strategy(f, ..., "write")` synchronously with `markdown_vault_mcp.git.strategy._get_access_token` patched; no test drives an authenticated write through `WriteCallbackDispatcher`.

### Expected

A write performed by an authenticated user whose token carries the configured claims produces a commit whose author name/email are the claim values, as #565 documented and README still documents.

### Reproduction

Established by static analysis at `3ea29e0`, not by a live authenticated deployment run:

1. Traced the write path: `_server_tools/writer.py` → `DocumentManager` → `WriteCallbackDispatcher.fire` (`vault.py:490/523`) → dispatcher worker thread (`write_callback.py:129-155`) → `GitWriteStrategy.__call__` → `_extract_claim`.
2. Read FastMCP's `get_access_token` in the installed pinned version: outside a request scope it returns `None` rather than raising.
3. Confirmed a fresh `threading.Thread` does not inherit the request's contextvars on CPython 3.13 (unlike `asyncio.to_thread`, which copies the context — the reason the OKF-side read works).
4. Confirmed no test exercises the claims through the dispatcher (grep of `tests/test_git.py`).

A live reproduction (OIDC-authenticated server, write via MCP, inspect `git log --format='%an %ae'`) has not been run.

### Suspected cause

`[verified: static read of the wiring above]` — the claims are read on the dispatcher thread, where the request context that carries the access token does not exist, so the fallback branch is always taken. The failure is silent because `get_access_token` returns `None` instead of raising there. I have not reproduced it in a live deployment.

### Open questions / scope

- `[unverified]`: whether any deployment configuration exists where the strategy is still invoked synchronously in-request (which would make the claims work there); none was found in `src/`.
- A regression test presumably has to drive the write through `WriteCallbackDispatcher` rather than calling the strategy directly — the synchronous-call tests pass today.
- Not this issue's scope: the broader identity unification (`Principal`/`IdentitySource`, #1160) and the semantic sub-vs-name/email claim disagreement described there. #1160's wave-2 work would subsume this fix; this issue exists so the defect can also be fixed standalone in 4.2 if that lands first.
- Related history: #565 introduced the claims; #599 introduced the deferred dispatcher. `[unverified]`: whether the claims ever worked through the deployed path or were broken from the moment the two features were combined.

pvliesdonk:
## Closes / Refs

Closes #1160, closes #1218. Refs #1161 (future-proofing epic)

**Merge order note:** please merge this before #1224 — both touch `config_sections/_assembly.py`; #1224 will take a base-merge afterwards.

## What & why

**Problem.** `GIT_COMMIT_NAME_CLAIM`/`GIT_COMMIT_EMAIL_CLAIM` were inert on the deployed write path: `GitWriteStrategy` read the OIDC claims via FastMCP's `get_access_token()` per commit, but the strategy runs on `WriteCallbackDispatcher`'s daemon thread, where no request context exists — the read silently returned `None` and every commit fell back to the static author, while the same write's OKF provenance (contextvar through `asyncio.to_thread`) recorded the human subject (#1218). More broadly, write identity was read from the MCP context in two places with two different rules (#1160).

**Fix.** Identity is now resolved **once, at the MCP tool edge**, into a frozen `Principal` (new private `_identity.py`; credentials and permissions deliberately excluded). The write tools bind it on a contextvar around their `to_thread` call via `write_identity_scope()` — which also derives the OKF write-intent actor from the *same* Principal, removing the rule disagreement by construction. `WriteCallbackDispatcher.fire()` snapshots the bound Principal into the queue item (capture-at-fire — the only point still inside the copied request context) and forwards it only to callbacks opting in via `ACCEPTS_PRINCIPAL_ATTR` (mirroring the #894 `old_path` opt-in). `GitWriteStrategy` consumes `principal.display_name/email` as the commit `--author` with the same per-field static fallback; the `git/` package no longer imports `fastmcp` (guard-tested), keeping it usable from non-MCP drivers.

**Compatibility.** `tests/test_okf_write.py` passes **byte-for-byte unmodified** — the `_okf_write` resolvers prefer a bound Principal and otherwise keep the existing lazy `get_subject` path. Third-party three-argument write callbacks are untouched. The strategy's `commit_name_claim`/`commit_email_claim` ctor kwargs remain accepted (docstring-deprecated) and still feed the startup identity warning. Delete/rename/move_folder/attachment writes — previously identity-less — now carry the Principal for git attribution while their OKF enricher semantics stay byte-identical (principal-only binding).

- No `AccessPolicy`/permissions model and no credentials on the Principal — the service study's later concerns: #1161 out-of-scope list.
- No removal of the deprecated strategy ctor claim kwargs — they still inform `_check_identity`'s startup warning; removal rides the next major (cf. #1225's pattern).
- No change to conflict-commit identity (`git/conflict.py` background commits keep the documented static server identity — per-user attribution genuinely doesn't apply there).

## Local review

- [x] Ran a local code-review pass on the cumulative diff before creating the PR (range `923e8b1..d6e6779`; verified the per-field author fallback preserves old semantics, the 2×2 dispatcher dispatch reproduces each prior case, the previously-unbound write ops keep OKF behaviour byte-identical, and `test_okf_write.py` has a zero diff).
- [x] No commit carries `!` — the fix makes a documented feature work; public callback surface is additive (opt-in attribute).

## Docs impact

- [x] `README.md` — claim-var rows note resolution at tool-call time
- [x] `docs/` site pages — `docs/configuration.md` updated
- [x] `docs/design/` — new "Write identity: the Principal value (#1160, fixes #1218)" section
- [x] Inline docstrings — Google-style throughout `_identity.py`; deprecation notes on the strategy ctor kwargs; AGENTS.md tree gains `_identity.py`

Tests: new #1218 regression drives a write through the **real dispatcher worker thread** into a real git repo (author=claims, committer=static — the false-green the old synchronous tests had is documented in its docstring); principal-forwarding trio + fire-time-snapshot test; claim-extraction edge cases re-pinned on `resolve_mcp_principal`; fastmcp-import-clean guard on `git/`. Gates: pytest 3748 passed; ruff/mypy clean; `_identity.py`/`write_callback.py`/`writer.py` at 100% coverage; structural gate 100% (945 lines, 0 violations).

---

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -203,9 +203,9 @@
 # MARKDOWN_VAULT_MCP_GIT_COMMIT_NAME=markdown-vault-mcp
 # Git committer email for auto-commits.
 # MARKDOWN_VAULT_MCP_GIT_COMMIT_EMAIL=noreply@markdown-vault-mcp
-# OIDC claim key used as the commit author name (e.g. name); overrides GIT_COMMIT_NAME per request when an OIDC token is present.
+# OIDC claim key used as the commit author name (e.g. name); overrides GIT_COMMIT_NAME per request when an OIDC token is present. The claim is resolved when the tool call arrives and carried to the background commit, so it applies on every write.
 # MARKDOWN_VAULT_MCP_GIT_COMMIT_NAME_CLAIM=
-# OIDC claim key used as the commit author email (e.g. email); overrides GIT_COMMIT_EMAIL per request when an OIDC token is present.
+# OIDC claim key used as the commit author email (e.g. email); overrides GIT_COMMIT_EMAIL per request when an OIDC token is present. Resolved and carried the same way as the name claim.
 # MARKDOWN_VAULT_MCP_GIT_COMMIT_EMAIL_CLAIM=
 # Run git lfs pull on startup to fetch LFS-tracked attachments; set to false for repos without LFS.
 # MARKDOWN_VAULT_MCP_GIT_LFS=true
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -67,6 +67,7 @@ src/markdown_vault_mcp/
   okf_bundle.py        -- OKF bundle-zip export from live vault state, served via an okf-bundle download ref (#963)
   _okf_convention.py   -- OKF reserved-file maintenance after enforced writes: log.md bullet + index.md refresh (#964)
   _okf_write.py        -- OKF enforced-write runtime: contextvar actor + provenance stamp / verified clear (#964)
+  _identity.py         -- Principal write identity: tool-edge resolution, contextvar carry, claim-key registration (#1160)
   summarizer.py        -- Summarizer ABC + OpenAI-compatible chat-completions backend (#915)
   vault.py             -- thin composition root: lifecycle, wiring, facet accessors (index-write → indexing/coordinator.py)
   write_callback.py    -- WriteCallbackDispatcher: deferred git-commit callback worker (#599)
--- a/README.md
+++ b/README.md
@@ -304,8 +304,8 @@ Domain environment variables use the `MARKDOWN_VAULT_MCP_` prefix:
 | `MARKDOWN_VAULT_MCP_GIT_PUSH_DELAY_S` | `30.0` | No | Seconds of write-idle time before pushing; 0 pushes only on shutdown. |
 | `MARKDOWN_VAULT_MCP_GIT_COMMIT_NAME` | `markdown-vault-mcp` | No | Git committer name for auto-commits; set this in Docker where git config user.name is empty. |
 | `MARKDOWN_VAULT_MCP_GIT_COMMIT_EMAIL` | `noreply@markdown-vault-mcp` | No | Git committer email for auto-commits. |
-| `MARKDOWN_VAULT_MCP_GIT_COMMIT_NAME_CLAIM` | (none) | No | OIDC claim key used as the commit author name (such as name); overrides GIT_COMMIT_NAME per request when an OIDC token is present. |
-| `MARKDOWN_VAULT_MCP_GIT_COMMIT_EMAIL_CLAIM` | (none) | No | OIDC claim key used as the commit author email (such as email); overrides GIT_COMMIT_EMAIL per request when an OIDC token is present. |
+| `MARKDOWN_VAULT_MCP_GIT_COMMIT_NAME_CLAIM` | (none) | No | OIDC claim key used as the commit author name (such as name); overrides GIT_COMMIT_NAME per request when an OIDC token is present. The claim is resolved when the tool call arrives and carried to the background commit, so it applies on every write. |
+| `MARKDOWN_VAULT_MCP_GIT_COMMIT_EMAIL_CLAIM` | (none) | No | OIDC claim key used as the commit author email (such as email); overrides GIT_COMMIT_EMAIL per request when an OIDC token is present. Resolved and carried the same way as the name claim. |
 | `MARKDOWN_VAULT_MCP_GIT_LFS` | `true` | No | Run git lfs pull on startup to fetch LFS-tracked attachments; set to false for repos without LFS. |
 | `MARKDOWN_VAULT_MCP_FILE_WATCHER` | `true` | No | Watch the vault for external filesystem changes; auto-disabled when git pull or the webhook is active. Requires the file-watcher extra. |
 | `MARKDOWN_VAULT_MCP_FILE_WATCHER_DEBOUNCE_S` | `2.0` | No | Seconds of quiet after the last filesystem event before reindexing. |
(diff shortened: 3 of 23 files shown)


pvliesdonk:
@claude review

pvliesdonk:
On the review's minor nit (delete the now-unreferenced `RenameAwareWriteCallback`): verified, but deliberately **not** removing it in this PR. The Protocol predates this diff (shipped with the #894 rename opt-in in the last stable release) and lives in `markdown_vault_mcp.types`, one of the declared public-API modules — it is the documented structural type for a third-party callback implementing the `ACCEPTS_OLD_PATH_ATTR` opt-in, independent of whether the dispatcher's internal cast still names it. Removing a released public symbol is a library-interface break under the breaking-change policy, so it belongs with the next major's cleanup (alongside #1225), not in this `fix:`. The new `PrincipalAwareWriteCallback` docstring keeps the cross-reference so the pairing stays discoverable.

---

PR #1227, opened 2026-08-30
feat(index): record skip tombstones so FTS absence means deleted

### LINKED ISSUE #1129: [Decay]: FTS row absence conflates deleted, excluded and unparseable, so each consumer re-derives the distinction
## What

Absence of a path from the FTS `documents` table encodes two unrelated facts — "the file is gone or deliberately excluded" and "the file exists but this build could not parse it" — so any consumer that needs to tell them apart has to re-derive the distinction from the filesystem, and each one does it differently or not at all.

## Where

`src/markdown_vault_mcp/fts_index.py` (no row is written for a skipped file), with the two consumers in `src/markdown_vault_mcp/managers/index.py`:

- `_converge_embeddings` computes `stale_paths = [p for p in vec_by_path if p not in fts_by_path]` and drops those vectors unconditionally. `[verified: read at managers/index.py:1172 on main @ b0caf28]` Its own docstring enumerates the intended causes as "deleted or newly excluded while no server ran" — the parse-failure case reaches the same branch and is not named.
- PR #1115 needed the opposite answer at the sibling call site and recovered it by re-walking the vault and re-parsing every candidate absent from FTS (`_empty_embedding_build_is_authoritative`), threading a new `on_error` callback through `utils/fs.iter_markdown_files` to notice an incomplete walk. `[verified: read the PR diff at fe4dfba]`

A third representation of the same fact already exists: `ChangeTracker.skip_reasons()` persists `{path: {category, detail}}` for deterministic skips, maintained in lockstep with the skipped set and surfaced through `get_index_status`. `[verified: read tracker.py:248-333 and managers/index.py skipped_files]`

## Why it compounds

Every future consumer that cares whether a path is *really* gone faces the same choice: re-derive it from disk (cost, plus a different answer than the build that produced the FTS state), consult the tracker's parallel registry, or ignore the distinction. The three answers already coexist in one file, and nothing makes a new call site pick the same one — the ambiguity is re-litigated per call site rather than resolved once.

The immediate cost is that a correctness fix has to be written twice and can still be undone one code path later: #1115's guard protects the empty-save decision while the convergence pass, which runs on every boot, makes the same decision unguarded.

`[unverified]`: the unguarded convergence drop may be a live data-loss defect in its own right (a document that fails to parse once loses its vectors on the next boot and is not re-embedded until it parses again). I have not reproduced it. It belongs in its own issue rather than this one.

## Suggested direction

A starting point, not a prescribed refactor: make skipped files *present* in FTS — a `documents` row carrying the skip category and no chunks, excluded from search — so that absence means exactly one thing and `p not in fts_paths` becomes correct without a side registry. The existing additive-migration pattern (`PRAGMA table_info` + `ALTER TABLE ... ADD COLUMN`, `fts_index.py:378-428`) and the derived-content invalidation from #1124 (bump `INDEX_SEMANTICS_VERSION`, every deployed index rebuilds itself on next start) between them cover the rollout.

Two constraints worth recording, because they are what keep this a non-breaking change under `CLAUDE.md`'s policy and therefore releasable in a minor:

- Migration stays automatic. The FTS file is derived state rebuilt from the vault, so no operator action means no operator-surface break.
- Tombstones stay invisible to existing readers. `list_notes`, search, `list_documents` and `stats` must keep their current semantics; a consumer that wants tombstones asks for them explicitly. Changing what an existing public method returns would be a library-interface break and would move this out of a minor.

pvliesdonk:
## Closes / Refs

Closes #1129. Refs #1161 (future-proofing epic), refs #1115, refs #1130

## What & why

Skipped files (the four `SKIP_CATEGORIES`: `parse_error`, `encoding_error`, `missing_frontmatter`, `internal_error`) are now **present** in the FTS index as tombstone rows — `skip_category` non-NULL, `chunk_count = 0`, no child rows — so absence of a `documents` row means exactly one thing: *not an index candidate* (deleted, non-markdown, or excluded by pattern). All direct readers query the new `documents_live` view, making tombstones invisible to every existing consumer; child-table joins (search MATCH, chunks, TOC, backlinks, field values) are naturally tombstone-safe and deliberately unchanged.

**Behaviour fix:** a previously indexed note that becomes unparseable no longer serves its last-good content — `reindex` and `process_dirty_paths` replace the stale row with a tombstone, so the note drops out of search/read while staying listed in `get_index_status.skipped_files` (the tracker registry remains authoritative and is kept in lockstep, including tombstone GC when a skipped file leaves the registry). Previously the three pipelines handled the same failure three different ways (build: no row; reindex: stale row kept; watcher: mixed) — they now converge on one shared choke point.

The same fix reaches the vector sidecar: convergence and the empty-build authoritative check treat a tombstoned absence as deliberate, replacing the #1115/#1130 per-candidate re-parse/walk machinery with a tombstone lookup plus one transient-aware stat (net −35 lines in `managers/embeddings.py`; the vanished-root and transient-failure protections the characterization suites pin are preserved).

**Broken links, by decision:** links to tombstoned files stay broken, `get_outlinks` reports `exists=False` for them, and wikilinks never resolve to a tombstone — a skipped file contributes no readable content, so pretending its target exists would hide exactly the problem the tombstone records.

**Migration story:** automatic and non-breaking (minor-releasable). The columns are additive (`ALTER TABLE` on open, `chunk_count` precedent) and `INDEX_SEMANTICS_VERSION` is bumped 2 → 3, so the first start after upgrade cold-rebuilds once and materialises tombstones for existing skips; no operator action, no re-embedding (chunking provenance unchanged). The warm-restart gate now keys on `has_documents()` (tombstones included) so an all-tombstone vault short-circuits like any completed build.

- Excluded-by-pattern files are NOT tombstoned — exclusion is config-static and derivable without filesystem state; tombstoning it would couple index contents to `exclude_patterns`.
- `get_index_status` keeps reading the tracker registry; deriving that surface from tombstones is a possible follow-up, kept out to keep this diff reviewable.
- Walk-incompleteness stays guarded by the #1115 `on_walk_error` mechanism — an unreadable directory yields no candidate and hence no tombstone, so it cannot be represented by one.

## Local review

- [x] Ran a local code-review pass on the cumulative diff before creating the PR (range `923e8b1..4fef21f`; verified the migration follows the additive precedent, `upsert_note`'s delete-first clears a prior tombstone for live↔tombstone transitions, the warm-restart gate change preserves the empty-index rebuild, the lockstep GC respects the transient-OSError policy, and the four updated characterization tests encode exactly the approved semantics change).
- [x] No commit carries `!` — additive schema, automatic migration, tombstones invisible to all existing readers; `feat` per the changelog rules (the semantics-version bump and stale-row fix are user-visible).

**Reviewer flags:** a per-file `S608` ignore for `fts_index.py` was added inside the PROJECT-RUFF-IGNORES sentinels (code-owned SQL fragments, all values parameter-bound — same gate-only rationale as the `git/` S603/S607 entry); `get_outlinks`' existence join was added to the reader list for consistency with the broken-links decision.

## Docs impact

- [ ] `README.md` — no operator-facing configuration change
- [x] `docs/` site pages — `docs/tools/index.md` sentence on unparseable notes dropping from search/read
- [x] `docs/design/` — new "Skip Tombstones (#1129)" section + DDL updated
- [x] Inline docstrings — module docstring documents the absence semantics; the normative `INDEX_SEMANTICS_VERSION` docstring records version 3

Gates: pytest 3764 passed; ruff/mypy clean; diff coverage 92%+ (remaining miss is the duplicate-column race branch, same as the untested `chunk_count` precedent); structural gate 100%. New `tests/test_skip_tombstones.py` (36 tests) incl. a parametrized 17-consumer reader-exclusion matrix and the reindex stale-content regression.

---

### DIFF
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -586,6 +586,67 @@ lifespan behind the initial build job, #665) + explicit `reindex` tool call
 Architecture supports adding `watch_interval` or watchdog integration later
 without refactoring.
 
+### Skip Tombstones (#1129)
+
+**Absence semantics.** Absence of a `documents` row means exactly one thing:
+the path is *not an index candidate* — deleted, non-markdown, or excluded by
+pattern. Every surfaced deterministic skip (the four `SKIP_CATEGORIES`:
+`parse_error`, `encoding_error`, `missing_frontmatter`, `internal_error`)
+instead stays PRESENT as a *tombstone* row: `skip_category` non-NULL,
+`chunk_count = 0`, no child rows. Excluded-by-pattern files stay absent —
+exclusion is config-static, so their absence is already unambiguous — and so
+do transient `OSError` skips (nothing is recorded; the file retries next
+scan). Walk incompleteness is not representable as tombstones, so the
+embeddings empty-build check keeps its `on_walk_error` source walk.
+
+**Invariant.** A tombstone row exists iff the path is currently in the
+tracker's `skip_reasons` registry, which stays authoritative for
+`get_index_status.skipped_files`; the two are kept in lockstep by the three
+skip-recording pipelines (`build_index`, `reindex`/`_parse_changed_notes`,
+`process_dirty_paths`) through the shared `IndexManager._tombstone_skip`
+helper, and by `IndexManager._sync_tombstones`, which garbage-collects
+tombstones whose path left the registry (a skipped file deleted from disk
+leaves the registry silently — it was never indexed — so its tombstone is
+dropped after each `update_state`).
+
+**Readers.** All direct-`documents` readers query the `documents_live` view,
+so tombstones are invisible to every existing consumer — non-breaking by
+construction. Queries that reach `documents` through child tables (the FTS5
+MATCH join, `list_chunks`, `get_toc`, `get_backlinks`, `list_field_values`)
+are naturally safe because tombstones have no child rows.
+
+**Stale-row fix.** Before #1129, a file that *became* unparseable kept its
+last-good FTS row on `reindex` (and on `process_dirty_paths`), serving stale
+content from search/read indefinitely. Tombstoning replaces that row, so the
+stale content stops being served while the path remains distinguishable from
+a deletion.
+
+**Links to tombstones stay broken — by decision.** `get_broken_links` /
+`count_broken_links` / `get_outlinks.exists` check `documents_live`, and
+vault-wide wikilink resolution never resolves to a tombstone. A skipped file
+contributes no readable content, so pretending its link targets exist would
+hide exactly the problem the tombstone records.
+
+**Warm restart.** The coordinator's warm-restart gate keys on
+`FTSIndex.has_documents()` (ANY row, tombstones included), so an
+all-tombstone vault short-circuits like any completed build; reported stats
+stay on the live `count_documents()`.
+
+**Embeddings.** Tombstones give the vector convergence a direct answer to
+"is this FTS absence deliberate?" (#1130): a stale sidecar path is
+reclaimable iff it has a tombstone (any category), is excluded, or its
+source is gone from disk (one `stat`, transient-`OSError`-aware, with an
+unavailable source *root* confirming nothing); and an empty embedding build
+is authoritative iff every candidate absent from FTS has a tombstone. The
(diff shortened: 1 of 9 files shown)


pvliesdonk:
@claude review

PR #1237, opened 2026-08-30
refactor: decouple versioning, index, and identity behind protocol seams

### LINKED ISSUE #1229: [Decay]: every git consumer type-hints the concrete GitWriteStrategy, so there is no seam to substitute a versioning backend
## What

The git package has been decomposed internally (#893 extracted `RepoBootstrap` and `PushScheduler`, #1219 decomposed `get_file_history`), but every consumer still names the concrete `GitWriteStrategy` in its type hints. No `Protocol` or ABC exists for versioning, history, or syncing anywhere in `src/`. [verified: grep for `Protocol` under `src/` at 8e20b25 — the only Protocols are `RenameAwareWriteCallback` and `PrincipalAwareWriteCallback` in `types.py`, plus the row Protocols in `managers/_ranking.py`]

## Where

- `src/markdown_vault_mcp/vault.py:337` — `git_strategy: GitWriteStrategy | None`.
- `src/markdown_vault_mcp/managers/git_query.py:47` — `git_strategy: GitWriteStrategy | None`, but the class calls exactly two methods on it (`get_file_history`, `get_file_diff`).
- `src/markdown_vault_mcp/_server_tools/git.py:43` — a runtime `isinstance(strategy, GitWriteStrategy)` gate plus reads of the private `strategy._managed`; the same module also calls `strategy._git` (line 61), `strategy._resolve_force_repo()` (433, 460) and `strategy._head_sha()` (437, 470). Four private members of the strategy are load-bearing in the MCP tool layer.
- `src/markdown_vault_mcp/config_sections/_assembly.py:336-342` — the same object is wired as both `git_strategy` and `on_write`.

By contrast `_github_webhook.py` needs only pull and already goes through `vault.force_pull` without naming the class at all. [verified: read at 8e20b25]

This is the "git strategy is a god class" finding in the decoupling study (branch `docs/architectural-studies`, `docs/design/studies/decoupling-and-layering.md` §1.2 / §2.2), whose §2.2 resolution notes the consumers already want different subsets of the surface.

## Why it compounds

The consumers want three different things — history only, pull only, commit plus sync — but each is typed against the union, so a change to any part of the strategy's surface is a change every consumer is nominally coupled to. The private-member reaches in `_server_tools/git.py` are the sharper cost: they make the MCP tool layer depend on the strategy's internals, so any further decomposition of `strategy.py` has to preserve four private names or update the tool layer in lockstep. `RemoteStore`, which the service-separation study's separated mode would need, has nowhere to plug in. [unverified]: whether a non-git versioning backend is wanted at all — no consumer has asked for one. I have not verified the cause.

## Suggested direction

A starting point, not a prescribed refactor: narrow interfaces named for the three concerns the study identifies (history, syncing, versioning), with `GitWriteStrategy` remaining the single object that implements all of them — the study is explicit that splitting into three *objects* is not the goal, and `RepoBootstrap`/`PushScheduler` sharing one `threading.Lock` plus `Vault.close()`'s `on_write is git_strategy` identity check are concrete reasons not to. Whatever private members the tool layer needs would have to become public first.

### LINKED ISSUE #1230: [Decay]: managers depend on the concrete FTSIndex/VectorIndex, so the documented Qdrant escape hatch has nothing to plug into
## What

Every manager type-hints the concrete `FTSIndex` and `VectorIndex`. No `Protocol` or ABC exists for the keyword index, the link graph, or the vector store — even though the embedding *provider* feeding the vector store has been an ABC since early on (`providers.py:148`), and `ChunkStrategy` is a Protocol (`scanner.py:31`). The storage the provider feeds never received the same move. [verified: grep for `Protocol`/`ABC` under `src/` at 8e20b25]

## Where

- `src/markdown_vault_mcp/managers/search.py:179`, `link.py:43`, `document.py:183`, `index.py:104`, `embeddings.py:98`, `indexing/coordinator.py:55` — all `fts: FTSIndex`.
- `VectorIndex` appears concretely at `managers/search.py:225,232,313`, `managers/index.py:114-115`, `managers/embeddings.py`, `vault.py:971-976`.
- `fts_index.py` is 2366 lines in one class carrying several concerns: the relational store (`documents`/`sections`/`tags`/`links`), FTS search, build-state bookkeeping, TOC, and a contiguous link-graph block (roughly lines 1830–2300).

What each manager actually calls is far narrower than the class, and the subsets are close to disjoint: `link.py` uses the graph methods plus `get_note`; `document.py` uses four read methods; `coordinator.py` uses only build-state. [verified: per-consumer grep of `self._fts.`/`fts.` call sites at 8e20b25]

This is the "index/search stack is concrete" finding in the decoupling study (branch `docs/architectural-studies`, `docs/design/studies/decoupling-and-layering.md` §1.3 / §2.3).

## Why it compounds

`docs/design/design.md` already carries a risk note that at tens of thousands of notes a different vector backend (Qdrant is named) should be evaluated. That evaluation currently has no seam to plug into: substituting a backend means touching six manager constructors and whatever else names the class. Meanwhile the graph queries and the search queries answer different questions — structure versus relevance — but live in one class behind one type, so nothing records that a future backend could serve them separately. [unverified]: whether the vault sizes in use are anywhere near the threshold that would force the backend swap. I have not verified the cause.

## Suggested direction

A starting point, not a prescribed refactor: interfaces for the concerns the study names — keyword/relational, link graph, vector — with `FTSIndex` and `VectorIndex` unchanged as the current implementations. Worth noting before starting that one SQLite database backs both the keyword and graph concerns today, so an interface split has to stay honest about consumers that legitimately need both.

### LINKED ISSUE #1231: [Decay]: the subject-resolution rules exist twice — _okf_write still reads the MCP context directly after #1160
## What

#1160 / #1226 introduced `Principal` and resolved it once at the MCP tool edge, and `git/` is now verifiably free of `fastmcp` imports. But `_okf_write.py` kept its pre-`Principal` fallback branch: when no principal is bound it imports `fastmcp_pvl_core.get_subject` directly and re-implements the `human:<subject>` / `local`-sentinel rules that `_identity.resolve_mcp_principal` also implements. The rules now exist in two places. [verified: read `src/markdown_vault_mcp/_okf_write.py` lines 114-151 at 8e20b25]

## Where

- `src/markdown_vault_mcp/_okf_write.py:126-131` (`resolve_write_actor`) and `:146-151` (`resolve_human_subject`) — the two remaining `from fastmcp_pvl_core import get_subject` reads below the adapter line. They are reached from `DocumentManager` via the write-enrichment hook (`managers/document.py:859`), so domain code transitively depends on the transport.
- `resolve_write_actor` itself has no caller in `src/` — `_server_tools/writer.py` uses `principal.okf_actor()` directly through `write_identity_scope` since #1160. [verified: grep for `resolve_write_actor` under `src/` at 8e20b25 returns only the definition]
- `src/markdown_vault_mcp/config_sections/_assembly.py:135-138` — `configure_identity_claims` is called from inside `_build_git_strategy`, so the git-strategy builder is what configures the identity layer's claim keys.

This is the remainder of the "identity is scattered and MCP-coupled" finding in the decoupling study (branch `docs/architectural-studies`, `docs/design/studies/decoupling-and-layering.md` §1.1 / §2.1) after #1160 closed the main part.

## Why it compounds

Two copies of the same rule can disagree — which is the exact failure #1160 was filed against, in the same file. A driver that is not the MCP server (a CLI, a test exercising the write path) still hits a branch that reaches for the request context, and the dead `resolve_write_actor` is a plausible-looking entry point that a future caller could adopt without noticing it bypasses the bound principal. The claims registration living inside the git builder means the two seams the study treats as separate are entangled at their construction point. [unverified]: whether the two rule copies can currently produce different answers for the same request — they looked equivalent on reading, but I did not construct a case that distinguishes them. I have not verified the cause.

## Suggested direction

A starting point, not a prescribed refactor: one place owns the subject rules, and the other callers reach it. Note when investigating that the deprecated `commit_name_claim`/`commit_email_claim` constructor kwargs shipped in stable v4.1.0 [verified: `git merge-base --is-ancestor` against the v4.1.0 tag], so they are on the public library surface and removing them is a separate, breaking decision — not part of consolidating the rules.

pvliesdonk:
## Closes / Refs

Closes #1229, closes #1230, closes #1231

Also files, but does not close, the deferred seams from the same study: #1232, #1233, #1234, #1235, #1236.

## What & why

Epic #1161 decomposed the god classes internally, but every consumer still named the concrete class, so there was no seam to substitute a backend behind. This installs three of the decoupling study's seams as interfaces, behaviour-preserving:

- **Versioning** (`git/interfaces.py`) — `HistorySource`, `Syncer`, `Versioner`, and `VersionedStore` composing all three. `GitWriteStrategy` stays the single object implementing them: the collaborators share one lock, and `Vault.close()` tells the commit callback from the store by object identity, so splitting into three *objects* would break both. Splitting the type costs nothing. `GitQueryManager` now depends on `HistorySource` alone, the `git_sync` tool on `Syncer`.
- **Search/index** (`interfaces.py`) — `KeywordIndex`, `GraphStore`, `KeywordGraphIndex`, `VectorStore`. `FTSIndex` and `VectorIndex` satisfy them structurally, with no inheritance change. This is the seam the design doc's standing "if tens of thousands, evaluate Qdrant" note needs.
- **Identity** — `_okf_write.py` kept a pre-`Principal` fallback that re-implemented the `human:<subject>` and `"local"`-sentinel rules `resolve_mcp_principal` already implements. Two copies of one rule can disagree, which is the failure #1160 was filed against in that same file. It now resolves a `Principal` instead, and the dead `resolve_write_actor` is gone. `configure_identity_claims` moved out of the git-strategy builder into `to_vault_instances`.

Each protocol carries exactly what its consumers call, not everything the current implementation exposes — a protocol wider than its consumers is a god interface with extra steps.

Three behaviour notes for the reviewer:

1. The `git_sync` managed-mode gate weakens from a nominal `isinstance(strategy, GitWriteStrategy)` to a structural `isinstance(strategy, Syncer) and strategy.is_managed`. All in-tree wiring passes the same object; mypy carries the signature guarantee, and a fake `Syncer` in the new tests pins that the coupling is genuinely gone. That gate previously read four private members of the strategy, which is why `is_managed`, `resolve_force_repo`, `head_sha`, and `branch_name` are now public.
2. `resolve_human_subject` now also reads `get_claims()` on its fallback path. That returns `None` outside a request context, so the resolved subject is unchanged. The existing `resolve_human_subject` tests are deliberately untouched — their passing unchanged is the proof.
3. Renaming `_head_sha` / `_resolve_force_repo` would break an out-of-tree monkeypatcher, but private names carry no compatibility promise and the `markdown_vault_mcp.git.subprocess.run` patch target is untouched.

No commit carries `!`: nothing on the operator surface moves, and the package root's public import surface (`tests/public_import_surface.txt`) is empty, so none of these names were on it.

- Multi-vault (`VaultRegistry`): #1232
- Per-user permissions (`AccessPolicy`): #1233
- Format-agnostic core (`DocumentParser`): #1234
- `ArtifactStore` and the markdown-vs-attachment dispatch duplicated in ~7 places: #1235
- Removing the deprecated `commit_name_claim` / `commit_email_claim` kwargs — they shipped in stable v4.1.0 on an importable class, so removal is breaking: #1236
- Splitting `GitWriteStrategy` into separate objects, or extracting the OKF dialect hooks. The adapter layer needed no work: all 77 `asyncio.to_thread` calls were already confined to the MCP modules.

## Local review

      existed at the **last stable release** (see the breaking-change policy
      in `AGENTS.md`) — MCP tool-surface changes alone do not earn a `!`.

## Docs impact

- [ ] `docs/` site pages — no MCP tool, resource, prompt, or env var changed.
- [x] `docs/design/` — new `interfaces.py` section; facet protocols documented in the git section; `fts_index.py` / `vector_index.py` conformance noted; the Qdrant risk note now points at `VectorStore`; the identity section records that the subject rules have one owner.
- [x] Inline docstrings — both new modules, the promoted public methods, and the retyped consumers.

`AGENTS.md`'s Project Structure tree carries both new modules.

---
_Generated by [Claude Code](https://claude.ai/code/session_0197qu2QerLNpJSGQAjLjSNN)_

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -54,6 +54,7 @@ src/markdown_vault_mcp/
     query.py           -- read-only git history/diff queries; lock-free pure functions
     types.py           -- PullResult/PushResult + pull/push reason-code constants
   scanner.py           -- file discovery, frontmatter parsing, chunking
+  interfaces.py        -- KeywordIndex/GraphStore/KeywordGraphIndex/VectorStore: the search/index storage seam (#1230)
   fts_index.py         -- SQLite FTS5 schema, BM25 search
   _fts_connection.py   -- per-thread sqlite connection registry + SQLITE_LOCKED retry (#760)
   vector_index.py      -- numpy embeddings, cosine similarity
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -46,6 +46,7 @@ src/markdown_vault_mcp/
   git/
     __init__.py        -- package facade preserving the historical single-module import surface (incl. test patch targets)
     _run.py            -- low-level git subprocess + credential plumbing
+    interfaces.py      -- HistorySource/Syncer/Versioner/VersionedStore: the versioning seam; GitWriteStrategy is the one implementation (#1229)
     strategy.py        -- GitWriteStrategy: auto-commit per write; composes RepoBootstrap + PushScheduler over one shared lock (#893)
     bootstrap.py       -- RepoBootstrap: managed-clone bootstrap, remote-protocol validation, memoised git-root discovery (#893)
     push_scheduler.py  -- PushScheduler: deferred-push timer + pending-flag mechanics and push execution (#893)
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -69,7 +69,7 @@ src/markdown_vault_mcp/
   okf_bundle.py        -- OKF bundle-zip export from live vault state, served via an okf-bundle download ref (#963)
   _okf_convention.py   -- OKF reserved-file maintenance after enforced writes: log.md bullet + index.md refresh (#964)
   _okf_write.py        -- OKF enforced-write runtime: contextvar actor + provenance stamp / verified clear (#964)
-  _identity.py         -- Principal write identity: tool-edge resolution, contextvar carry, claim-key registration (#1160)
+  _identity.py         -- Principal write identity: tool-edge resolution, contextvar carry, claim-key registration (#1160); sole owner of the subject rules (#1231)
   summarizer.py        -- Summarizer ABC + OpenAI-compatible chat-completions backend (#915)
   vault.py             -- thin composition root: settings-first dual-mode construction (#1158), lifecycle, wiring, facet accessors (index-write → indexing/coordinator.py)
   write_callback.py    -- WriteCallbackDispatcher: deferred git-commit callback worker (#599)
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -2458,6 +2458,16 @@ untouched. `GitWriteStrategy.__call__` consumes the Principal's
 `display_name`/`email` as the commit `--author` (committer stays the static
 identity); the `git/` package imports nothing from `fastmcp` (guard-tested),
 so non-MCP drivers supply a `Principal` instead of faking request context.
+
+`resolve_mcp_principal()` is the **only** place the subject rules live
+(#1231) — `human:<subject>`, and the `"local"` sentinel counting as no human
+identity. `_okf_write.py` used to re-derive them in a fallback branch that
+read `get_subject()` itself; it now resolves a `Principal` instead, so the
+rules cannot drift between two copies. The dead `resolve_write_actor()` went
+with that branch: since #1160 the write tools stamp
+`principal.okf_actor(...)` directly. Registering the claim keys likewise
+moved out of the git-strategy builder into `to_vault_instances`, so
+configuring identity is not the git builder's job.
 Background paths without an acting person — the pull-loop conflict commits,
 the webhook, the transfer-route uploads — carry no Principal and use the
 static identity / tool actor, as before.
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -2529,6 +2529,12 @@ Uses the schema defined in [Database Schema](#database-schema). Note that
 with RRF scoring in hybrid mode; each file appears once with up to
 `chunks_per_file` matching sections).
 
+`FTSIndex` satisfies the `KeywordIndex` and `GraphStore` protocols
+(`interfaces.py`, #1230) structurally — it does not inherit from them. The
+managers are typed against those protocols rather than against this class, so
+the concrete backend is substitutable; `Vault` constructs the `FTSIndex` and
+that construction is where mypy proves the conformance.
+
 ### `vector_index.py`: Numpy Embeddings
 
 Adapted from ifcraftcorpus `embeddings.py`. Rename `EmbeddingIndex` to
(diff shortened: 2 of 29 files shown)


pvliesdonk:
@claude review
@codex review

PR #1260, opened 2026-09-01
fix(docs): adopt template v7.0.1 and de-fork the OIDC pages

### LINKED ISSUE #1248: [Bug]: the configuration reference documents only oidc-proxy and marks its client credentials Required, so readers configure the fallback mode
### What happened?

`docs/configuration.md` presents oidc-proxy as if it were *the* OIDC configuration: it marks the client id and secret as **Required: Yes** and states that OIDC activates when all four variables are set. Following that page lands the reader in the mode the deployment guide calls a fallback. The recommended mode is not reachable from that page, because reaching it means *omitting* two variables the page says are required.

### Observed

`docs/configuration.md:421-427`, inside the `DOMAIN-CONFIG-VARS` block:

> Optional token-based authentication for HTTP deployments. OIDC activates when all four required variables are set.

| Variable | Required |
|---|---|
| `MARKDOWN_VAULT_MCP_BASE_URL` | Yes |
| `MARKDOWN_VAULT_MCP_OIDC_CONFIG_URL` | Yes |
| `MARKDOWN_VAULT_MCP_OIDC_CLIENT_ID` | Yes |
| `MARKDOWN_VAULT_MCP_OIDC_CLIENT_SECRET` | Yes |

`docs/deployment/oidc.md:12` says the opposite about the same two variables:

> | **remote** (recommended) | `BASE_URL`, `OIDC_CONFIG_URL` | Local JWKS validation. No client credentials needed. |
> | **oidc-proxy** | `BASE_URL`, `OIDC_CONFIG_URL`, `OIDC_CLIENT_ID`, `OIDC_CLIENT_SECRET` | Full OAuth proxy with session management. |

Three things make the wrong choice hard to notice:

1. **The word "remote" does not appear in `configuration.md`'s OIDC section.** A reader who never opens the deployment guide has no way to learn a second mode exists.
2. **`MARKDOWN_VAULT_MCP_AUTH_MODE` is not documented on the site at all** — it is one of the three undocumented universal variables in #1246. So even a reader who suspects there is a choice cannot find the variable that states it.
3. **The auto-detected mode is never logged.** `resolve_auth_mode` (`fastmcp-pvl-core` 4.11.3, `_auth.py:83`) returns `oidc-proxy` when all four are set, and the only INFO line about the mode is on the explicit-override path (`_auth.py:74`, `"auth_mode=%s (explicit via AUTH_MODE)"`). Auto-detection logs nothing at INFO or above, so a running server gives no signal which of the two modes it chose.

The result is a configuration surface where the recommended mode is expressed only by absence. Setting four variables is an affirmative act; omitting two is not, and nothing distinguishes "I chose remote" from "I did not finish configuring".

Reported by the maintainer as a recurring pattern rather than a one-off: readers, LLM agents configuring the server among them, repeatedly end up in proxy mode. This report came from exactly that — a deployment walkthrough written against `configuration.md` that set all four variables and silently selected the fallback.

The cost is not only conceptual. Proxy mode needs a confidential client with a redirect URI pointing at the MCP server, so the wrong choice propagates into the identity provider's configuration before anyone notices.

### Expected

The configuration reference names both modes, marks the client credentials as required only for the mode that needs them, and gives a reader enough to choose deliberately. A running server can be asked which mode it is in.

### Reproduction

Read `docs/configuration.md:419-433`, then `docs/deployment/oidc.md:8-15`. Mode selection read from `resolve_auth_mode` at `_auth.py:83` and the dispatch at `_auth.py:614` in the installed `fastmcp-pvl-core` 4.11.3; logging checked by enumerating `logger.info` / `logger.warning` calls in that file. I did not start a server and observe the selected mode.

### Suspected cause

`[unverified]` The section likely predates remote mode and was not revisited when it landed, which would make it the same class as #1247 — a hand-written block on this page drifting from generated content elsewhere. I have not verified the cause; I did not check when remote mode was introduced relative to this section.

### Open questions / scope

- Whether the deeper fix belongs in `fastmcp-pvl-core` rather than here: expressing "I want remote" currently requires omitting variables, and there is no affirmative spelling short of `AUTH_MODE`. Logging the resolved mode on the auto-detect path would also cost one line. Not filed upstream pending a view on whether the documentation fix is sufficient.
- `[unverified]` Whether adopting the generated configuration page (#1246) fixes this by itself. The generated OIDC table in `docs/deployment/oidc.md` is correct because its heading scopes it to proxy mode; this page's required-marking is hand-written domain content, so it may survive the adoption unchanged.
- Overlaps #1247, which reports one row of the same table, and #1246, which reports the same page's opening. Separate observations, same section.

pvliesdonk:
Closes #1248

Template `v7.0.0` → `v7.0.1`. That release carries exactly one change, pvliesdonk/fastmcp-server-template#559, which is the upstream fix for #1248: the generated configuration reference marked the oidc-proxy client credentials **Required** under a plain `## Authentication` heading, with `remote` named nowhere on the page. Setting four variables is an affirmative act and omitting two is not, so a reader following that page landed in proxy mode with nothing distinguishing "I chose remote" from "I did not finish configuring" — and proxy mode needs a confidential client with a redirect URI, so the wrong choice reached the identity provider before anyone noticed.

## The de-fork is the point

`docs/deployment/oidc.md` and `docs/guides/authentication.md` conflicted, both times because this project had already corrected the same prose locally — *outside* their `DOMAIN-OIDC-EXTRA` / `DOMAIN-AUTH-EXTRA` sentinels, in files copier re-renders. That is a fork of template-owned prose, and it is why these two pages have re-conflicted on every update.

All four hunks resolved to the template side, per the resolution policy. The fork is gone and these pages should now update cleanly.

## Nothing was dropped on the floor

Local content the template side displaced moved into the sentinel blocks instead:

- `DOMAIN-AUTH-EXTRA` gains a **Choosing an OIDC mode** section carrying the three things the template deliberately does not assert: that remote is the better default *here*, the operational reason (oidc-proxy re-validates the upstream token on every request, so the shortest upstream token lifetime becomes the session lifetime — which is exactly the failure the Known Limitations section directly above describes), when oidc-proxy is still right (a provider needing Dynamic Client Registration emulated, or the proxy's session management), and how to move an existing deployment across.
- `DOMAIN-OIDC-EXTRA` gains a pointer to it, so the reference page still routes a reader to the recommendation.

The upstream page deliberately states only the mechanical difference between the modes — an editorial "recommended" would have to hold for six downstreams. Here it is this project's own position, so it lives in this project's own block.

## One thing that would have broken quietly

`docs/guides/oidc-providers.md` — a project-owned page the template does not touch — linked twice to `authentication.md#remote-mode`, an explicit `{ #remote-mode }` anchor that only existed on the forked heading. Taking the template side drops it. Both links retargeted to the template heading's own anchor (`#how-remote-mode-works`) rather than pinning the anchor back onto a template-owned line, which would have re-forked the page on the next update. `mkdocs build --strict` confirms both resolve.

## Gate

Run in full per the template-update procedure:

| Check | Result |
|---|---|
| conflict markers | none remain |
| `git diff --check` | clean |
| `uv lock --check` / `uv sync --locked` | clean |
| `scripts/gen_config_surface.py --check` | 116 vars, no drift |
| `scripts/vendor_spa.py --check` | up to date |
| ruff check + format | pass |
| mypy `src/ tests/` | 217 files, no issues |
| pytest | 3896 passed, 4 skipped |
| `mkdocs build --strict` | pass |
| `pre-commit run --all-files` | pass, Vale included |

`.copier-seeded-changes.md` reports no seeded file changed between the two refs, and `UPGRADING.md` has no section for this jump — a patch within the same minor whose release notes name only #559.

## Noted, not fixed here

`mkdocs build --strict` emits four INFO lines about links into `configuration.md` anchors that no longer exist (`#file-watcher`, `#github-webhook-push-triggered-pull`, `#one-time-transfer-links`, `#write-protection`). They are pre-existing — `origin/main` emits the same four — and come from that page becoming generated in #1255, not from this change. Left alone rather than swept in.


pvliesdonk:
@claude review

PR #1262, opened 2026-09-01
feat(sync): push-triggered reindex for GitLab-hosted vaults

### LINKED ISSUE #1178: [Feature]: push-triggered reindex for GitLab-hosted vaults

A managed vault hosted on GitLab has no way to reach the staleness window that a GitHub-hosted one gets. The push-triggered pull path is GitHub-specific end to end, so a GitLab vault is left with the periodic pull loop, and its staleness is whatever `MARKDOWN_VAULT_MCP_GIT_PULL_INTERVAL_S` is set to.

What the server does today, at `134b1c1`:

- `server.py:216-230` mounts `POST /github-webhook` only when `config.sync.github_webhook_secret` is set.
- `_verify_signature` (`src/markdown_vault_mcp/_github_webhook.py:41`) validates `X-Hub-Signature-256: sha256=<hex>` as an HMAC-SHA256 of the raw body, constant-time.
- `docs/configuration.md` puts the resulting window at roughly 2 seconds, against up to `GIT_PULL_INTERVAL_S` for the poll.

The polling fallback is not a close substitute for a vault with a second author. Tightening the interval costs a git fetch per tick against the remote, and the filesystem watcher — the other fast path — is deliberately disabled whenever the pull loop is active, to avoid scanning a working tree mid-checkout. So on GitLab there is no configuration that gets external edits indexed promptly.

GitLab does send push deliveries; the shapes just differ from GitHub's, which is why the existing route does not accept them as-is. `[verified: GitLab webhook documentation, docs.gitlab.com/user/project/integrations/webhooks/, read 2026-08-27]`

- Event type arrives as `X-Gitlab-Event`, valued `Push Hook` for a push.
- Authentication has two forms: a signing token computing an HMAC-SHA256 over the payload, delivered as `webhook-signature: v1,{base64_signature}`; and a legacy plain-text secret in `X-Gitlab-Token`, which GitLab describes as offering weaker guarantees and no longer recommends for new webhooks.

So both the header names and the signature encoding differ from GitHub's — base64 behind a `v1,` prefix rather than hex behind `sha256=`. The configuration surface is GitHub-specific too: the variable is named `GITHUB_WEBHOOK_SECRET` and the route is `/github-webhook`.

Encountered while setting up a GitLab-hosted vault alongside a GitHub-hosted one, where the two vaults could not be given the same sync behaviour.

### Desired outcome

A managed vault on GitLab reaches the same staleness window as one on GitHub, without polling the remote more aggressively and without giving up the safety property that keeps the watcher and the pull loop from running at once.

### Implementation ideas

Non-binding; this issue does not prescribe a shape. Two constraints worth carrying into whatever design is chosen: the verification differences above mean the existing HMAC check is not reusable verbatim, and the current variable and route names are host-specific, so where a second host lands relative to them is a decision this issue leaves open.

### Open questions / scope

- Not about GitLab CI, GitLab authentication, or any other GitLab integration — only push-triggered reindex.
- Whether the legacy plain-token form should be accepted at all, given GitLab's own guidance, is undecided here.
- `[unverified]`: other hosts (Gitea, Forgejo, Bitbucket) may present the same gap. Not checked, and out of scope for this issue.

pvliesdonk:
Closes #1178

A GitLab-hosted managed vault had no way to reach the staleness window a GitHub-hosted one gets. The push path was GitHub-specific end to end, so GitLab was left with the periodic pull loop — and tightening that costs a fetch per tick against the remote, while the file watcher is not an alternative because it is deliberately disabled whenever the pull loop runs. On GitLab there was no configuration that got external edits indexed promptly.

## The issue body's signature description is wrong

Worth leading with, because an implementation built from it would never verify. #1178 says GitLab sends "a signing token computing an HMAC-SHA256 over the payload". It does not. GitLab follows the [Standard Webhooks](https://www.standardwebhooks.com/) specification, and the HMAC covers `{webhook-id}.{webhook-timestamp}.` concatenated with the raw body. `[verified: docs.gitlab.com/user/project/integrations/webhooks/, read 2026-09-01]`, whose own Ruby example is:

```ruby
message  = "#{message_id}.#{timestamp}.#{body}"
digest   = OpenSSL::HMAC.digest('sha256', raw_key, message)
expected = "v1,#{Base64.strict_encode64(digest)}"
```

Three consequences the body does not mention, each of which changes the code: `webhook-id` and `webhook-timestamp` arrive as separate headers and are both digest inputs; the timestamp needs a freshness check, which GitLab's docs require explicitly; and the header may carry several space-separated signatures. All three are implemented and pinned by tests. Correction also posted on the issue.

## One handler, two hosts

Both hosts notify the same fact and the response is identical, so this adds a `WebhookProvider` record for what actually differs — how a delivery authenticates, the event header and its push value, the delivery-id header, whether the host has a handshake event — over one shared handler. The routes stay separate because the *credentials* are, not because the logic is.

`_github_webhook.py` becomes `_webhooks.py`: half its traffic is no longer GitHub's, and a module name that lies is worse than a rename touching 15 lines. Log prefixes now come from the provider. **The 30 existing GitHub tests pass against the refactor unchanged**, which is the evidence that the extraction preserved behaviour.

`/github-webhook` and `GITHUB_WEBHOOK_SECRET` are untouched, so no operator has anything to change.

## Two GitLab credentials, deliberately

`GITLAB_WEBHOOK_SIGNING_TOKEN` and `GITLAB_WEBHOOK_SECRET_TOKEN` are separate fields because GitLab has two authentication mechanisms with genuinely different security properties, not one spelled two ways. They are named after GitLab's own UI labels, so an operator can match the field they are filling in.

The issue leaves open whether the plain-text form should be accepted at all, "given GitLab's own guidance". The version facts decide it: **`webhook-signature` was introduced in GitLab 19.0 and became generally available in 19.1** (the `webhook_signing_token` flag was removed in 19.1). `[verified: GitLab 19.0 release notes + gitlab-org/gitlab#480633]` Refusing `X-Gitlab-Token` would leave push-triggered reindex unreachable on every self-managed install below 19.0 — the deployment this issue was filed from. So both are accepted, the signing token is tried first, and startup logs a warning when the plain token is the only credential, so the weaker choice is visible rather than silent. Setting both accepts either, which is what lets a live webhook migrate.

## The replay window

`GITLAB_TIMESTAMP_TOLERANCE_S` is 300 s, the Standard Webhooks convention, applied in **both** directions — a clock ahead of ours is as suspect as one behind. Without it a captured delivery replays forever. Because the timestamp is inside the digest, re-stamping a captured delivery to look fresh invalidates its signature; both halves have tests, including the inclusive window edge so ordinary skew is not a rejection.

## The file-watcher gate was a correctness bug waiting to happen

`should_start_file_watcher` took `github_webhook_secret` and computed `git_active = git_pull_active or bool(github_webhook_secret)`. Left alone, the watcher would have run alongside GitLab-driven checkouts — precisely the mid-checkout partial scan the gate exists to prevent. It now takes `SyncConfig.webhook_configured`, which covers every host, and a parametrized test asserts each of the three credentials flips it.

## Complexity

The ping-event guard pushed `handle()` past the budget (C901 12 > 10, 9 returns > 6) and the structural gate caught it. Rather than a `noqa`, the push phase is now `_process_push`: the request-shaped concerns and the git-shaped ones were always two phases in one function. Gate back to 100%.

## Not in scope

- Gitea, Forgejo, Bitbucket — explicitly out per the issue, and no plugin registry was built for hosts nobody asked for.
- **No `INDEX_SEMANTICS_VERSION` bump.** Nothing changes how a note's stored rows derive from its bytes.

## Gate

`pytest` 3950 passed, 3 skipped · ruff check + format · mypy 218 files · structural gate 100% · `pre-commit run --all-files` including Vale · `gen_config_surface.py` regenerated (`.env.example`, `packaging/env.example`, `examples/*.env`, `server.json`, `wizard-spec.json`, `docs/configuration.md`) · the 15 config-wizard smoke tests exercise the regenerated spec.

`_webhooks.py` is at **100%** line and branch coverage. Getting there removed an unreachable `None` guard in `_timestamp_is_fresh` rather than writing a test for a branch its only caller cannot reach.

Docs: `docs/guides/git-integration.md` restructured for both hosts (including a GitLab setup walkthrough and the migration path), `docs/design/design.md` gains the design rationale, `README.md` and the module map updated. The guide's outbound link to a `configuration.md` anchor that no longer exists is fixed in passing, since that line was being rewritten anyway.


pvliesdonk:
Addressed the Repowise health-gate findings.

**`make_server` complexity (critical, "worsened").** Real, and the gate was right to flag it: the wiring block was `make_server`'s second-largest branch and this change was adding a host to it. The structural gate missed it because `C901` reports at the function's definition line, which is not in the diff — pre-existing debt the diff gate deliberately does not block.

Fixed by moving the mounting decision to `register_webhook_routes` in `_webhooks.py`, beside the handlers it mounts. `server.py`'s DOMAIN-WIRING block is now one call.

| | `make_server` cyclomatic complexity |
|---|---|
| `origin/main` | 22 (D) |
| this PR, before the move | 26 (D) |
| this PR now | **18 (C)** |

So the feature lands four points below the baseline rather than four above it.

**`TestWebhookWiring` low cohesion (LCOM4=4, high).** Also fair — twelve methods sharing no state, and the two hosts' gates are genuinely separate concerns. Split into `TestGitHubWebhookWiring` and `TestGitLabWebhookWiring` over a module-level route helper.

**Two broad `except Exception` (low, ×2).** Not changed, deliberately. Both are pre-existing, carried over by the rename rather than written here — the attribution reads `new_file` because `_github_webhook.py` → `_webhooks.py` makes the whole file new to the analyser. Each is load-bearing and documented in place: the handler's contract is 401 / 200 / 503, so an exception escaping `force_pull` would surface as an unhandled 500 with a traceback (#1128), and a failed reindex must not turn a successful pull into a retry that pulls again over a half-updated index. Narrowing either would reintroduce the bug it was written to fix.

pvliesdonk:
@claude review
@codex review

pvliesdonk:
Both review findings addressed. The P1 was correct and worse than reported — thank you, that would have shipped a feature that rejected every real delivery.

## P1 — signing token decoded before the HMAC (fixed)

Verified at the source before changing anything. GitLab's own verification example derives the key first:

```ruby
raw_key = Base64.strict_decode64(signing_token.delete_prefix('whsec_'))
digest  = OpenSSL::HMAC.digest('sha256', raw_key, message)
```

`[verified: docs.gitlab.com/user/project/integrations/webhooks/, read 2026-09-01]` Two corrections to the finding as written: there is also a **`whsec_` prefix to strip**, and the encoding is **standard** base64, not URL-safe.

`gitlab_hmac_key` now does that decode once when the provider is built, so a token that cannot be a key is a startup `ConfigurationError` naming the variable rather than a 401 per delivery. A missing prefix warns rather than fails — the prefix is GitLab's convention, not key material, and a future version could change it.

**The tests were the actual defect.** They reproduced the same wrong derivation, so they proved the implementation agreed with itself rather than with GitLab. They now build the key from the token the way GitLab's example does, and two state the contract directly: a delivery signed with the decoded key authenticates, one signed with the token text does not. Reverting the fix (`hmac_key = signing_token.encode()`) fails **ten** tests including both, which I checked by mutation rather than assuming.

**The documentation was wrong in the same direction** and is fixed too: it told operators to generate the signing token with `openssl rand -hex 32`. GitLab generates that one and displays it once; only the plain secret token is operator-chosen. `[verified: same page — "Select Generate signing token. Copy and save the token now because it is displayed only once."]`

## P2 — watcher on stdio (filed, not fixed here)

Correct observation, and it is **pre-existing**: before this PR the gate took `github_webhook_secret` and behaved identically, so the GitLab credentials inherit the hole rather than introduce it. `[verified: read the pre-#1178 signature and body of should_start_file_watcher]`

Not fixed here because the gate cannot see the transport: `Service` is constructed no-args by the template's `server_lifespan` and has no transport at all, so closing it means plumbing one through — more than this issue should carry, and a behaviour change for existing GitHub stdio deployments too.

Filed as #1263 with the three call sites and both candidate fixes (pass the transport, or warn that the credential is inert on stdio).

## Also in this push

`S105` fired on `GITLAB_SIGNING_TOKEN_PREFIX = "whsec_"` — a published format marker, not a credential. Handled with a `per-file-ignores` entry rather than a `# noqa`, matching the convention already in `pyproject.toml` for gate-only rules, which RUF100 would otherwise strip.

Gate: 3942 passed · `_webhooks.py` 100% line and branch · structural gate 100% · mypy 218 files · Vale clean.

pvliesdonk:
@claude review
@codex review

pvliesdonk:
Second review round addressed — three items, all verified before changing anything.

## Codex P2 — non-ASCII credential returns 500 instead of 401 (fixed)

Confirmed, and it reaches **all three** comparison sites, not the two named. GitHub's `X-Hub-Signature-256` path predates GitLab's and shares the exposure exactly, so I swept the class rather than patching the reported instances:

```
gitlab signing -> RAISES TypeError comparing strings with non-ASCII characters is not supported
gitlab secret  -> RAISES TypeError comparing strings with non-ASCII characters is not supported
github         -> RAISES TypeError comparing strings with non-ASCII characters is not supported
```

`verify` runs outside the handler's exception guard, so the documented 401 became an unhandled 500 with a traceback. All three now route through one `_credentials_match` helper comparing encoded bytes: same timing property, and a value that could never legitimately match simply does not match. Every value that *can* match — a hex digest, a base64 digest, a configured token — is ASCII, so the encoding is byte-identical to the wire for anything that matters.

Four tests pin it, including one end-to-end asserting 401 through the handler. That one hands the client raw header bytes, because httpx refuses to encode a non-ASCII `str` header — a hostile client has no such scruples. Reverting `_credentials_match` to the raw `compare_digest` fails all four, mutation-checked rather than assumed.

## claude — design.md credited the wrong module (fixed)

Correct, and it contradicted the paragraph three sentences below it. The `webhook_inert` warning moved to `register_webhook_routes` in 704b406; `design.md:3751` still said `server.py`. This is exactly the stale-doc class where a doc describes changed code without being in the diff, so thank you for catching it.

## claude — window-edge test was near the edge, not on it (fixed)

Fair. It used `now - TOLERANCE + 1` (299 s) while claiming to pin an inclusive boundary. It now sits at exactly `GITLAB_TIMESTAMP_TOLERANCE_S` with an injected clock, and I verified the boundary behaves as the docstring now says:

```
offset 300s -> True
offset 301s -> False
```

## Not changed

P2 from the first round (watcher inert on stdio) stays out of scope and filed as #1263 — both reviewers agreed it is pre-existing and needs the transport threaded into `Service`.

Gate: 3961 passed · `_webhooks.py` 100% line and branch · structural gate 100% · mypy 218 files · Vale clean.

pvliesdonk:
@claude review
@codex review

pvliesdonk:
Third round addressed. From here I run a full local review of the cumulative diff before every push, not just before opening the PR.

To be clear about what these rounds were: I did the research and wrote the design, so every one of these defects was mine to catch before it reached a reviewer. The signature was built from an issue summary and a truncated docs excerpt, with tests that could only confirm my own assumption.

## Empty decoded signing key (fixed)

A genuine fail-open. `whsec_` on its own decodes cleanly to `b""`, HMAC accepts an empty key, and the route would have mounted authenticating any forged delivery. A zero-length key is now a startup `ConfigurationError`. Removing the guard fails both new tests.

## URL-safe alphabet

GitLab's documented examples decode with the **standard** alphabet, in both languages:

```ruby
raw_key = Base64.strict_decode64(signing_token.delete_prefix('whsec_'))
```
```python
raw_key = base64.b64decode(signing_token.removeprefix('whsec_'))
```

`Base64.strict_decode64` rejects `-`/`_` outright, so a URL-safe token would fail GitLab's own example, and the docs describe the alphabet as URL-safe nowhere. `[verified: docs.gitlab.com/user/project/integrations/webhooks/, 2026-09-01]`

Handled anyway: the token is a value this server neither controls nor can re-issue, and the failure mode if that reading is wrong is "the server refuses to start". The decode now folds `-`/`_` onto `+`/`/` before validating. The fold is unambiguous — a token valid under both alphabets contains only the characters they share — and costs one `str.replace`. Not a two-decoder fallback: `urlsafe_b64decode` takes no `validate` parameter, so the fallback path could not reject a malformed token as strictly as the primary one.

## Over-long design.md lines (fixed)

Rewrapped, and the paragraph now also records both new guards.

## Also in this push

The legacy secret-token path was tested only by passing plain dicts straight to `verify`, which cannot show the header name matches what a request actually carries — and it is the credential every self-managed GitLab below 19.0 will use. It now has an end-to-end round trip through real request headers, one authenticating and one refusing. Renaming the header in the source fails three tests including the new one.

Two candidates were checked and refuted rather than reported: header lookup is case-insensitive through Starlette (my first probe suggested otherwise because I built the `Headers` object with un-lowercased raw keys, which real ASGI servers never produce), and the log-message shape follows the pattern already established in this module.

Gate: 3966 passed · `_webhooks.py` 100% line and branch · structural gate 100% · mypy 218 files · pre-commit clean.

pvliesdonk:
@claude review
@codex review

PR #1265, opened 2026-09-01
feat(git): commit once per tool call, not once per file

### LINKED ISSUE #1264: [Feature]: Commit granularity does not match tool-call granularity

A single MCP tool call that writes many files produces one commit per file, and blocks every concurrent write until they all finish.

Observed on a personal vault (3,918 indexed documents) running 4.1.0 with `MARKDOWN_VAULT_MCP_GIT_REPO_URL` set:

Calling `okf_convert_links` once, with no folder argument, returned:

```
files_changed: 2595   links_converted: 7620   links_skipped: 2   notes_scanned: 3918
```

That one call produced **2,596 commits**, each touching exactly one file:

```
80d7ca5 write: troubleshooting/herd-no-code-coverage-driver.md
d4f047d write: troubleshooting/cowork-artifact-callmcptool-400.md
3b0e512 write: topics/zed-upstream-outline-highlight-fix.md
...
```

`[verified: how]` — `git rev-list --count` between the commit preceding the call and the one following it returned 2596; `git show --name-only --format=''` on sampled commits returned exactly one path each.

The server log for the same window shows 2,599 `Git: committed` lines against 3 `Git: pushed to remote` lines, so the deferred-push batching worked as documented — it is the commit cadence, not the push cadence, that tracks file count. `[verified: how]` — `grep -c` on `server.log`.

While those commits were being made, two consecutive `edit` calls returned no response and were aborted by the client at its 300 s idle timeout. `stats` called between them answered immediately, and the server remained listening throughout. `[verified: how]` — client-side timeout errors observed twice; `stats` round-trip in between returned normally; `lsof` confirmed the listener was up.

I have not verified that the timeouts were caused by the commit queue. `[unverified]` — the write-callback dispatcher drains a single FIFO worker thread, so a queue of 2,595 sequential `git commit` invocations ahead of a later write is consistent with what was observed, but I did not instrument the worker to confirm the later `edit` was waiting rather than failing for another reason.

Two consequences on a synced vault:

- History granularity does not match intent. One logical migration is 2,596 entries, so `git log` no longer answers "what changed and why" at the level a reader asks it.
- The vault is effectively single-threaded for writes for the duration of any bulk call.

### Desired outcome

A commit whose contents correspond to one logical operation, so that history reads at the granularity of what caused the change rather than of how many files it touched, and so that a bulk operation does not serialise unrelated writes behind it.

Per-file commits remain the right granularity for a single-file `write` or `edit`; the gap is only where one call touches many files (`okf_convert_links`, `move_folder`, `rename` with `update_links=True`).

### Implementation ideas

#54 analysed this problem and recommended Option A (commit per write, deferred push), which is what shipped. Its Option B (one commit per LLM intent, via an explicit `commit` tool) was ranked lower on one stated axis: *"LLM must reliably call `commit` — risk of forgotten flush."*

That issue also proposed a `MARKDOWN_VAULT_MCP_GIT_MODE` env var (`auto` / `manual` / `immediate`) to let the two coexist. `[verified: how]` — no `GIT_MODE` key appears in `config.py`'s env reads on `main` (`2856c3c4`), so the knob appears never to have been built.

A boundary derived from the in-flight tool call rather than from an LLM-issued `commit` would reach Option B's outcome without its flush-risk. I have not verified that this is the right seam for this codebase.

`[unverified]`: `move_folder` and `rename(update_links=True)` may share this path, since both report multiple files to the write callback (#1244).

### Open questions

- Whether this should be gated behind the `GIT_MODE` knob #54 proposed, or change the default.
- Whether the 300 s write timeouts above are the same problem or a separate one worth its own issue.

---

mikebronner:
## Closes / Refs

Closes #1264

## What & why

One `okf_convert_links` call over a 3,918-document vault produced **2,595
commits** for one logical migration, and concurrent MCP writes queued behind
2,595 sequential `git commit` invocations. This makes the commit boundary the
**tool call** rather than the file, so history reads at the granularity of what
caused a change.

`CommitScopeMiddleware` binds a scope for each `on_call_tool`; the dispatcher
snapshots it in `fire()`, buffers writes by token, and commits each group once
as `"<tool>: N files"`.

A call that wrote a **single** file keeps that file's own path as the subject
(`write: note.md`), so ordinary writes read in `git log` exactly as they always
have. That is what makes the boundary unconditional rather than a mode: the two
boundaries differ only where one call touches many files, which is the defect
#1264 reports and not a preference to configure. #54 anticipated a
`GIT_MODE` knob; none is added, and no configuration changes.

Three properties it depends on, each pinned by a test:

- **The scope is read in `fire()`, never on the worker.** `fire()` runs on the
  request's `to_thread` worker whose copied context carries the contextvar; the
  dispatcher thread has no request context — the same constraint as
  `current_principal` (#1218/#1226).
- **Grouping is keyed by token, not queue position.** The queue is FIFO, but a
  shared server interleaves concurrent calls, and position-based grouping would
  fold two unrelated calls into one commit.
- **A drain flushes open scopes.** `drain()` is what the puller waits on before
  merging (#571/#677); returning with a group still buffered would let the merge
  run over writes that are on disk but in no commit. A drain landing mid-call
  splits that call into two accurate commits rather than one.

**Backward compatible by construction.** A write with no owning tool call still
commits alone, and a callback that has not set `ACCEPTS_BATCH_ATTR` is never
buffered at all — it is fired per write as it arrives, keeping both its contract
and its timing, mirroring the `ACCEPTS_OLD_PATH_ATTR` and
`ACCEPTS_PRINCIPAL_ATTR` opt-ins. Staging stays per item and scoped, so a batch
never widens what a delete or rename sweeps in (#894, #1244); only the commit is
shared. `WriteBatchItem` deliberately drops the written content, which nothing
on the batch path reads — staging reads the file from disk.

`_stage_one`, `_commit_staged` and `_CommitIdentity` are extracted so the
single-write and batched paths share one copy of the staging rules and of the
commit-header injection guard.

- **Does not squash or rewrite history** — #54's "squashing verdict: don't"
  still holds. This changes where the commit boundary is drawn, not what happens
  to commits after they exist.
- **Does not change push cadence.** Deferred push is untouched; it already
  batched correctly (2,599 commits produced only 3 pushes in the reported case).
- **Does not address the 300 s write timeouts** as a separate concern. They are
  consistent with queueing behind the commit backlog and should disappear with
  this change, but causation was not instrumented — noted as `[unverified]` in
  #1264.
- **Adds no configuration.** An earlier revision of this branch gated the
  boundary behind `MARKDOWN_VAULT_MCP_GIT_COMMIT_MODE`; that knob is gone, and
  no generated config surface changes.

## Docs impact

- [x] `docs/design/` — "Write + Git Integration" rewritten; `module-map.md`
      gains `_commit_scope.py`
- [x] Inline docstrings — `GitWriteStrategy` and the dispatcher module docstring
      both said "per write" and were corrected
- [ ] `README.md` / `docs/` site pages — no user-facing config or tool surface
      changed

Three pre-existing `ai-tells.VerbTricolon` errors in the generated config help
text (the `docs/configuration.md` preamble, `WRITE_PROTECT_EXISTING`,
`OKF_VERIFY`) are reworded at their source in `config.py`, for sentence
structure only with no change of meaning. CI stays green on them either way, but
the pre-commit `vale` hook lints any changed file, so anyone regenerating the
config surface hits them locally.

## Test status

3,932 pass, 18 skip. Structural gate 100% (0 violations); `mypy` clean;
`_commit_scope.py` at 100%.

`TestScopeSurvivesARealToolCall` drives `move_folder` and `write` through a real
`Client(server)` against a git-backed vault. Every other test in the file binds
the scope and fires on one thread — the case that works trivially — and
`tests/test_git.py` has no `Client(server)` at all, so its 33 commit-subject
assertions never bind a scope. Mutation-checked: forcing `fire()` to read `None`
makes the bulk case fail with three per-file rename commits instead of one.

One claim is untested through the wire: the middleware's `finally` is unit-tested
to close the scope when a tool raises, but "a tool that raises after writing
still commits what it wrote" has no end-to-end test — no existing tool writes and
then fails (`move_folder` aborts at the gate before moving).

---

Co-authored-by: Peter van Liesdonk <peter@liesdonk.nl>


mikebronner:
ok, all done now. please let me know if you would like any changes

pvliesdonk:
Thanks for this — the diagnosis in #1264 and the architecture here are both
right, and I've taken it with two commits on top rather than a review round.

Most importantly, we removed the knob. Not only because we believe this should
be the new default (we do), but also because my review showed it did not in
fact provide the safe backwards compatibility it intended: `fire()` attached
the scope and the worker buffered by token regardless of `commit_mode`, so
`write` mode paid the whole cost of grouping — content retained, commits
deferred to after the call — and then committed per file anyway. And a lot of
the complexity here exists only because of the knob. Dropping it took the
replay branch, the buffered `content` and seven generated config surfaces with
it.

What made that safe was your own `test_a_single_file_commit_reads_naturally`:
it pins `write: 1 file`, and under an always-on boundary that is what every
ordinary write would read as. A one-item batch now delegates to
`_stage_and_commit`, so `write: note.md` is back — and with that, the two
boundaries differ only in the bulk case, which is the bug.

The second commit adds the end-to-end test. Nothing here or in `test_git.py`
drove a real tool call, so the contextvar wiring the whole design rests on was
never actually exercised — which is also why all 33 commit-subject assertions
stayed green through this change.

Description updated, since we squash everything here.

PR #1267, opened 2026-09-01
fix(okf): exclude reserved files from the stats OKF histograms

### LINKED ISSUE #1251: [Bug]: stats.okf counts reserved index.md/log.md as notes, so a conformant vault reports untyped_count while okf_validate reports zero findings
### What happened?

On a freshly seeded, fully conformant OKF vault, `stats` reports six untyped notes while `okf_validate` reports zero notes and zero findings. Both are describing the same six files, which are all reserved `index.md`.

### Observed

Live deployment, `markdown-vault-mcp` 4.1.0rc0 with `fastmcp-pvl-core` 5.1.0, `OKF_MODE=auto`, `OKF_WRITE=true`, `REQUIRED_FIELDS=title`. The vault holds a root `index.md` carrying `okf_version: "0.2"` plus five folder `index.md` files, and no other notes yet.

`stats`:

```json
"okf": {
  "mode": "auto", "declared_version": "0.2", "types": {},
  "untyped_count": 6,
  "status": {"stable": 6},
  "trust": {"unverified": 6},
  "stale_count": 0
}
```

`okf_validate` on the same vault, same moment:

```json
{"mode": "auto", "declared_version": "0.2", "active": true,
 "total_notes": 0, "conformant_notes": 0,
 "missing_type": {"count": 0, "examples": []},
 "root_index_missing": false}
```

`get_index_status` reports `documents_indexed: 6` with `skipped_files: []`, and `list_documents` returns exactly those six `index.md` paths — so the two tools are looking at the same set.

The divergence is in whether reserved files are notes:

- `_audit_file` (`src/markdown_vault_mcp/okf.py:523`) checks `if name in OKF_RESERVED_FILENAMES:` and returns **before** `acc.total += 1`, so reserved files never enter `total_notes`.
- The `stats` OKF aggregation (`src/markdown_vault_mcp/managers/search.py:1411`) iterates `self._fts.list_notes()` with no reserved-filename check, so every indexed document lands in `untyped_count` and in the `status` and `trust` histograms.

Three other call sites do honour the distinction: the ranking downweight (`okf.py:313`, applying `OKF_RESERVED_WEIGHT`), `generate_index` when building listings (`managers/okf_migrate.py:163`), and the convention maintainer (`_okf_convention.py:106`). The stats aggregation is the one that does not.

Two consequences on this vault:

- The first OKF number a client sees is wrong-looking. `stats`'s own description says to call it at the start of a session to understand what the vault contains, so `untyped_count: 6` on a vault with zero conformance findings is the opening impression — and the visible remedy (add `type:` to those files) targets files the spec exempts.
- The `status` and `trust` histograms are entirely navigation scaffolding here: all six entries under `stable` and `unverified` are `index.md` files, not content. The distortion shrinks as real notes arrive but never goes away.

### Expected

The two OKF surfaces agree on what counts as a note, or `stats` says which population it is reporting so the number can be read correctly.

### Reproduction

Seed a vault with a root `index.md` declaring `okf_version` plus one `index.md` per folder and no other notes, then call `stats` and `okf_validate`. Observed on the live deployment described above; the code paths were then read at `origin/main`.

### Suspected cause

`[verified: read search.py:1400-1434 and okf.py:516-530 at origin/main]` The audit filters reserved files before counting and the stats aggregation does not. What I have **not** verified is intent: `stats` may deliberately report over everything indexed rather than over the OKF note population. I have not verified the cause of the divergence — whether it is an oversight or a deliberate difference in scope is not something the code states.

### Open questions / scope

- If the difference is deliberate, the field names carry no hint of it; whether the fix is filtering or naming is undecided here.
- `[unverified]` Whether `log.md` shows the same effect. This vault has no `log.md` yet, so only `index.md` was observed.
- Not about the ranking pipeline, which applies `OKF_RESERVED_WEIGHT` to these files correctly.

pvliesdonk:
Closes #1251

## What

`stats`'s `okf` aggregation iterated every indexed row, so a bundle's reserved
`index.md` / `log.md` files landed in `untyped_count` and in the `status` and
`trust` histograms. `okf_validate` filters them before counting, per spec, so a
fully conformant bundle reported untyped notes on one surface and zero findings
on the other — and the visible remedy (add `type:`) targeted files the spec
exempts.

The aggregation now reports over the same note population the audit covers, and
counts the reserved files apart as a new `reserved_count` key so the section
still reconciles with `document_count`.

## Why the divergence existed

`okf_stats` and its "reserved bundle files … are included" docstring landed in
#968; `okf_validate`, which excludes them, landed in #971. The two surfaces were
written one release phase apart and never reconciled — the sentence documented
the behaviour but gave no rationale, and `docs/design/okf.md` already said
reserved files are excluded from conformance checks per spec. This diff replaces
that sentence with a positive statement of the population being reported rather
than deleting it.

This makes five of five reserved-file call sites agree (ranking downweight,
audit, index generation, convention maintainer, and now stats).

Those five sites hand-roll the predicate in three spellings, and #1251 is what
happens when one of them is written without it — filed as #1268 (Decay), not
fixed here.

## Scope decisions

- **`fix:`, not `feat!:`** — a miscount correction, not a deliberate semantic
  redefinition of the tool surface. `reserved_count` is additive.
- **No `INDEX_SEMANTICS_VERSION` bump** — this is read-side aggregation over
  unchanged stored rows; nothing about how a note's rows are derived from its
  bytes changed.
- The issue's `[unverified]` open question about `log.md` is settled in the
  fixture: `OKF_RESERVED_FILENAMES` is `("index.md", "log.md")` and the loop
  filtered neither, so the new scaffold vault covers both.

## Tests

- `test_stats_okf_section` — updated expectations, plus `reserved_count`.
- `test_stats_and_validate_agree_on_the_note_population` — pins the issue's
  *Expected* as an invariant on the mixed vault: `types` sum + `untyped_count`
  == `okf_validate`'s `total_notes`, `types` sum == `conformant_notes`, and the
  three counts sum to `document_count`.
- `TestReservedOnlyBundle` — reproduces the issue's vault shape (root
  `index.md` + a folder `index.md` + `log.md`, no other notes): empty
  histograms, `untyped_count: 0`, `reserved_count: 3`, `total_notes: 0`.

All three fail if the `continue` is removed (mutation-checked).

## Docs

Swept the whole mirror class in one commit: the `okf_stats` docstring, the
`stats` tool docstring, `docs/tools/index.md`, and both `docs/design/okf.md`
channels plus `docs/design/design.md`.

## Self-review fd0b1489..HEAD

Charters: rules, diff, comments, history. Coverage: past-PR charter skipped
(no review history on the touched region beyond #968); normative charter n/a
(no schema or RFC-2119 prose changed). One finding, fixed before push:
`docs/design/okf.md` channel 4 read "excluded from conformance checks (§4) and
from the `stats` histograms (channel 2) per spec" — the spec exempts reserved
files from the `type` rule only, it says nothing about a stats surface. Reworded
so "per spec" attaches only to the conformance claim.

Gates: `pytest -x -q` 3968 passed · ruff check/format clean · mypy clean ·
`scripts/structural_gate.sh` 100% (91 diff lines, 0 violations) · Vale clean.

### DIFF
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -1384,7 +1384,9 @@ what is implemented today:
 - **Stats & config surface**: `stats` gains an `okf` section (mode,
   declared version, `type` histogram + untyped count, status/trust
   breakdowns, stale count — aggregated from the same `documents` rows
-  `stats` already reads); `config://vault` reports `okf_mode` /
+  `stats` already reads, minus the reserved files, which are counted
+  apart as `reserved_count` so the histograms cover the same note
+  population `okf_validate` audits); `config://vault` reports `okf_mode` /
   `okf_active` / `okf_declared_version`; the default server instructions
   gain an OKF guidance sentence whenever the mode permits detection
   (same pre-clone rationale as the conventions sentence).
--- a/docs/design/okf.md
+++ b/docs/design/okf.md
@@ -171,7 +171,11 @@ frontmatter blob; not persisted):
    `sources` are surfaced on `read` (full) and counted on search hits
    (`sources_count`), not inlined into result lists.
 2. `stats` gains an `okf` section: declared version, effective mode,
-   `type` histogram, status/trust-tier breakdowns, stale count.
+   `type` histogram, status/trust-tier breakdowns, stale count. The
+   histograms cover the note population — reserved files are excluded
+   from them and counted apart as `reserved_count` (#1251), so the
+   section reconciles with the `okf_validate` audit rather than
+   contradicting it on a conformant bundle.
 3. Default server instructions gain an OKF paragraph (composed in
    `_instructions.py`, wired through the existing DOMAIN-WIRING seam):
    what the tiers mean, that `log.md`/`index.md` conventions apply, and —
@@ -184,7 +188,8 @@ frontmatter blob; not persisted):
    conditionally, rather than gated on the marker file existing.
 4. Reserved files `index.md`/`log.md` remain indexed (they are searchable
    navigation) but are tagged for the ranking layer (§5) and excluded from
-   conformance checks (§4) per spec.
+   conformance checks (§4) per spec. They are not notes, so the `stats`
+   histograms (channel 2) exclude them as well.
 
 **Indexing integration:** OKF scalar fields ride the existing
 `document_tags` machinery — when detection is on, the effective
--- a/docs/tools/index.md
+++ b/docs/tools/index.md
@@ -212,7 +212,7 @@ Get an overview of the vault's size, capabilities, and configuration. Call this 
 }
 ```
 
-On an OKF bundle (see `MARKDOWN_VAULT_MCP_OKF_MODE` in [Configuration](../configuration.md)), the response also carries an `okf` section: the configured mode, the declared spec version, a per-`type` histogram plus an untyped count, `status` and trust-tier breakdowns, and the stale-note count.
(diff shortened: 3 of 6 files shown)


pvliesdonk:
@claude review

PR #1270, opened 2026-09-02
feat(embeddings): embed queries and documents asymmetrically

### LINKED ISSUE #1135: [Feature]: per-call input_type on EmbeddingProvider, so asymmetric models embed queries and documents differently

Retrieval-tuned embedding models distinguish between embedding a **document** (at index time) and embedding a **query** (at search time) — Voyage exposes this as the `input_type` request parameter, and other model families use equivalent mechanisms (instruction prefixes). Embedding both sides identically leaves retrieval quality on the table for these models; the asymmetry is the vendor's own recommended usage.

The provider ABC cannot express this today: `EmbeddingProvider.embed(list[str])` carries no signal about which side is being embedded, so `VectorIndex.search` embeds queries exactly as `IndexManager` embeds documents. `[verified: read `providers.py` (ABC signature) and `vector_index.py` `search` → `self._provider.embed([query])` on main]`

This was deliberately excluded from #950 (noted in #949 as a follow-up requiring an `embed` signature change) — the `voyage` preset ships without it, sending every request untyped.

### Desired outcome

A provider that supports asymmetric embedding can be told, per call, whether it is embedding documents or a query. Providers without the concept ignore the signal, and behavior for all existing configurations stays byte-identical by default.

### Implementation ideas

Non-binding: a keyword parameter with a neutral default on `EmbeddingProvider.embed`, forwarded as `input_type` by the Voyage provider and ignored elsewhere.

### Open questions / scope

- `[unverified]` Whether changing the ABC signature counts as a public-library-interface break under `CLAUDE.md`'s policy: `markdown_vault_mcp.providers.EmbeddingProvider` is importable and could be subclassed downstream, though the root-surface guard does not cover submodule paths. Needs a call before the change is typed `feat` vs `feat!`.
- `[unverified]` Whether enabling `input_type="document"` on the index side changes stored vector values enough to warrant sidecar invalidation for existing Voyage vaults (vectors embedded untyped vs typed are not comparable). If so, the rollout needs the same identity-mismatch treatment `VectorIndex.load` already applies to provider/model changes.
- Out of scope: instruction-prefix asymmetry for local models (fastembed) — different mechanism, only worth an issue if someone observes the need.

### Related

Refs #949, #950.

pvliesdonk:
## Closes / Refs

Closes #1135

## What & why

Voyage's models are retrieval-tuned: the vendor asks callers to say whether
each text is a document or a query, and prepends a different retrieval prompt
to each. The provider ABC could not express that, so `VectorIndex.search`
embedded queries exactly as `IndexManager` embeds documents.

`EmbeddingProvider` now has two embedding doors. `embed` is the document side
(every existing call site was already document-side); `embed_query` is the
search side, and `VectorIndex.search` is its only caller. **Only `embed` stays
abstract** — `embed_query` is concrete and delegates to it — so a provider
whose model draws no distinction, and any subclass written before the split,
keeps embedding both sides identically with nothing to implement.
`VoyageProvider` overrides both, routing through a private `_embed` so the
dimension cache is filled once whichever side embeds first.

`input_type` reaches the wire as `extra_body`, and the shared transport omits
the field entirely unless a caller asks for it, so OpenAI and Ollama requests
stay exactly `{model, input}` — which #1169 pinned with a test because Voyage
answers unimplemented OpenAI fields (`dimensions`, `user`,
`encoding_format="float"`) with HTTP 400. That pin now reads
`{model, input, extra_body}` for Voyage only; two new tests assert the other
two providers still send exactly `{model, input}` on **both** doors.

### The one operator-visible consequence

Typed vectors sit in a different space from the untyped ones an existing
Voyage vault holds. So providers also report a `provider_variant` token
(`""` by default, `"input_type"` for Voyage) that joins provider/model as a
third identity component in the vector sidecar's `index_metadata`. An absent
key reads back as `""`, so every vault on the other three providers keeps
loading warm; an existing **Voyage** vault fails the identity check once on
first start after upgrade and takes the same `build_embeddings(force=True)`
self-heal a model change takes — one automatic re-embed pass, documented in
the embeddings guide.

I traced that path rather than assuming it: both `load_or_self_heal` callers
route a `VectorIndexCompatibilityError` to `BuildEmbeddings(force=True)`
(`managers/embeddings.py:355`, `indexing/coordinator.py:345`), and `force=True`
constructs a fresh empty `VectorIndex` before the cold build
(`managers/embeddings.py:503`). It does **not** take the convergence path,
which diffs on text and would have left the old untyped vectors in place —
i.e. the feature would have silently not taken for exactly the vaults it is for.

While there, the load-time identity comparison moved into
`_check_persisted_identity`; a third inline comparison in `VectorIndex.load`
was the growth #1169's own notes flagged as the `PLR0911` shape.

### Resolving the issue's open questions

1. **Is this a public-library-interface break?** No. The abstract set is
   unchanged, so a downstream `EmbeddingProvider` subclass stays instantiable
   and inherits symmetric behaviour; `tests/test_providers.py` pins that
   (`embed_query` / `provider_variant` are not in `__abstractmethods__`, and a
   subclass implementing only `embed` embeds both sides through it). Typed
   `feat`, not `feat!`. `EmbeddingProvider` is not in
   `tests/public_import_surface.txt`, so the root-surface guard is untouched.
   The honest caveat: a **duck-typed** provider — an object passed to
   `VectorIndex` that does not subclass the ABC — now needs `embed_query` and
   `provider_variant`. Three in-repo test doubles were exactly that shape and
   are converted to ABC subclasses in this PR, which is also what makes them
   the working proof of the compatibility claim above.
2. **Does the index side warrant sidecar invalidation?** Yes, and it is
   implemented as `provider_variant` above. Voyage's docs say vectors embedded
   with and without `input_type` "remain compatible", but a vault that
   re-embeds only changed notes would hold two prompt regimes indefinitely and
   rank them against each other, so the identity check takes the same posture
   as a model flip.

### Not `INDEX_SEMANTICS_VERSION`

`AGENTS.md` scopes that constant to how a note's stored **rows** derive from
its bytes. FTS rows are untouched here; the vector sidecar carries its own
identity mechanism, which is what this extends.

- **Instruction-prefix asymmetry for local models (fastembed).** A different
  mechanism, explicitly out of scope in #1135; no issue filed, per the issue's
  own "only worth an issue if someone observes the need".
- **No configuration switch.** Typed embedding is how the `voyage` provider
  embeds, not a knob — a flag would only make the one-time re-embed optional
  at the price of a config surface nobody would flip. No new env var, so no
  wizard, `server.json`, or `config-contract` wiring.
- **No `_EMBEDDING_BATCH_SIZE` tuning** for remote providers — still the other
  follow-up named in #949/#950, still unfiled.
- **No live Voyage call.** The `openai` SDK is faked in-process throughout, so
  the tests pin that `input_type` reaches `embeddings.create` and that
  `extra_body` merges into the JSON body (verified against the installed SDK,
  `openai/_base_client.py:531`). That Voyage's server honours the field is a
  docs citation (docs.voyageai.com: accepted values `null` / `query` /
  `document`), not something CI can prove.

## Local review

- [x] No commit carries `!`; see open question 1 above for why this is additive.

## Docs impact

- [ ] `README.md` — no env var, tool, or flag changed
- [x] `docs/` site pages — `guides/embeddings.md` (the asymmetry and the
      one-time re-embed), `api/providers.md` (the two doors)
- [x] `docs/design/` — `design.md` providers section: the two doors, why this
      was additive, the variant token, and why `INDEX_SEMANTICS_VERSION` is
      not the mechanism
- [x] Inline docstrings

## Verification

- `uv run pytest -q` — 3987 passed, 3 skipped
- `uv run ruff check .` / `ruff format --check .` — clean
- `uv run mypy src/ tests/` — clean (218 files)
- `diff-cover` vs `origin/main` — 100% patch coverage (25 changed lines, 0 missing)
- `scripts/structural_gate.sh` — 100%, 0 violations over 546 lines
- `vale` on both changed site pages — clean
- Mutation-checked: reverting `search` to `embed` fails
  `test_search_embeds_the_query_through_embed_query`; dropping
  `provider_variant` from the save payload fails all five
  `TestProviderVariantIdentity` tests.

---

### DIFF
--- a/docs/api/providers.md
+++ b/docs/api/providers.md
@@ -34,6 +34,24 @@ For OpenAI-compatible APIs, set `OPENAI_BASE_URL` and
 `MARKDOWN_VAULT_MCP_VOYAGE_MODEL`. It is never auto-detected: select it
 explicitly.
 
+## Documents and Queries
+
+`EmbeddingProvider` has two embedding doors. `embed()` is the document side,
+used everywhere text is embedded for storage; `embed_query()` is the search
+side. Only `embed()` is abstract. `embed_query()` defaults to it, so a
+provider whose model draws no query/document distinction writes nothing
+extra and embeds both sides identically.
+
+```python
+vectors = provider.embed(["a stored note"])       # index side
+query_vector = provider.embed_query(["a search"])  # search side
+```
+
+`VoyageProvider` overrides both, sending Voyage's `input_type` parameter so
+the vendor prepends its document or query retrieval prompt. Because that
+changes the embedding space, it also reports a non-empty `provider_variant`,
+which the vector sidecar records next to the provider and model names.
+
 ## API Reference
 
 <!-- vale off -->
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -2699,11 +2699,35 @@ sidecar records ``voyage`` as its provider identity. Unlike the other three,
 for an unrelated tool must not silently take over an index and force a
 re-embed.
 
+**Asymmetric embedding (#1135)**: retrieval-tuned models treat a stored
+document and a query against it differently, so the ABC has two embedding
+doors — ``embed`` (documents, the indexing side) and ``embed_query`` (the
+search side). Only ``embed`` is abstract; ``embed_query`` defaults to it, so a
+provider whose model draws no distinction — and any subclass written before
+the split — stays symmetric with nothing to implement, which is why this was
+an additive ``feat`` rather than an ABC break. ``VoyageProvider`` overrides
+both, sending Voyage's ``input_type`` (``"document"`` / ``"query"``) as
+``extra_body`` through the shared transport, which omits the field entirely
+unless a caller asks for it — OpenAI and Ollama answer unknown request fields
+with the same HTTP 400 as the three above. ``VectorIndex.search`` is the only
+query-side call site; every other embedding call is document-side.
+
+Typed vectors sit in a different space from the untyped ones a pre-#1135
+Voyage vault holds, so the provider also reports a ``provider_variant``
+token (``""`` by default, ``"input_type"`` for Voyage) that joins
+``provider``/``model`` as a third identity component in the vector sidecar's
+``index_metadata``. An absent key reads back as ``""``, so vaults on the other
+three providers keep loading warm while an existing Voyage vault fails the
+check once and takes the same ``build_embeddings(force=True)`` self-heal a
+model change takes. This is the vector sidecar's own identity mechanism, not
+the FTS one: ``INDEX_SEMANTICS_VERSION`` governs how a note's stored *rows*
+derive from its bytes, and those are untouched here.
+
 **Acceptance bar for named provider presets (#1134)**: a named
 ``EMBEDDING_PROVIDER`` value is reserved for a vendor whose API has behavior the
 generic OpenAI-compatible transport cannot express. Voyage clears it on two
 counts — the strict request-shape rejections above, and the query/document
-``input_type`` asymmetry that needs a per-call ABC parameter (#1135); Ollama
+``input_type`` asymmetry now implemented in ``embed``/``embed_query`` (#1135); Ollama
 clears it as a local runtime with no API key, a CPU-only mode, and native
 context-length and reachability probes. Every other OpenAI-compatible endpoint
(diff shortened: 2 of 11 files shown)


pvliesdonk:
@codex review
@claude review

PR #1274, opened 2026-09-02
fix(git): scope the auto-commit's no-diff check to the operation's own paths

### LINKED ISSUE #1249: [Bug]: an auto-commit that stages nothing still commits the operator's staged work
## What happens

`_stage_and_commit` (`src/markdown_vault_mcp/git/strategy.py`) decides whether to commit with a **repository-wide** check:

```python
check_result = subprocess.run(
    ["git", "-C", root, "diff", "--cached", "--quiet"],
    capture_output=True,
)
if check_result.returncode == 0:
    return   # nothing staged, skip
```

That question is "does the index differ from HEAD anywhere", not "did *this operation* stage anything". So whenever a write operation's own `git add` runs but stages nothing, and the operator happens to have unrelated content staged in the index, the check sees their diff and commits it — under this operation's message (`write: notes/alpha.md`) — and the deferred push ships it.

The clearest trigger is writing byte-identical content: the `git add` succeeds, stages no change, and the commit that follows carries only the operator's work. `delete` and `rename` have the same shape.

[verified: read `_stage_and_commit` at 47bb94f; the check sits between the per-operation staging branch and the commit, and nothing between them is scoped to the operation's paths]

## Why it compounds

This is the same class of leak #894 was opened to fix — the rename path staging with a pathspec-less `git add -u` and sweeping in a concurrent edit. #894 scoped the *staging*; the *decision to commit* stayed repository-wide, so the hole is narrower but still open.

It also silently defeats the operator's intent in the way that is hardest to notice: their staged-but-uncommitted work is deliberately uncommitted, and it disappears into a commit attributed to a vault write they did not make.

## Where

`src/markdown_vault_mcp/git/strategy.py`, `_stage_and_commit`, the `git diff --cached --quiet` block after the staging branch.

## Suggested direction

Scope the check to the paths the operation touched — `git diff --cached --quiet -- <pathspec>` — using the same pathspec the staging branch built. The `rename` branch already knows both sides; `write`/`delete` know their single path.

The `old_path is None` legacy branch stages repository-wide (`git add -u`) and has no meaningful pathspec, so it would keep the current behaviour, which is consistent: that branch is imprecise by construction.

## Notes

Found while addressing review feedback on #1244. That PR fixed the adjacent case where `_stage_rename` stages *nothing at all* (both sides gitignored) by propagating the no-op to the caller, which closes the trigger it introduced. This issue is the residual, pre-existing case where the `git add` does run. Deliberately left out of #1244: fixing it touches all three staging branches, well beyond that PR's scope.

[unverified]: whether any deployment has hit this. It needs the operator to keep staged-but-uncommitted work in a vault the server is also writing to, which may be rare in practice.

pvliesdonk:
## Closes / Refs

Closes #1249

## What & why

`git diff --cached --quiet` without a pathspec answers "does the index differ
from `HEAD` anywhere", which is not the question a commit decision asks.
Whenever a write's own `git add` ran but recorded nothing — writing
byte-identical content is the clearest way in — the check saw an operator's
deliberately-uncommitted staged work and committed it under the write's
message, and the deferred push shipped it.

`_stage_one` now returns **the pathspec it staged** rather than a boolean, and
`_index_matches_head` asks the question scoped to it. Returning what staging
actually did, rather than recomputing a pathspec from the operation, keeps the
check exact: a rename that skipped an ignored or untracked side never names it,
so a force-added ignored file the operator happened to stage cannot leak in
through the check either.

Two conventions carry the tri-state, both spelled out in the docstrings:

- `None` — nothing was staged, so there is no pathspec to ask about and the
  caller must not commit at all. This is #1244's early return, unchanged.
- an **empty** pathspec — git's own spelling of "the whole repository",
  returned only by the unscoped `old_path is None` rename branch, whose
  staging really is repository-wide. #1249 says to leave that branch's
  behaviour alone, and this does.

**Both check sites are fixed, not just the one #1249 names.** The issue was
written against `_stage_and_commit` before #1265 landed; #1265 added a second
repository-wide check in `_stage_and_commit_batch`. A batch unions its items'
pathspecs, and widens back to the repository when any item reports the empty
one — otherwise an unscoped rename would drop out of the check entirely and
lose the commit it needs.

- **Does not scope `git commit` itself:** #1273. `_commit_staged` invokes
  `git commit` with no pathspec, so a write that *does* have a diff of its own
  still carries any unrelated staged work into its commit. That is a different
  case from the one #1249 reports — the operation stages something there — and
  the substitute (`git commit -- <pathspec>`) reads the named paths from the
  working tree rather than the index and is refused during a merge, both of
  which touch the pull/rebase path. Filed with a verified repro transcript and
  milestoned v4.2.
- **Does not change the unscoped legacy rename branch.** Its staging is
  imprecise by construction (#894 is the issue that scoped the rest), so its
  check stays repository-wide to match, which is what #1249 asks for.
- **Adds no configuration and changes no tool surface.**

## Local review

- [x] No commit carries `!`. `_stage_rename` and `_stage_one` are private and
      not re-exported; `_stage_and_commit`, which `git/__init__.py` does
      re-export for the historic import surface, keeps its signature.

## Docs impact

- [x] `docs/design/` — "Write + Git Integration" gains the scoping paragraph
      and states the residual `git commit` reach with its issue number
- [x] Inline docstrings — `_stage_rename`, `_stage_one`,
      `_stage_and_commit_batch`, plus the new `_index_matches_head`

## Test status

4,011 pass, 18 skip. Structural gate 100% (0 violations); `mypy` clean; patch

Every new test was mutation-checked against the code it pins:

- forcing `_index_matches_head` back to the pathspec-less form fails
  `test_a_byte_identical_write_leaves_staged_work_alone` and
  `test_a_batch_that_stages_nothing_leaves_staged_work_alone`;
- dropping the batch's widening (`_index_matches_head(root, scope)`) fails
  `test_an_unscoped_rename_widens_the_batch_check_back_to_the_repository`,
  which is the regression that widening exists to prevent.

## Self-review c094ca5..3c1b6cb

Charters: rules, diff, comments, history. Conformance charter: no schema,
wire format, or RFC-2119 prose in this diff. Past-PRs charter skipped — the
touched region's review history is #1265, merged hours ago and authored here.

- `src/markdown_vault_mcp/git/strategy.py:1813` — important/verified — the new
  `_index_matches_head` docstring offered "a delete of an already-deleted
  path" as a second way to reach the check; that path never reaches it.
  Evidence: `git add -u -- a.md` on a path already deleted *and committed*
  exits 128, `fatal: pathspec 'a.md' did not match any files`, so the
  operation raises in staging. Fix: cite only the byte-identical write, which
  is reproduced in a test. Resolved in `15f2bb8d`.
- `src/markdown_vault_mcp/git/strategy.py:1929` — minor/verified —
  `_stage_and_commit_batch`'s docstring said "Only the commit is shared",
  which after this change also omits the shared no-diff check.
  Fix: name the union explicitly. Resolved in `3c1b6cbf`.

Nothing else survived refutation.

---

### DIFF
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -2467,10 +2467,30 @@ rule covers is still staged, with `git add -u` rather than `-A`: its deletion is
 real, dropping it would leave git serving content the vault no longer has there,
 and `-u` acts on tracked files only, so the exclude rules never apply to it.
 And when the filter leaves nothing at all to stage, `_stage_rename` reports that
-to its caller, which skips the commit rather than falling through — the
-`git diff --cached --quiet` check below it is repository-wide, so an operator's
-own staged work would otherwise be committed under the rename's message and
-pushed, the very leak this section exists to prevent.
+to its caller, which skips the commit rather than falling through — there is no
+pathspec to ask a diff check about when nothing was staged, and the historic
+repository-wide check would have committed an operator's own staged work under
+the rename's message and pushed it, the very leak this section exists to
+prevent.
+
+**Scoping the commit decision, not just the staging (#1249).** Staging being
+scoped was only half of it: `git diff --cached --quiet` without a pathspec
+answers "does the index differ from `HEAD` anywhere", which is not the question
+a commit decision asks. Whenever an operation's own `git add` ran but recorded
+nothing — a byte-identical write is the clearest case — that check saw an
+operator's deliberately-uncommitted work and committed it. `_stage_one`
+therefore returns **the pathspec it staged** rather than a boolean, and
+`_index_matches_head` asks the question scoped to it. The batched path unions
+its items' pathspecs. Two conventions carry the tri-state: `None` means nothing
+was staged and the caller must not commit at all, and an **empty** pathspec is
+git's own spelling of "the whole repository", returned only by the unscoped
+`old_path is None` rename branch whose staging really is repository-wide — a
+batch containing one widens its own check back to match, or a rename that
+reported no pathspec would be dropped from the check and its commit skipped.
+
+The commit itself is still repository-wide: `git commit` takes no pathspec, so
+a write that *does* have a diff of its own carries any unrelated staged work
+along with it (#1273).
 
 **Write identity: the `Principal` value (#1160, fixes #1218).** "Who is
 acting" is resolved **once, at the MCP tool edge**, into a frozen
--- a/src/markdown_vault_mcp/git/strategy.py
+++ b/src/markdown_vault_mcp/git/strategy.py
@@ -1810,7 +1810,7 @@ def _index_matches_head(root: str, pathspec: Sequence[str]) -> bool:
     The question a commit decision has to ask is "did this operation stage
     anything", not "does the index differ from HEAD anywhere". Asking the
     repository-wide form let an operation that staged nothing of its own —
-    a byte-identical write, a delete of an already-deleted path — see the
+    writing byte-identical content is the clearest way in — see the
     operator's deliberately-uncommitted staged work and commit it under the
     operation's message.
 
--- a/src/markdown_vault_mcp/git/strategy.py
+++ b/src/markdown_vault_mcp/git/strategy.py
@@ -1673,7 +1673,7 @@ def _is_ignored(root: str, path: Path) -> bool:
     return result.returncode == 0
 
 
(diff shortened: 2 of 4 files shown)


pvliesdonk:
@claude review

PR #1275, opened 2026-09-02
fix(git): commit the operation's own paths, not the whole index

### LINKED ISSUE #1273: [Bug]: a vault write's auto-commit sweeps in the operator's unrelated staged work
## What happened?

Expected a vault write's auto-commit to contain only what that write touched; a
commit subject `write: vault-note.md` arrived carrying an unrelated
`operator.md` the operator had staged and deliberately not committed.

## Observed

Against a repository whose index held one staged-but-uncommitted file of the
operator's own, a single vault write produced:

```
$ git show --name-only --format=%s HEAD
write: vault-note.md

operator.md
vault-note.md

$ git diff --cached --name-only
(empty)
```

The operator's `operator.md` is in the commit, attributed to a vault write they
did not make, and their index is now empty. With deferred push enabled the
commit ships.

## Expected

The commit contains `vault-note.md` only, and `operator.md` stays staged and
uncommitted — still theirs to commit when they choose.

## Reproduction

Checked at `c094ca57` (`main`), and again with #1249's fix applied on top:

```python
# repo with seed.md committed; operator.md written and `git add`-ed
from markdown_vault_mcp.git import _stage_and_commit
(repo / "vault-note.md").write_text("# vault\n")
_stage_and_commit(repo, repo / "vault-note.md", "write")
```

The batch path (`_stage_and_commit_batch`, since #1264) reaches the same
`_commit_staged` and behaves the same way.

## Suspected cause

`[verified: read `_commit_staged` in `src/markdown_vault_mcp/git/strategy.py`
and reproduced the transcript above]` The `git commit` invocation there is
built with `-m <subject>` and an optional `--author`, and no pathspec, so it
commits the whole index rather than the paths the operation staged.

## Open questions / scope

This is **not** the case #1249 covers. #1249 is about an operation that stages
*nothing* nevertheless committing, and its fix scopes the
`git diff --cached --quiet` decision to the operation's own pathspec. That
decision being right does not narrow the commit that follows it: staging is
scoped (#894, #1238, #1249) while the commit is not.

`[unverified]`: whether `git commit -- <pathspec>` is a safe substitute here.
Partial commits read the named paths from the working tree rather than the
index, and are refused during a merge — both interact with the pull/rebase
path in `strategy.py`. I have not verified the cause of any such interaction.

`[unverified]`: the unscoped legacy rename branch (`old_path is None`) stages
repository-wide and so reports no pathspec at all; it may have no meaningful
scope to commit under.

---

pvliesdonk:
## Closes / Refs

Closes #1273. Follows #1274, which scoped the commit *decision*.

## What & why

Scoping the decision left the act unscoped. `git commit` without a pathspec
commits the **whole index**, so a write that *did* have a diff of its own still
carried an operator's deliberately-uncommitted work into a commit titled after
a vault write they never made, and the deferred push shipped it:

```
$ git show --name-only --format=%s HEAD
write: vault-note.md

operator.md
vault-note.md
```

`_commit_staged` now takes the same pathspec `_index_matches_head` already
receives and names it, making the commit *partial* — every other index entry is
left where the operator put it. The unscoped legacy rename branch reports the
empty pathspec and keeps the whole-index invocation, because `--only` with no
paths is rejected outside `--amend` and that branch has no narrower truth to
offer.

**Two differences from the whole-index form, both accepted deliberately** and
both verified against real repositories before choosing this over a temporary
`GIT_INDEX_FILE`:

- A partial commit records the named paths' **working-tree** content rather
  than what was staged a moment earlier. An external edit landing between the
  `git add` and the `git commit` is committed rather than left dirty — the
  repository ends up describing what is on disk, which is what the vault
  mirrors.
- Git refuses it while a merge is in progress
  (`cannot do a partial commit during a merge`). That is the better failure,
  not a regression: the whole-index form would have swept the operator's entire
  in-progress merge into a `write: <note>` commit. The write is on disk, the
  `CalledProcessError` is logged, `schedule_push()` is skipped, and the
  operator's merge stays theirs. Pinned by
  `test_a_write_during_a_merge_commits_nothing`.

A conflicted **rebase** moves the other way: a partial commit *succeeds* there,
where the whole-index form fails outright on `Committing is not possible
because you have unmerged files`. That state is not hypothetical — the comment
on `_force_pull_rebase_fallback` says the defensive abort exists so "any
per-write commit path" does not trip over a leftover `rebase-merge`, and
`abort_in_progress_rebase` is allowed to fail.

- **Does not touch `conflict.write_conflict_files`' pathspec-less commit.**
  Its `NOTE:` says the whole-index reach is intentional there — it also has to
  capture upstream content staged by the rebase's own merge — and it holds
  `_lock` for the same reason. Narrowing it would break conflict resolution.
- **Does not retry or defer the write refused mid-merge.** The file is on disk
  and staged; the operator's own merge commit carries it. Retrying would only
  hit the same refusal, and unstaging it would fight the operator's index.
- **Adds no configuration and changes no tool surface.**

## Local review

- [x] No commit carries `!`. `_commit_staged` is private and not re-exported;
      `_stage_and_commit`, which `git/__init__.py` re-exports for the historic
      import surface, keeps its signature.

## Docs impact

- [x] `docs/design/` — the "still repository-wide (#1273)" paragraph #1274 left
      behind is replaced with the scoped commit and its two trade-offs
- [x] Inline docstrings — `_commit_staged` and `_stage_and_commit_batch`

## Test status

4,014 pass, 18 skip. Structural gate 100% (0 violations); `mypy` clean; patch

Three new tests, each mutation-checked by dropping `--only` from the command:

- `test_a_real_write_commits_only_its_own_path` — the transcript above,
  inverted into an assertion;
- `test_a_batch_commits_only_the_paths_it_wrote` — the same on the batch path;
- `test_a_write_during_a_merge_commits_nothing` — the behaviour change a
  reviewer will ask about, pinned rather than described.

The existing `#1238` / `#1244` rename tests are the arbiter for whether
`--only` chokes on an ignored or tracked-but-ignored pathspec; they pass
unchanged.

## Self-review 5d1d8139..e99687a8

Charters: rules, diff, comments, history. Conformance charter: no schema, wire
format, or RFC-2119 prose in this diff. Past-PRs charter: the touched region's
only review history is #1274, authored here hours ago and merged clean.

No findings survived refutation. Two candidates struck:

- `--author` and `--only` in one `git commit` argv — refuted: git parses
  options before `--`, and the six identity tests in `tests/test_git.py` that
  assert on `--author` now run through the scoped path unchanged.
- `conflict.py:461` pathspec-less commit inconsistent with the new rule —
  struck as pre-existing on untouched lines, and its `NOTE:` documents the
  reach as intentional. Recorded above instead.

---

### DIFF
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -2488,9 +2488,30 @@ git's own spelling of "the whole repository", returned only by the unscoped
 batch containing one widens its own check back to match, or a rename that
 reported no pathspec would be dropped from the check and its commit skipped.
 
-The commit itself is still repository-wide: `git commit` takes no pathspec, so
-a write that *does* have a diff of its own carries any unrelated staged work
-along with it (#1273).
+**Scoping the commit itself (#1273).** Scoping the decision left the act
+unscoped: `git commit` without a pathspec commits the **whole index**, so a
+write that *did* have a diff of its own still carried an operator's unrelated
+staged work into a commit titled after a vault write they never made.
+`_commit_staged` therefore takes the same pathspec and names it, making the
+commit *partial* — every other index entry is left where the operator put it.
+The unscoped legacy rename branch reports the empty pathspec and keeps the
+whole-index invocation, because `--only` with no paths is rejected outside
+`--amend` and that branch has no narrower truth to offer.
+
+A partial commit differs from the whole-index form in two ways, both accepted
+deliberately. It records the named paths' **working-tree** content rather than
+what was staged a moment earlier, so an external edit landing between the
+`git add` and the `git commit` is committed rather than left dirty — the
+repository ends up describing what is on disk, which is what the vault mirrors.
+And git refuses it outright while a merge is in progress
+(`cannot do a partial commit during a merge`), which is the better failure: the
+whole-index form would have swept the operator's entire in-progress merge into
+a `write: <note>` commit. The write is on disk, the `CalledProcessError` is
+logged, and the operator's own merge commit carries the staged file. A
+conflicted *rebase* — the state `resolve_conflicts_safely`'s defensive abort
+exists to clean up, and cannot always — is unaffected in the other direction: a
+partial commit succeeds there, where the whole-index form fails outright on
+"unmerged files".
 
 **Write identity: the `Principal` value (#1160, fixes #1218).** "Who is
 acting" is resolved **once, at the MCP tool edge**, into a frozen
--- a/src/markdown_vault_mcp/git/strategy.py
+++ b/src/markdown_vault_mcp/git/strategy.py
@@ -1890,6 +1890,7 @@ def _stage_and_commit(
         root,
         f"{operation}: {rel_path}",
         _CommitIdentity(commit_name, commit_email, author_name, author_email),
+        scope,
     )
 
     logger.info("Git: committed %s (%s)", rel_path, operation)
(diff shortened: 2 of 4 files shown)


pvliesdonk:
@claude review

PR #1276, opened 2026-09-02
perf(git): ask the exclude rules about both sides of a rename at once

### LINKED ISSUE #1250: Batch the two check-ignore probes in _stage_rename into one call
## What

`_stage_rename` (`src/markdown_vault_mcp/git/strategy.py`) now runs up to four `git` subprocesses per renamed file: `_is_tracked` for the old path, `_is_ignored` for each of the two paths, then the `add` itself (two adds when a tracked-but-ignored deletion has to be staged separately). Before #1244 it was one.

`git check-ignore` accepts multiple pathspecs in a single invocation and reports per-path, so the two `_is_ignored` calls collapse into one.

## Where

`src/markdown_vault_mcp/git/strategy.py` — `_is_ignored` and its two call sites in `_stage_rename`.

## Why it is worth doing

`move_folder` fires one rename callback per moved file, so the subprocess count scales linearly with subtree size. It is proportional rather than quadratic, and each call is cheap, but a large vault reorganization pays it on every file, and process spawn dominates the actual work here.

Noted as non-blocking during review of #1244 and deliberately not fixed there — that PR was a correctness fix and batching would have made the staging logic harder to review against the four tracked/ignored quadrants it had to get right.

## Suggested direction

Give the probe a plural shape — `_ignored_paths(root, paths) -> set[str]` — running `git check-ignore --no-index -z --stdin` (or passing both pathspecs as arguments) and parsing the reported subset. Note that exit code 1 means "none ignored" and is not an error, and that `--no-index` is load-bearing: without it a tracked file under an ignored directory reads as "not ignored" while `git add` still refuses it. `tests/test_git.py::TestRenameStagingSkipsIgnoredPaths` pins that behaviour and must keep passing unedited.

pvliesdonk:
## Closes / Refs

Closes #1250

## What & why

`_stage_rename` probed `git check-ignore` once per path. `move_folder` fires a
rename callback per moved file, so the probe count scales with the size of the
moved subtree, and spawning the process costs more than matching the rules
does.

`_ignored_paths` replaces `_is_ignored` with a plural shape: one invocation,
the excluded subset back. Per renamed file that takes the worst case from four
`git` processes to three, and — because the old-path exclude probe sat behind
an `elif` on `_is_tracked` — the best case from three to three, never worse.

**Why `--stdin -z` rather than passing both paths as arguments.** `-z` is
accepted *only* with `--stdin`; the argument form exits with
`fatal: -z only makes sense with --stdin`. Without `-z` the reply is
newline-delimited, which is ambiguous for a filename containing a newline. Git
echoes each ignored path back **verbatim**, so the strings returned compare
equal to the strings that were sent — verified for absolute and relative input
alike before relying on it.

**Exit codes.** 1 means "none of them" and is the ordinary negative answer, not
a failure. Above that is a failure, and it answers "none excluded" exactly as
the single-path probe did — reporting an exclusion nobody asked for would drop
a real path from the staging pathspec, while reporting none lets the `git add`
that follows fail loudly with git's own stderr.

One behaviour does change, deliberately: that failure is now **warned** rather
than swallowed. The old probe discarded any non-zero return code silently, so
when the paths turned out not to be excluded after all, the `git add` succeeded
and nothing recorded that the filter had answered from a failure rather than
from the exclude rules.

`--no-index` is unchanged and still load-bearing — the index-aware default
calls a tracked file under an ignored directory "not ignored" while `git add`
still refuses it. It is now pinned at the probe itself, not only through the
staging that trusts it.

- **Does not batch `_is_tracked`.** #1250 scopes itself to the two exclude
  probes; `_is_tracked` is asked once per rename already, and folding it in
  would mean reconciling `ls-files` output with `check-ignore` output in one
  place for no further saving.
- **Does not change what gets staged or committed.** The four
  tracked/ignored quadrants `TestRenameStagingSkipsIgnoredPaths` pins are
  untouched, and that class passes unedited — which #1250 asks for explicitly.
- **Does not batch across renames.** The callback boundary is one file; the
  saving here is within a single rename, not across a `move_folder`.

## Local review

- [x] No commit carries `!`. `_is_ignored` and `_ignored_paths` are private and
      not re-exported; the public import surface is unchanged.

`perf:` is the honest type here — no user-visible behaviour changes, and per
`AGENTS.md` it cuts no release and reaches no changelog entry. Retitle to
`fix:` if you read the new warning as worth a patch release on its own.

## Docs impact

- [x] `docs/design/` — the exclude-probe paragraph gains the batching, the
      `--stdin -z` reasoning, and the exit-code convention
- [x] Inline docstrings — `_ignored_paths` carries the rationale `_is_ignored`
      had, plus what is new

## Test status

4,020 pass, 18 skip. Structural gate 100% (0 violations); `mypy` clean; patch

Six new tests in two classes. The two that pin the saving count
`check-ignore` spawns through a real `_stage_and_commit` rename, and are
mutation-checked: restoring two separate `_ignored_paths` calls fails both.
The other four pin the probe itself — the mixed batch, the tracked-under-an-
ignored-directory case that `--no-index` exists for, the empty batch that asks
git nothing, and the failing probe that warns and reports no exclusions.

## Self-review 77b5b81..3583bf3

Charters: rules, diff, comments, history. Conformance charter: no schema, wire
format, or RFC-2119 prose in this diff. Past-PRs charter: the region's review
history is #1274 and #1275, both authored here today.

One finding, found and fixed during the review:

- `src/markdown_vault_mcp/git/strategy.py:1690` — important/verified — the
  probe failure was logged at `DEBUG`, invisible at the default level.
  Evidence: the `logging-standard` skill's level table assigns "degraded but
  continuing: API errors with fallback" to `WARNING`, and this is exactly that
  — with the aggravating detail that the fallback can succeed silently.
  Fix: raise to `WARNING`. Resolved in `3583bf3a`.

One candidate struck: redacting the probe's stderr through `redact()`, as the
network-facing helpers do. Refuted — `check-ignore` contacts no remote and
takes no credential, `redact(text, None)` is a documented no-op, and calling it
would imply a leak path that does not exist here.

---

### DIFF
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -2462,6 +2462,15 @@ file "not ignored", but `git add` still refuses to name a tracked file living
 under an ignored directory, so only `--no-index` predicts the rule the filter
 exists to satisfy.
 
+Both sides are asked in **one** call, not one each (#1250). `move_folder` fires
+a rename callback per moved file, so the probe count scales with the size of
+the moved subtree, and spawning the process costs more than matching the rules
+does. The paths go over `--stdin` with `-z`: `-z` is accepted only with
+`--stdin`, and the newline-delimited reply it would otherwise produce is
+ambiguous for a filename containing a newline. Git answers with the excluded
+subset, echoing each path back verbatim, and exit code 1 means "none of them"
+rather than a failure.
+
 Two consequences of that filter are load-bearing. A **tracked** path an exclude
 rule covers is still staged, with `git add -u` rather than `-A`: its deletion is
 real, dropping it would leave git serving content the vault no longer has there,
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -2468,8 +2468,10 @@ the moved subtree, and spawning the process costs more than matching the rules
 does. The paths go over `--stdin` with `-z`: `-z` is accepted only with
 `--stdin`, and the newline-delimited reply it would otherwise produce is
 ambiguous for a filename containing a newline. Git answers with the excluded
-subset, echoing each path back verbatim, and exit code 1 means "none of them"
-rather than a failure.
+subset, echoing each path back verbatim. It defines exactly two exit codes —
+0 for "some of them" and 1 for "none" — so 1 is an ordinary negative reply and
+not a failure, while every other code is one, the negative code a signal
+produces included.
 
 Two consequences of that filter are load-bearing. A **tracked** path an exclude
 rule covers is still staged, with `git add -u` rather than `-A`: its deletion is
--- a/src/markdown_vault_mcp/git/strategy.py
+++ b/src/markdown_vault_mcp/git/strategy.py
@@ -1670,6 +1670,12 @@ def _ignored_paths(root: str, paths: Sequence[Path]) -> set[str]:
     **verbatim**, so the strings returned here compare equal to the strings
     that were sent.
 
+    Git defines exactly two answers: 0 for "some of them" and 1 for "none",
+    so 1 is an ordinary negative reply rather than a failure. Every other
+    code is a failure and is answered as "none excluded" — the negative code
+    POSIX reports for a signal included, which is why the guard tests
+    membership rather than ordering.
+
     Args:
         root: Git repository root, as a string.
         paths: Absolute paths to probe.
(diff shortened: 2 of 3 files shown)


pvliesdonk:
@claude review

pvliesdonk:
## Self-review 77b5b81..6c562cd3 (round 2)

Charters: rules, diff, comments, history, past-PRs. Conformance charter: no
schema, wire format, or RFC-2119 prose in this diff. Coverage: full.

Two findings, both resolved.

- `src/markdown_vault_mcp/git/strategy.py:1689` — important/verified — the
  probe's failure guard was `result.returncode > 1`, which a **negative** code
  does not satisfy, so a signalled `check-ignore` fell through to parsing its
  empty reply as "none excluded" and lost the warning. Caught by @claude's
  review above; confirmed by mutating the guard back, which fails the new
  `test_a_signalled_probe_is_a_failure_too`.
  Fix: `result.returncode not in (0, 1)` — git defines exactly two codes, and
  membership says so where ordering only implied it. Resolved in `a1d545d2`.

  Swept the class rather than the instance: `returncode` is compared in five
  other places across `src/`, all `== 0` / `!= 0`, which is already
  signal-safe. This guard was the only ordered comparison, so the class has
  one member.

- `src/markdown_vault_mcp/git/strategy.py:1690` — minor/verified — the fix
  left the same rationale written twice, once in the docstring paragraph and
  once in a ten-line inline comment that re-derived it. Two copies of a
  rationale drift.
  Fix: the docstring keeps the contract ("why membership, not ordering"); the
  comment keeps only what a reader at the branch needs (why `WARNING` and not
  `DEBUG`) and points at it. `docs/design/design.md` gains the same exit-code
  clause, so the three places now say one thing. Resolved in `6c562cd3`.

One candidate struck on refutation: whether the docstring is now over-long at
four rationale paragraphs. Each answers a distinct "why not the obvious
thing" — one call not two, `--no-index`, `--stdin -z`, the exit codes — and
every one of them is a git behaviour that was verified against a real
repository rather than recalled. Compressing them would cost the next reader
the verification.

Re-verified after the fixes: 4,021 pass / 18 skip, structural gate 100%
(0 violations), `mypy` clean, patch coverage 100%.

The `perf:` vs `fix:` title question @claude flagged is still yours — the
`logger.warning` is the only operator-visible difference in the diff.

---

PR #1277, opened 2026-09-02
fix(sync): keep the file watcher when a webhook cannot deliver on this transport

### LINKED ISSUE #1263: [Bug]: a webhook credential on stdio disables the file watcher but mounts no route, leaving no external-change detection
### What happened?

A stdio deployment with a webhook credential set and no git pull loop ends up with no external-change detection at all. The credential disables the file watcher, but no webhook route is mounted under stdio, so nothing pulls and nothing rescans. Externally edited notes stay stale until something else triggers a reindex.

### Observed

`should_start_file_watcher` (`src/markdown_vault_mcp/_file_watcher.py:349`) treats a configured webhook as a reason to stand down:

```python
git_active = git_pull_active or webhook_configured
return file_watcher_enabled and not git_active
```

That is correct where the webhook can actually deliver. Under stdio it cannot: `register_webhook_routes` (`src/markdown_vault_mcp/_webhooks.py`) returns before mounting anything when `transport == "stdio"`, because there is no HTTP server to receive a POST. The gate never learns that — `Service` (`src/markdown_vault_mcp/domain.py:89`) has no transport at all. `[verified: grepped `transport` in domain.py — no occurrences; Service is constructed no-args by the template's server_lifespan]`

So with `MARKDOWN_VAULT_MCP_GITHUB_WEBHOOK_SECRET` (or either GitLab credential) set, `--transport stdio`, and no `GIT_REPO_URL`:

- no webhook route exists, so no delivery can arrive;
- `git_pull_active` is False, so no pull loop runs;
- the watcher is disabled anyway, because a credential is present.

This is not new. On `main` before #1178 the gate took `github_webhook_secret` and behaved identically; the GitLab credentials inherit it rather than introduce it. `[verified: read the pre-#1178 signature and body of should_start_file_watcher]`

Surfaced by an automated reviewer on PR #1262.

### Expected

The watcher stands down only when something is actually going to drive reindexing. A credential that mounts no route on this transport does not.

### Reproduction

Set `MARKDOWN_VAULT_MCP_GITHUB_WEBHOOK_SECRET`, run with `--transport stdio` and no `GIT_REPO_URL`, then edit a note outside the server. It does not appear in search results. I reasoned this from the three call sites above rather than running it, so the reproduction is unverified.

### Suspected cause

`[unverified]` The gate was written when the webhook was the only non-pull trigger, and the transport check lives at the mounting site rather than in the config, so the two conditions were never expressed together. I have not verified the cause.

### Open questions / scope

- The fix needs the transport where the gate is evaluated, and `Service` does not currently receive it — that plumbing is why this was not folded into #1178.
- `[unverified]` Whether a stdio deployment with a webhook credential set is a configuration worth supporting at all, or whether the better answer is a startup warning that the credential is inert on this transport. Both address the observation; this issue does not pick one.
- Not about the webhook handlers themselves, which are transport-correct.

pvliesdonk:
## Closes / Refs

Closes #1263

## What & why

A webhook credential stood the file watcher down whether or not the webhook
could actually deliver. Under stdio `register_webhook_routes` mounts no route —
there is no HTTP server to receive a POST — so a stdio deployment with a
credential and no `GIT_REPO_URL` had **no external-change detection at all**:
nothing to deliver to, no pull loop, and a watcher that had already stood down.
Externally edited notes stayed invisible until something else triggered a
reindex.

The bug's shape is two sites answering the same question differently: the
mounting site knew the transport, the gate did not. Both now read one
predicate, `webhook_routes_active(config, transport)`, fed the same transport
`make_server` was called with. That transport reaches the gate through
`set_pending_transport`, staged for the template's no-arg `Service()` exactly
the way the config already is (#609) — it is a `make_server` argument rather
than a config field, because the CLI's `--transport` flag shadows any env
value, and like the staged config it survives a ref-counted lifespan re-entry.

#1263 floated two candidate fixes and deliberately did not pick one. **Both
are here**, because they answer different halves: the gate change restores
change detection, and a new `webhook_transport_inert` startup warning says the
credential itself is doing nothing on this transport, instead of
`register_webhook_routes` returning silently. This is the plumbing PR #1262's
review round said was out of scope there.

Intended behaviour change, worth naming: an existing stdio deployment with a
webhook credential now runs the file watcher where it previously ran nothing.

One corner, pre-answered: stdio + managed git + `GIT_PULL_INTERVAL_S=0` +
credentials now runs the watcher while a `git_sync` call can rewrite the tree.
That is exactly what the same deployment without credentials already did — it
is parity with today, not a new hazard, and the mid-checkout scan #558 guards
still keeps the watcher off wherever the pull loop or a live webhook runs.

- Does not give `SyncConfig` a transport field or make the transport
  config-derived: `config.server.transport` exists but the CLI flag shadows it,
  and a gate reading config while the mounting site reads the argument is this
  exact bug class reborn.
- Does not change the webhook handlers, their authentication, or their outcome
  set — those are transport-correct already (#1263 says so explicitly).
- Does not touch the watcher/pull-loop exclusion itself: a live webhook on
  HTTP/SSE still stands the watcher down, unchanged since #558.

## Local review

- [x] No commit carries `!`. Nothing an operator must change to upgrade, and
      `should_start_file_watcher` / `Service` are not part of the public import
      surface (`markdown_vault_mcp/__init__.py` exports neither); `Service`
      gained an optional keyword only.

<details>
<summary>Self-review 889ff0d6..08b225f3</summary>

Charters: rules, diff, comments, history, conformance, past-PRs.
Coverage: full. Conformance charter: no normative content in this change
(`server.json` moves only a generated description string; the version is
untouched, so the three manifests stay in lockstep).

- `src/markdown_vault_mcp/_file_watcher.py:5` — important/verified — the module
  docstring still named the GitHub webhook and said a configured secret alone
  keeps the watcher off, which is the behaviour this diff changes.
  Evidence: "Used when neither the git periodic-pull loop nor the GitHub
  webhook is configured" (module docstring at HEAD~2).
  Fix: point it at `webhook_routes_active` and drop the GitHub-only naming.
  Resolved in dd945e83.
- `src/markdown_vault_mcp/config_sections/sync.py:39` — important/verified —
  `webhook_configured`'s docstring claimed the file-watcher gate asks it
  directly, which stopped being true in this diff.
  Evidence: "so the gate asks this rather than naming a provider" (property
  docstring at HEAD~2).
  Fix: say the gate reads `webhook_routes_active`, which weighs this against
  the transport. Resolved in dd945e83.
- Minors, aggregated: the file-watcher test header's covered-modes list did not
  mention the new stdio/http lifespan cases (08b225f3).

Past-PR charter: #1262's review rounds filed this as P2 and named both
candidate fixes plus the reason it was deferred ("the gate cannot see the
transport: `Service` is constructed no-args"). This PR implements the plumbing
that round described.

Gate: 4031 passed · patch coverage 100% (15 diff lines, diff-cover against
`origin/main`) · structural gate 100% · mypy 220 files · ruff and Vale clean.

</details>

## Docs impact

- [x] `README.md` — regenerated env-table description for `FILE_WATCHER`.
- [x] `docs/` site pages — `configuration.md` (prose + table),
      `guides/git-integration.md` (the "file watcher steps aside" note now says
      which transports, and names the startup warning).
- [x] `docs/design/` — `design.md` gains "A credential is not a delivery
      (#1263)" under the push-webhook section, plus the two prose sites that
      described the old gate; `module-map.md` line for `_file_watcher.py`.
- [x] Inline docstrings — `should_start_file_watcher`, `webhook_routes_active`,
      `register_webhook_routes`, `Service`, `set_pending_transport`,
      `SyncConfig.webhook_configured`, and the `_file_watcher` module docstring.

Generated artifacts (`server.json`, `.env.example`, `examples/*.env`,
`packaging/env.example`, `wizard-spec.json`) come from
`scripts/gen_config_surface.py`; `--check` passes.

---


pvliesdonk:
@claude review

pvliesdonk:
Both minor observations addressed — one was a real defect, so thanks for the pointer.

**Observation 1 — the "three startup warnings" docstring.** Flagged as a maintenance note, but following it turned up an actual inconsistency this PR introduced: `design.md:3893` still read "the gates and **both** startup warnings", written when there were two. Adding `webhook_transport_inert` made a third, and I had updated the count in the docstring without updating its mirror in the design doc — the exact class where a counted enumeration rots silently.

Fixed in eb652db3 by removing the count from both sides rather than incrementing it: the docstring and `design.md` now say "every startup warning about a credential that is set but weak or inert". `grep -rn 'startup warning'` over `src/`, `docs/` and the root Markdown confirms no other counted instance remains.

**Observation 2 — the stdio + managed git + `GIT_PULL_INTERVAL_S=0` parity argument.** No change; we agree, and #558's guards are untouched.

Re-verified narrowly: mypy clean (98 files), `test_webhooks.py` + `test_server.py` 330 passed, Vale clean.

PR #1280, opened 2026-09-02
fix(config): normalize attachment-extension spellings on the way in

### LINKED ISSUE #1239: [Bug]: an ATTACHMENT_EXTENSIONS value written as PDF or .pdf silently matches nothing
## What happened

The attachment allowlist is compared against a path extension that has been lower-cased and dot-stripped, but the configured values are used exactly as written. So an operator who sets `MARKDOWN_VAULT_MCP_ATTACHMENT_EXTENSIONS=PDF,PNG` or `.pdf,.png` — both natural spellings — gets an allowlist that matches nothing, and every attachment is refused with a message listing the extensions they just configured. [verified: read `effective_attachment_extensions` and the comparison sites at 29ee69b; pinned as current behaviour by `test_config_side_extensions_are_not_normalized_today` in `tests/test_utils_content_kind.py`]

## Where

`src/markdown_vault_mcp/utils/content_kind.py` — `effective_attachment_extensions` returns `frozenset(attachment_extensions)` verbatim. The path side is normalized by `artifact_suffix` (`Path(path).suffix.lstrip(".").lower()`), so `"report.pdf"` becomes `"pdf"` and never equals `"PDF"` or `".pdf"`.

The resulting error is actively confusing, because it echoes the configured values back:

```
Extension '.pdf' is not in the attachment allowlist. Allowed: .PDF, .PNG.
```

## Expected

An extension list is normalized on the way in the same way the path side is, so case and a leading dot do not change the meaning. `pdf`, `PDF`, and `.pdf` all name the same file type.

## Reproduction

Set `MARKDOWN_VAULT_MCP_ATTACHMENT_EXTENSIONS=PDF`, then `read` any `.pdf` in the vault.

## Notes

The wildcard `*` is unaffected. Normalizing would widen what is accepted rather than narrow it, so no currently-working configuration would break — but it is a behaviour change to a documented operator surface, which is why #1235 pinned it rather than fixing it in a behaviour-preserving refactor.

The docs may also deserve a look: if the reference states the bare-lowercase form explicitly, this is closer to a validation gap (accept the spelling, or reject it loudly) than to a silent-mismatch bug. [unverified] I have not checked what the configuration reference currently promises.

pvliesdonk:
## Closes / Refs

Closes #1239

## What & why

`effective_attachment_extensions` returned the configured list verbatim while
the path side was lower-cased and dot-stripped by `artifact_suffix`, so an
operator who wrote `MARKDOWN_VAULT_MCP_ATTACHMENT_EXTENSIONS=PDF` or `.pdf` —
both natural spellings — got an allowlist that matched nothing, and a rejection
that echoed their own spelling back at them.

Both sides are now normalized the same way (strip whitespace, drop leading
dots, lower-case), at the one choke point every caller shares — `ArtifactStore`,
the transfer sink, `SearchManager`, `GitQueryManager` — so direct library
callers passing `attachment_extensions=["PDF"]` are covered too. An entry that
normalizes to nothing (`.`) is dropped rather than kept as `""`, which
`artifact_suffix` returns for an extension-less path and which would otherwise
allowlist every `Makefile`-style file. The `*` wildcard is unaffected.

Verified end to end through `resolve_attachment_extensions` → the allowlist:

| configured | effective | `report.pdf` | `Makefile` |
|---|---|---|---|
| `PDF` | `['pdf']` | ✅ | ❌ |
| `.pdf` | `['pdf']` | ✅ | ❌ |
| `PDF, .PNG` | `['pdf', 'png']` | ✅ | ❌ |
| `*` | `['*']` | ✅ | ✅ |
| `.` | `[]` | ❌ | ❌ |

This only widens what is accepted, so no working configuration changes meaning
— not breaking under the `AGENTS.md` policy. The test that pinned the asymmetry
as current behaviour (added deliberately by #1242 as "a visible one-line flip")
is flipped to assert the fix.

- Does not touch the two case/resolution divergences among the `.md` routing
  predicates that `content_kind`'s module docstring preserves — those are a
  separate, deliberate thing tracked at #1240.
- Does not add validation that *rejects* an unusual spelling. #1239 raised
  normalize-vs-reject as an open question; the configuration reference promises
  no rejection, so normalize-on-ingest is the change that matches what is
  documented. Rejecting would break configurations that work today.
- Does not normalize any other list-valued config var (`exclude_patterns`,
  `indexed_frontmatter_fields`, …). Nothing suggests they have the same
  asymmetry, and sweeping them would widen surfaces this issue did not observe.

## Local review

- [x] No commit carries `!` — this widens an operator surface rather than
      breaking one.

<details>
<summary>Self-review 9c0a3b2d..(HEAD) — no surviving findings</summary>

Charters: rules, diff, comments, history, conformance, past-PRs.
Coverage: full.

Checked and cleared:
- **Manifest lockstep** — `server.json` changed only two `description` strings;
  no version line touched (`git diff -- server.json | grep version` empty).
- **`INDEX_SEMANTICS_VERSION`** — no bump needed: the allowlist feeds no stored
  index row (attachments are never indexed; `grep attachment` over `fts_index.py`,
  `indexing/`, `scanner.py` is empty). It is read at query time only.
- **Config contract** — the `help` metadata in `config.py` is the single source;
  `.env.example`, `packaging/env.example`, `examples/*.env`, `wizard-spec.json`,
  `docs/configuration.md` and `server.json` were regenerated with
  `scripts/gen_config_surface.py`, not hand-edited. This var has no mcpb
  install-screen entry.
- **Doc-mirror class** — swept every `without (leading) dot` / `lowercase` claim
  in `src/`, `tests/`, `docs/`, `README.md`. The claims describing the *effective
  frozenset* (`types.py:958`, `utils/__init__.py:151,159`, `search.py:679`) were
  already written as lower-case/no-dot and this change makes them true rather
  than stale; only the three describing what a *caller may write* needed the
  edit (`config.py` help, `vault.py:277`, `git_query.py:44`).
- **Recorded intent** — `git log --follow` on `content_kind.py` shows #1242
  pinning this asymmetry on purpose with the follow-through planned. Not a
  silent revert.
- **Refuted** — non-`str` elements now raise instead of silently mismatching
  (mypy territory, and loud beats silent); `.*` normalizes to the wildcard
  (an extension allowlist has no narrower meaning for `.*`, so that is the only
  sensible reading); per-call frozenset construction is unchanged in cost.

Gates: `pytest -x -q` 4040 passed · ruff check/format clean · mypy clean ·
`structural_gate.sh` 100% (0 violations, 56 lines) · `content_kind.py` at 100%
line and branch coverage.
</details>

## Docs impact

- [ ] `README.md` — does not mention this var
- [x] `docs/` site pages — `docs/configuration.md` (regenerated)
- [x] `docs/design/` — env table entry plus a paragraph recording the
      both-sides-normalized rule under Attachment Support
- [x] Inline docstrings — `effective_attachment_extensions`, `Vault`,
      `GitQueryManager`

---


pvliesdonk:
@claude review
@codex review

PR #1286, opened 2026-09-03
fix(git): accept the object IDs a SHA-256 repository mints

### LINKED ISSUE #1284: [Bug]: a SHA-256 repository's commit IDs are rejected by get_diff, which get_history just returned
## What happened?

Expected `get_diff` to accept a commit SHA that `get_history` just returned;
in a SHA-256 repository it rejects it as invalid.

## Observed

Against `86f3050d`, on a vault whose repository was created with
`git init --object-format=sha256`, `get_history` returns 64-character commit
IDs and `get_diff` refuses them:

```
GitQueryManager(GitWriteStrategy(), repo).get_history("n.md", limit=2)
  -> [(64, '1e88a55cd360'), (64, '7724a33294e3')]     # (len(sha), sha[:12])

.get_diff("n.md", since_sha='7724a332…8678ca')
  -> ValueError: Invalid SHA '7724a33294e3c6d065e94046b18564ad84db048f2b98
     27397cab59b62a8678ca': must be 4-40 lowercase hex digits
```

The repository already carries deliberate SHA-256 support elsewhere:
`tests/test_git.py:4446-4496` (from #683) resolves the empty-tree object
against the repository's hash algorithm rather than the SHA-1 constant.

Four places also state the object-ID width as a fact:
`src/markdown_vault_mcp/types.py:688` and `:713` ("Full 40-character commit
SHA" on `HistoryEntry.sha` / `CommitDiff.sha`),
`src/markdown_vault_mcp/_server_tools/git.py:243`, and
`docs/tools/index.md:933`.

## Expected

A SHA that `get_history` returned is accepted by the tools documented to take
one, and the prose describing a commit SHA holds for every repository this
server can be pointed at.

## Reproduction

```bash
git init --object-format=sha256 vault && cd vault
git config user.email t@t && git config user.name t
printf 'v1\n' > n.md && git add -A && git commit -m c1
printf 'v2\n' > n.md && git commit -am c2
```

```python
from markdown_vault_mcp.git.strategy import GitWriteStrategy
from markdown_vault_mcp.managers.git_query import GitQueryManager

m = GitQueryManager(GitWriteStrategy(), Path("vault"))
h = m.get_history("n.md", limit=2)
m.get_diff("n.md", since_sha=h[-1].sha)   # raises
```

Ran both against `86f3050d` with git 2.x supporting `--object-format=sha256`.

## Suspected cause

`[verified: ran the reproduction above, then read the validator and both call
sites]` The shared SHA check in
`src/markdown_vault_mcp/managers/git_query.py:198` matches
`[0-9a-f]{4,40}`, which no 64-character object ID satisfies.

## Open questions / scope

- Surfaced while reviewing #1137, whose revision-read parameter would flow
  through the same validator; the `get_diff` constraint itself predates that
  work.
- `[unverified]` Whether any surface outside the four listed above narrows an
  object ID to 40 characters. I grepped for `{40}` / "40-character" / "hex
  digits" across `src/`, `docs/` and `tests/` and found only those, plus
  `tests/test_git_interfaces.py:92`, which asserts 40 hex on a SHA-1 fixture.

---

pvliesdonk:
## Closes / Refs

Closes #1284. Refs #1137.

## What & why

On a vault whose repository was created with `git init --object-format=sha256`,
`get_history` returns 64-character commit IDs and `get_diff` rejected every one
of them — the shared validator bounded a SHA at 40 hex digits, so the only SHAs
the vault could hand a caller were SHAs it refused to take back. The bound is
now 64, and the four places that stated the 40-digit width as a fact say what is
actually true of the repository in front of them.

The width was never the check's purpose: it exists to keep caller-supplied text
out of the git argv, and git reports an unknown ref itself. That purpose is now
written down at the constant and pinned by a test — `HEAD~1`, `@{upstream}`,
`--all`, `-x`, an uppercase digit and a branch name are all still rejected
before the strategy is called. Widening a bound is exactly the change that can
quietly turn a shape check into a length check; it did not.

**This is a precondition for #1137, not only a cleanup.** The revision parameter
proposed there flows through this same validator, so the feature would inherit
both the rejection and the argv-safety contract. It lands separately because the
`get_diff` rejection predates that work and is a defect on its own.

### Tests

- `TestSha256Repository` drives a **real** `--object-format=sha256` repository —
  `get_history` → `get_diff` round trip — because the defect was an assumption
  about what git returns, which a fake cannot contradict. It uses the existing
  skip pattern (probe `init.returncode`, skip with git's own stderr) from
  `test_empty_tree_resolution_matches_repo_object_format` (#683), so a git
  without SHA-256 support skips rather than fails.
- Mutation-checked: with the bound back at 40, both new acceptance tests fail.

- **No widening of the shape.** 41–63 hex digits is not an object ID in either
  algorithm, and the pattern still admits it. The check is an argv guard, not a
  proof the object exists; a `{40}|{64}` alternation would buy a stricter error
  message for input git rejects anyway.
- **Does not touch the "at least 4 hex digits" prose** in the four caller-facing
  docstrings and the `since_sha` docs row. Those state a lower bound and remain
  true; only the "Full 40-character commit SHA" claims were false.
- **`short_sha` still documented as 7 characters** — verified, not assumed: `%h`
  returns 7 characters in a SHA-256 repository too.
- **No new issue filed for `get_history`'s own SHA-256 behaviour** — there is no
  defect there; it already returns whatever width git mints.

## Local review

      Report below.
- [x] Nothing here carries `!`: no operator surface, no public library
      interface, and the accepted input set only grows.

## Docs impact

- [ ] `README.md` — states no object-ID width; nothing to correct.
- [x] `docs/` site pages — `docs/tools/index.md`: the `get_history` response
      table's `sha` row.
- [x] `docs/design/` — the `get_file_diff` bullet naming the validator pattern.
- [x] Inline docstrings — `HistoryEntry.sha`, `CommitDiff.sha`, the
      `get_history` tool's returns block, and the new constant's rationale.

## Self-review 86f3050d..9d97093b

Charters: rules, diff, comments, history, conformance.
Coverage: full except the optional past-PR-reviews charter, which was skipped.

No blocking or important findings survived refutation. Four candidates were
raised and struck:

- The widened pattern admits 41–63 hex digits — struck: the constant documents
  the check as an argv guard, and `git/query.py` already translates an unknown
  ref's `CalledProcessError` into `ValueError`.
- `short_sha: Abbreviated 7-character SHA` sits next to a changed line and could
  have rotted in the very repository shape this diff is about — struck by
  execution: `git log --format=%h` returns 7 characters in a SHA-256 repository.
- The 40-digit bound might have been a deliberate earlier fix being reverted —
  struck: `git log -S` traces it to #333's original "SHA hex validation" and
  #612's verbatim move, neither of which ties it to SHA-1.
- `docs/tools/index.md`'s `since_sha` row does not name the new upper bound —
  struck as a nit; the sentence states a lower bound and stays true.

No normative content in this change (the MCP input schema for `get_diff` is
unchanged).

---


pvliesdonk:
@claude review
@codex review

PR #1288, opened 2026-09-03
feat(git): read a note at a revision, and name the one an overwrite replaced

### LINKED ISSUE #1137: [Feature]: agent-reachable overwrite recovery on git vaults — read-at-revision + previous_revision breadcrumb

On a git write-through vault, an overwritten note's previous content is safe in history — but not reachable through the MCP surface. `get_history` returns the commit log and `get_diff` returns diffs (truncated for large notes via `_truncate_diff`), so the only in-band "recovery" is reverse-engineering old content out of a possibly-truncated diff. No tool returns a note's content at a revision, and nothing tells the agent *which* revision holds the pre-overwrite content. In practice "git is the recovery" means a human with shell access to the checkout, not the agent.

This is the git-side complement of the #1064 overwrite guard: the guard makes an overwrite deliberate; this makes a deliberate (or historical) overwrite reversible from the same seat it was made from.

`[verified: read `_server_tools/git.py` (three registered tools: `get_history`, `get_diff`, `git_sync`), `managers/git_query.py`, and `git/query.py` — `resolve_path_at_ref` and an internal `git show` exist as plumbing, but no content-at-revision surface]`

## Desired outcome

Two additive pieces:

1. **Read-at-revision** — a `revision` parameter on `read` (or a small `get_version` tool, whichever fits the tool budget better): returns the note's content at that commit, rename-aware. The plumbing is two existing pieces glued together: `resolve_path_at_ref` for the path the note had at that ref, `git show <sha>:<path>` for the bytes. Read-only and lock-free like the other git queries; hidden/no-op when the vault is not a git repository, matching how the git tools gate today.
2. **`previous_revision` breadcrumb** — in write-through mode every write is an auto-commit, so the pre-overwrite content is exactly the parent of the write's own commit. An overwrite's `write` result carries `previous_revision: <sha>`, and the #1064 guard's messaging can reference the same route. Recovery becomes a two-step the agent can execute verbatim: `read(path, revision=<sha>)` → `write(..., if_match=...)`.

Both are additive (new optional parameter, new result field), so they fit a 4.x minor — no interaction with the 5.0 default flip (#1136).

## Explicitly out of scope

A `restore(path, revision)` convenience tool. The two-step above expresses it fully, and the tool budget is worth guarding. If it is ever added, it must go through the normal write pipeline (indexing, OKF `log.md` upkeep, auto-commit) — never a bare `git checkout`, which would bypass the index and the write callbacks.

## Documentation

`docs/tools/index.md` (`read` entry + git tools section), `docs/design/design.md` (write result schema + git query semantics), `README.md` if the tool table changes, and the server instructions if the recovery route is worth teaching clients (`_instructions.py`).

Refs #1063, #1064, #1136.

pvliesdonk:
## Closes / Refs

Closes #1137. Refs #1063, #1064, #1136, #1284, #1285.

## What & why

An overwritten note's previous content was safe in git history but not reachable
through the MCP surface: no tool returned a note's content at a revision, and
nothing said which revision to ask for. "Git is the recovery" meant a human with
shell access to the checkout. This is the git-side complement of the #1064
overwrite guard: that one makes an overwrite deliberate, this one makes a
deliberate (or historical) overwrite reversible from the seat it was made from.

Two additive pieces. `read(path, revision=<sha>)` returns the note as it stood at
that commit; an overwriting `write` returns `previous_revision`, the commit
holding the content it replaced. Recovery is then executable verbatim:

```text
write(path, content=…)                → previous_revision: 9f2c1ab
read(path, revision="9f2c1ab")        → the replaced content
read(path)                            → the current etag
write(path, content=…, if_match=…)    → restored
```

Verified byte-exact, frontmatter included: `read_revision` returns the raw file
and `write` with no `frontmatter` argument stores content verbatim, so the two
compose without the header being re-serialised or doubled.

### The design decision, and why the first attempt was closed

PR #1281 was closed after review found it returning **a different note's
content**: after `a.md` → `b.md` and a new note created at `a.md`, reading
today's `a.md` at an old revision handed back what now lives in `b.md`, into a
route whose next step is a write.

The root cause was not a slip. It promised note *identity* while delivering
git's *by-path* reading. Git records paths, blobs, and add/rename events —
nothing else — so identity has to be **derived from what git records, and
refused where the records do not reach**. For a read whose caller's next action
is a write, bytes belonging to another note are worse than any error.

So resolution walks git's own records newest-first, carrying a tracked path: an
`R` record moves it back, an `A` on the tracked path stops the walk and raises
(git is saying the file there was born after the revision asked for), a `C`
raises likewise (a copy is a birth, not a lineage), `D`/`M`/`T` continue, and
**any unrecognised record raises** — an unfamiliar record is not evidence of
continuity.

Established by experiment rather than reasoning (all in `tests/test_git_revision.py`):

- **`-m --first-parent` is load-bearing.** A rename performed *while resolving a
  merge* belongs to no parent's diff, so a plain walk returns **zero records**,
  falls through to trusting today's path, and returns whatever occupied that
  name back then. With the flags it reports as an ordinary `R096`.
- **`--find-renames=30`** (the threshold #338 already set) matters: a rename that
  also rewrote half the note is `A` at git's 50% default and `R031` at 30%. The
  test asserts git's default would *not* pair them, so the threshold cannot be
  quietly dropped. Passing it also overrides the operator's `diff.renames`
  (checked for true/false/copies), so the behaviour is this code's, not their config's.
- **A rename that rewrote the note almost entirely in one commit is `A` at any
  threshold** — unconnectable, so it refuses.
- Refusals: reused name, delete-then-recreate, untracked note, non-ancestor
  revision, symlink at that revision, non-UTF-8 content, over the read cap.
- A note **deleted** since is still readable at a revision that has it; that is
  how a deleted note is recovered.

### The breadcrumb is verified, not assumed

`previous_revision` names a commit only when one **provably** holds the replaced
content: the note is tracked, and `git diff --quiet <that commit> -- <path>` is
clean. The comparison is against the *named commit* rather than `HEAD` because
the commit worker runs on its own thread and does not hold the write lock, so a
HEAD-relative check can answer about a commit the first probe never named.

Content that never reached a commit gets **no** breadcrumb — two writes in quick
succession leave the first one's content uncommitted, and git cannot recover it
at all. The field is absent rather than null, and the concurrency bound is
stated where the field is defined: it can name a revision one generation older
than what was replaced, never a revision of a *different* note.

The probe is wired into `WriterFacet`, not `DocumentManager`. That class is also
the server's own maintenance write path — OKF link conversion rewrote 2,595
notes in one call (#1264) — and a probe there would spend several subprocesses
per file on a result nobody reads. A test asserts it fires zero times on that path.

### `RevisionReader`, not a wider `HistorySource`

`tests/test_git_interfaces.py` pins that a two-method history-only backend
satisfies `HistorySource`, and `design.md` states that property deliberately
(#1229). Widening it would have made the narrower capability inexpressible, so
the revision surface is a fourth protocol folded into `VersionedStore`, with
`GitQueryManager` gating on `isinstance`.

- **No `restore(path, revision)` tool.** The issue scopes it out and the two-step
  expresses it. If ever added it must go through the write pipeline (indexing,
  OKF `log.md` upkeep, auto-commit), never a bare `git checkout`. No tracking
  issue filed; say the word.
- **No revision reads for attachments**, and therefore no breadcrumb on an
  attachment overwrite — a revision read returns text, attachment blobs are binary.
- **No breadcrumb on `edit` / `append` / `delete` / `rename`.** The issue scopes
  the field to `write`. `append` and `write_attachment` share the result type and
  never set it, so the tool layer omits the key rather than shipping a
  permanently-null field on their responses.
- **`get_history` / `get_diff` keep their present `--follow` behaviour**, reused
  names included. Fixing that changes what those tools *list*, is not needed for
  recovery, and is **#1285** — filed off this work, and it answers the
  "unverified" question on #1137: they are affected today.
- **No server-instructions snippet.** The budget is tight (#1253/#1258); the
  route is taught in the `write` result's own docstring, where a model that just
  overwrote something is already reading.
- **`_check_no_clobber`'s message is untouched.** That guard fires *before*
  anything is destroyed, so pointing it at a recovery route would mislead.
- **No `INDEX_SEMANTICS_VERSION` bump** — nothing changes how a note's stored
  rows are derived from its bytes; this path stores nothing.

## Not breaking

No `!`. No env var, CLI flag, deployment layout or on-disk format moves.
`git/interfaces.py` was introduced by `7b36411b` and `git tag --contains` on it
is empty, so that seam has never shipped; `WriteResult` gains a defaulted field;
and `read` gaining a parameter is a tool-surface change, which is additive and
dual-mode (omit it for the old behaviour).

## Local review

- [x] Ran `/code-review` on the cumulative diff — five charters, then a refute
      pass. Report below.
- [x] Nothing carries `!` — see **Not breaking**.

## Docs impact

- [x] `README.md` — the git bullet names the breadcrumb and the revision read.
- [x] `docs/` site pages — `docs/tools/index.md`: a `revision` parameter row, an
      "At a revision" response tab, a "Reading an earlier revision" section
      carrying the two-step and every refusal case, `previous_revision` on the
      `write` entry, and `get_history`'s `sha` row pointing at the route.
- [x] `docs/design/` — the resolution rule and why each flag is load-bearing,
      the breadcrumb's proof obligation and concurrency bound, `RevisionContent`,
      the `RevisionReader` protocol, the facet and manager tables, `module-map.md`.
- [x] Inline docstrings — every new public function, plus `docs/api/types.md`
      and `docs/api/facets.md`.

## Gates run locally

`uv run pytest -q` (4115 passed, 18 skipped) · `ruff check --fix .` → `ruff
format .` → `ruff format --check .` · `mypy src/ tests/` · `bash
scripts/structural_gate.sh` (100%, 0 violations over 1929 diff lines) ·
`uv run pre-commit run --all-files` (Vale included) · `diff-cover` against
`origin/main`: **95% patch coverage**, the 9 uncovered lines being
subprocess-failure branches · `tests/test_client_surface_budget.py` passes —
the added docs put `tool_input_schemas` 66 units over, so the prose was trimmed
rather than the budget raised.

## Self-review a6fd43a8..37578f9c

Charters: rules, diff, comments, history, conformance. Coverage: full, except the
optional past-PR-reviews charter (skipped). The diff charter was run twice by
subagent and both runs died on API errors, so I worked it myself; findings below
are mine.

**One blocker, found and fixed:**

- `src/markdown_vault_mcp/git/query.py` — blocker/verified — a note named
  `star*note.md` read back **empty** at any revision, which the documented
  recovery route would then write over the note. `<ref>:<path>` is a revision
  expression, not a pathspec: unmatched, git resolved it to the *commit* object,
  so the size check passed against a 117-byte commit, and `cat-file blob` then
  failed with a message on stderr and an exit status of **zero** — read as an
  empty note. Evidence: `git cat-file -t "<ref>:star*note.md"` → `commit`;
  `git cat-file blob …` → `fatal: … bad file`, `echo $?` → `0`. Fixed by looking
  the entry up through `git ls-tree -l -z` and fetching the blob by its hash.
- Same commit, second trap — `ls-files` and `log` *do* glob their pathspecs
  (`ls-tree` does not), so a note named `n*te.md` matches a tracked sibling
  `n!te.md`, satisfying the untracked-note guard and letting the walk return the
  content of the note deleted from that path. Every pathspec now carries
  `:(literal)`. Mutation-checked: reverting either fix fails a test. The second
  test's first fixture was itself bitten by the same hazard — `git rm 'n*te.md'`
  globbed and removed the decoy, so it passed while proving nothing until the
  mutation check exposed it.

**Important, fixed:** the `logging-standard` message shape on the one new log
call; `RevisionContent.content` documented as "frontmatter included" in four
places while a `section=` read replaces it with a bare section body; `git/query.py`'s
module docstring giving an exhaustive account of `git_root=None` across entry
points that now deliberately differ; every "three concerns / three facets / all
three" claim about the git seam, now count-free rather than incremented so the
next protocol does not reopen the sweep; a comment claiming git emits no copy
records without `--find-copies`, true of its own invocation but not of the
`--follow` walk added three functions away; a claim that the probe costs two
subprocesses when it runs three; `WriteResult`'s field doc enumerating when the
breadcrumb is absent without saying only an overwriting `write` sets it; and the
probe running before the read-only check, so a refused write still spent git calls.

**Struck on refutation:** pinning `diff.renames` alongside `diff.renameLimit`
(suggested by a charter) — `--find-renames=30` already overrides it, verified for
true/false/copies, so pinning would be redundant. Marking the walk's `C` branch
defensive like its older sibling — a passing test produces a real `C033` record,
so the branch is live. `README.md` needing no change — its search bullet already
names `read(path, section=heading)`, so the "README does not name parameters"
premise is false; extended instead.

**No normative content in this change:** no JSON Schema, OpenAPI or proto
artifact is touched; the MCP input schema is derived from the signature by
FastMCP and its budget guard passes; `tests/public_import_surface.txt` is
correctly unchanged, since the new names live in submodules and the package root
exports nothing.

---


pvliesdonk:
@claude review
@codex review

pvliesdonk:
## Review round 1 — all five findings fixed, in `5bda2249`

Every fix carries a test, and every test is mutation-checked: reverting the fix
fails exactly that test and nothing else.

### P1 — historical path outside the vault (`@codex`)

Reproduced exactly as described, and it is the most serious defect in the PR.
`validate_history_path` confines the *caller's* path; the path the walk
resolves to was confined by nothing, so on a vault nested in a larger
repository a note renamed in from outside returned a blob the vault never
contained:

```
RETURNED: 'SECRET OUTSIDE THE VAULT\n'  historical_path='outside-secret.md'
```

The resolved path must now lie beneath the vault prefix (empty prefix — vault
*is* the repo root — means nothing can fall outside it).
`test_history_outside_the_vault_is_refused` pins it, and asserts the refusal
message does not carry the content either.

### P1 — probe and write under one lock (`@codex`)

Right, and my mitigation was weaker than I claimed. I had *documented* a bound
("can never name a revision of a different note"); your rename-away-and-recreate
sequence is exactly the case that breaks it. The lock is a `threading.RLock`,
so `WriterFacet` can now hold it across probe and write via a new
`DocumentManager.write_scope()`, and `DocumentManager.write` re-acquires it
without deadlocking. The claim in `WriteResult`'s docstring is replaced by the
fix rather than kept as a caveat.

The bulk maintenance path is unaffected: `okf_convert_links` and the convention
maintainer call `DocumentManager.write` directly, so they pay neither the lock
hold nor the probe — still asserted by `test_maintenance_writes_do_not_probe_git`.

### P2 — section reads and the whole-note cap (`@codex`)

Confirmed against a 100-byte cap: the on-disk section read succeeded and the
revision one raised. `DocumentManager.read` returns a section before it ever
stats the file, and the cap's own error points callers at `section=` as the way
around it, so applying it here withdrew that escape hatch precisely where the
note is too large to read whole. A `section=` revision read is now exempt, and
the test asserts parity with the on-disk read while the *whole* note at that
revision still raises.

### P2 — Git LFS (`@codex`)

Taken, with one correction to the premise: this repository does not resolve LFS
per read — it runs `git lfs pull` at startup, so the working tree holds real
bytes and an ordinary read gets them by accident of the smudge filter. Your
conclusion stands regardless, and the consequence is data loss: restoring what
git has would write the pointer over the note. A blob beginning with the LFS
pointer version line is now refused, pointing the caller at the LFS object.
Resolving the pointer instead would need `git lfs smudge` and the object
fetched; refusing is the honest answer for a route whose next step is a write.

### `design.md` described the pre-fix mechanism (`@claude`)

Correct, and a good catch: the blocker fix swapped `cat-file -s` for
`ls-tree -l` in the code, and the paragraph describing it was left behind —
the code changed and the prose describing it did not. Rewritten to the shipped
mechanism, including why `<ref>:<path>` is avoided (it is a revision
expression, and an unmatched one resolves to the *commit*, after which
`cat-file blob` fails on stderr with exit status **zero** and reads back as an
empty note). The user-facing refusal list in `docs/tools/index.md` gained the
two new cases as well.

---

Gates after the round: 4119 passed · patch coverage 97% · structural gate 100%
over 1947 diff lines · ruff, mypy, Vale, gitleaks all clean.

---

pvliesdonk:
@claude review

PR #1291, opened 2026-09-03
fix(git): stop handing back non-ASCII note paths git octal-escaped

### LINKED ISSUE #1282: [Bug]: non-ASCII note paths come back octal-escaped from get_history, and empty from per-commit get_diff
## What happened?

On a vault holding notes with non-ASCII filenames, `get_history` reports paths
an agent cannot use, and `get_diff(per_commit=True)` returns an empty diff for
a commit that did change the note.

Found while reviewing #1281, which introduced a third `git log --name-only`
reader and hit the same behaviour there. That new instance carries
`-c core.quotePath=false`; these two do not.

## Observed

Against a repository with `café.md` renamed to `résumé.md` and edited in the
same commit, on `f7be9cf` (the #1281 branch; both call sites are unchanged
from `main`):

```
get_history(path=None)  → paths_changed = [['"caf\303\251.md"',
                                            '"r\303\251sum\303\251.md"'],
                                           ['"caf\303\251.md"']]
get_diff(path="résumé.md", since_sha=<first>, per_commit=True)
                        → one commit, diff == ''
```

The `get_diff` case is the quieter of the two: no error, just an empty diff
for a commit that rewrote the note.

## Expected

`paths_changed` carries paths in the same form every other tool accepts, so a
caller can pass one straight to `read`. A per-commit diff of a commit that
changed the note is not empty.

## Reproduction

```python
# in a repo with a committer identity configured
(repo / "café.md").write_text("# v1\n"); git add -A; git commit -m add
first = <that commit>
git mv "café.md" "résumé.md"; (repo / "résumé.md").write_text("# v2\n")
git add -A; git commit -m "rename+edit"

s = GitWriteStrategy()
s.get_file_history(repo, None, None, 10)                     # quoted paths
s.get_file_diff(repo, repo / "résumé.md", first, True)       # empty diff
```

## Suspected cause

`[verified: ran the reproduction above and compared the same commands with
`git -c core.quotePath=false`, which returns the raw UTF-8 names]` Git's
`core.quotePath` defaults to true, so `git log --name-only` emits non-ASCII
path components octal-escaped inside double quotes. `_vault_relative_paths`
and `_parse_log_block` in `src/markdown_vault_mcp/git/query.py` treat that
rendering as the literal filename.

## Open questions / scope

- `[unverified]`: paths containing a double quote, a tab, or a newline are
  quoted by git regardless of `core.quotePath`, so they may need the
  NUL-delimited (`-z`) form rather than the flag. I have not verified whether
  such a path can be created through this server's write path.
- Scope is the two `git log --name-only` readers in `git/query.py`
  (`get_file_history`'s command, and `_per_commit_diffs`). The third reader,
  `_path_at_ref_via_follow`, is not affected — it was introduced with the flag
  in #1281.

---

pvliesdonk:
## Closes / Refs

Closes #1282

## What & why

Git's `core.quotePath` defaults to true, so `git log --name-only` renders a
path with a non-ASCII component octal-escaped inside double quotes. Both
`git log --name-only` readers in `git/query.py` treated that rendering as the
filename: `get_history` reported `paths_changed` entries no other tool
accepts, and the `per_commit=True` walk in `get_diff` reused the quoted path
as that commit's pathspec, which returned an empty diff for a commit that had
rewritten the note. Both now run `git log -z`, which NUL-frames the header
fields and the `--name-only` paths alike.

`-z` rather than `-c core.quotePath=false`: the flag leaves the names git
quotes unconditionally — a double quote, tab, or newline in the path — still
quoted, so it answers the issue's open question only halfway. It is also what
`resolve_path_at_ref` and the `get_file_at_ref` walk already use.

Two notes on the issue text, both stale rather than wrong at the time:
`#1281` was closed and superseded by #1288, so the third reader it mentions
(`_path_at_ref_via_follow`) does not exist under that name on `main` — its
successor `_revision_walk_output` reads `-z --name-status`, not the flag.
And the issue's open question about quotes, tabs, and newlines is answered
here: `-z` covers them, and there is a test for all three.

Sweeping the class turned up the same defect in the rebase-conflict resolver
(`git/conflict.py`), where it is louder than a bad display string: a quoted
path made the `git show REBASE_HEAD:<path>` read miss the file, and the
`git checkout --ours` that follows runs with `check=True`, so a conflict on a
non-ASCII note raised `CalledProcessError` mid-resolution. Verified by
reverting the fix under the new test. It reads its conflicting paths under
`-z` now too.

**One deliberate extension beyond the issue's wording.** The issue asks only
that the per-commit diff not be empty. Verifying that showed the patch itself
still naming the note `"a/r\303\251sum\303\251.md"`, so `get_diff`'s patch and
`--stat` payloads now render under `-c core.quotePath=false`. Rendering and
framing are different problems and take different tools: `-z` is not available
for a patch body, and the config override cannot be trusted where output is
parsed. Nothing in the codebase parses these payloads — `managers/git_query.py`
returns them to the caller unaltered — so the override is safe there and
`--numstat` binary detection, which does parse, is deliberately left alone.

- Does not stop the subprocess text-mode decode from rewriting a carriage
  return inside a path. `-z` keeps the path in one token; text mode still
  translates a lone `\r` within it, and collapses a `\r\n` pair to a single
  byte, so the path can come back shorter than git wrote it. That is a
  pre-existing property of all
  five `-z` readers here, including the one merged two days ago in #1288, and
  fixing it means moving five call sites to bytes and decoding their `stderr`
  by hand. Filed with a verified repro as **#1290** and cited from both
  docstrings that would otherwise have overclaimed.
- Does not fix the note-identity reuse in `get_history` / `get_diff` tracked
  by **#1285**; this PR touches how their paths are framed, not which note
  they resolve.
- Does not change `_rename_aware_diff_args`'s rename threshold or the
  `\x1e` block sentinel. The sentinel's exposure widens marginally — under
  quoting, a raw `0x1e` in a path was escaped and could not split a block, and
  now it could — so the comment claiming it "cannot appear in filenames" was
  corrected rather than left standing. The server's own write path drops the
  byte (`validate_path("a\x1eb.md", ...)` resolves to `ab.md`), so only an
  externally-added note can carry one.

## Local review

      existed at the **last stable release** — none here; no `!`, and the
      change is invisible to any vault whose note names are ASCII.

## Docs impact

- [ ] `README.md` — no new env var, tool, flag, or config option.
- [ ] `docs/` site pages — `docs/tools/index.md`'s `paths_changed` row makes
      no claim about escaping and is now simply true.
- [x] `docs/design/` — both `get_file_history` and `get_file_diff` bullets
      record the `-z` framing and the `core.quotePath=false` rendering
      override, with the reason each is the right tool for its half.
- [x] Inline docstrings — `_split_z_block` (new), `_history_log_cmd`,
      `_vault_relative_paths`, `_parse_history_block`, `_parse_log_block`,
      `_per_commit_diffs`, and the `_HISTORY_SENTINEL` / `_READABLE_PATHS`
      comments.

## Self-review cf99858..4ab6bb0c

Charters: rules, diff, comments, history. Conformance: no formal schema or
RFC-2119 prose in range. Coverage: full for those five; the optional past-PR
charter was not run, so prior review threads on `git/query.py` — which #1288
touched two days ago — were not checked for conventions this diff re-violates.

Two rounds. **Round 1** (`cf99858..f6868db7`):

- `docs/design/design.md:4037` — important/verified — the `get_file_diff`
  bullet documented the `-z` half and not the `core.quotePath=false` half, so
  a reader of the doc alone would conclude the override had been rejected
  project-wide while the same commit adopts it in four invocations.
  Resolved in `bcc2a3f4`.
- `src/markdown_vault_mcp/git/query.py:70,727` — important/verified — two new
  docstrings claimed paths arrive "byte for byte"; text mode rewrites a
  carriage return inside a path. Verified by diffing a `text=True` read
  against a bytes read of the same command. Resolved in `b168a21d`, with
  #1290 filed.

**Round 2** (`f6868db7..b168a21d`, the commits round 1 produced) found the
precision commit was itself imprecise:

- `src/markdown_vault_mcp/git/query.py:74` and `docs/design/design.md:4036` —
  important/verified — "a lone carriage return is the one byte still
  rewritten" understates: text mode also collapses `\r\n` to a single `\n`,
  so the path comes back *shorter* than git wrote it. The design bullet
  separately still claimed `paths_changed` now carries paths every tool
  accepts.
  Evidence: for a note named `a\r\nb.md`, `text : '\na\nb.md\x00'` against
  `bytes: b'\na\r\nb.md\x00'`.
  Resolved in `4ab6bb0c`; #1290 amended with the same repro.

Round 2 surfaced nothing left open, so review converged here rather than
continuing.

Minors, aggregated and fixed: two `# type: ignore[misc]` in the new framing
tests replaced with an explicit `assert ... is not None`; the empty-token
guard in `_vault_relative_paths` was the one added line without coverage and
now has a unit test. One divergence considered and refuted: `#1290` is the
first *open* issue cited from shipped source here (every other citation —
#232, #338, #342, #472, #519, #560, #683, #745 — points at a closed one), but
the number is what gives whoever closes it a trail to the prose that must
change, so it lowers rot risk rather than raising it.

Archaeology confirms the direction rather than contradicting it: #1282 was
filed by `e888b91d`, the commit that chose `-c core.quotePath=false` for the
revision walk, and `72b46085` on the closed #1281 branch replaced that flag
with `-z` after a reviewer repro — "the flag was the cheaper answer, not the
right one". No previously-fixed pattern is re-introduced and no assertion
pinning the quoted rendering was removed; the two fixture edits in
`tests/test_git_query.py` and `tests/test_git_force_methods.py` are NUL-framing
adaptations only.

Both fixes are mutation-verified: reverting the `-z` fails all four log-reader
tests, and reverting the conflict fix raises `CalledProcessError` on
`git checkout --ours -- '"journ\303\251e.md"'`.

---


pvliesdonk:
@claude review

PR #1298, opened 2026-09-04
fix(git): stop attributing a previous note's commits to one that reused its name

### LINKED ISSUE #1285: [Bug]: single-note get_history and get_diff attribute a previous file's commits to a note that reused its name
## What happened?

Expected `get_history("a.md")` to list the commits of the note now at `a.md`;
it also lists commits belonging to a different note that used to hold that
name, and `get_diff` renders a diff spanning the two.

## Observed

Against `86f3050d`, in a repository where a note was renamed away and a
different note later created under the freed name:

```
commit 1 (a018c2a)  a.md = "ORIGINAL (now b.md)"
commit 2 (fc4fe3e)  git mv a.md b.md
commit 3 (cbcde5d)  a.md = "NEW note reusing the name"
```

```
GitWriteStrategy().get_file_history(repo, repo / "a.md", None, 10)
  -> cbcde5d c3   |  fc4fe3e c2   |  a018c2a c1

.get_file_diff(repo, repo / "a.md", "a018c2a", False)
  -> diff --git a/a.md b/a.md
     -ORIGINAL (now b.md)
     +NEW note reusing the name
```

`c2` and `c1` belong to the note that today lives at `b.md`; the note at
`a.md` was created at `c3`. The single-file diff pairs one note's content
against another's as though it were one file being edited.

## Expected

A single-file history and diff describe the note the caller named. Where the
name was reused, commits from before that note existed are not attributed to
it.

## Reproduction

The three commits above, then the two calls above, run against `86f3050d`.

## Suspected cause

`[verified: ran the reproduction, then compared `git log --follow
--name-status` output for the same path]` Both single-file queries pass
`--follow` (`src/markdown_vault_mcp/git/query.py:225` for history,
`:700` for the per-commit diff). For `a.md`, `git log --follow --name-status`
reports `A a.md` at `c3`, `D a.md` at `c2`, `A a.md` at `c1` — it crosses the
delete boundary and continues into the earlier file's history rather than
stopping at the birth of the file the caller named.

## Open questions / scope

- Scope is the two single-file `--follow` queries in `git/query.py`. Directory
  and vault-wide queries do not pass `--follow`.
- `[unverified]` Whether a note deleted and later recreated at the same path
  produces the same attribution. I did not test that shape here.
- Surfaced while designing #1137, whose revision read must not return one
  note's bytes under another note's name; the behaviour above is older than
  that work and is not a blocker for it. Sibling of #1282, which concerns a
  different defect in the same two readers.

---

pvliesdonk:
## Closes / Refs

Closes #1285

## What & why

`git log --follow` does not stop at the commit that created the file it
follows. Where a note's name was freed — by a rename or by a delete — and a
different note was later created under it, `--follow` walks straight past that
boundary and reports the previous occupant's commits. Both single-note readers
took that output at face value, so `get_history("a.md")` listed another note's
commits and `get_diff` rendered a patch pairing one note's content against
another's as though it were one file being edited.

The boundary is now established from git's own `--name-status` records:
`git log -z --follow --name-status --find-renames=30 [<range>] -- <path>`
walked newest-first carrying a tracked path, stopping at the `A` (or `C`)
record that creates the note — the same reasoning `get_file_at_ref` already
uses. The two `--follow` readers filter their commits by the result; the
single-range diff takes the empty tree as its left side when the note was born
inside the range, so it reads as the creation it is.

Three properties of that walk are load-bearing, and each is there for a
failure this change would otherwise have shipped:

- **Its own invocation, not a reading of the caller's stream.** `--since` /
  `--until` and the `-n` cap trim what the caller's `git log` returns; a window
  that closes before the note was created leaves no birth record to stop at,
  and the filter would silently revert to the old behaviour.
  `test_date_window_does_not_reopen_the_boundary` pins it.
- **`--find-renames=30`, the threshold the rest of the module pins (#338).** At
  git's 50% default, a commit that renames a note *and* rewrites it reports as
  a plain `A` — indistinguishable from a reuse. Without the pin,
  `tests/test_git.py::TestGetFileDiff::test_single_diff_across_rename`
  regressed from a content delta to a full-file add.
- **Unknown rather than empty on a failed invocation.** A git failure yields
  `shas=None`, and a caller then leaves its existing behaviour alone instead of
  filtering every commit away against a set it does not trust.

The delete-and-recreate shape the issue left `[unverified]` was checked and is
the same `A`/`D`/`A` boundary; it is covered by tests here.

- **A note renamed away and recreated under the freed name in one commit**
  (deferral, no tracking issue — nothing to hook a fix on): git records that as
  a plain `M` on the path, so no record distinguishes it from an edit. It
  defeats `get_file_at_ref`'s walk (#1137) for the same reason. Out of reach of
  any reading of git's records.
- **The 50%-vs-30% rename-threshold gap between the readers**: #1297. The
  callers' own `--follow` queries still use git's default, so a rename that
  also rewrote the note loses its pre-rename commits from `get_history` even
  though `read(revision=)` will serve them. The lineage filter cannot restore
  commits the query never returned; changing the callers' threshold is a
  behaviour change of its own.
- **The extra walk's cost**, accepted deliberately: single-note `get_history`
  gains one unwindowed `--follow --name-status` walk (and so loses `-n`'s early
  exit); the two diff paths gain one walk each, bounded to `<ref>..HEAD`.
  Correctness under a date window is what buys the unwindowed form.

## Local review

      existed at the **last stable release** — no `!` here: no operator or
      library surface changes, and this restores intended `get_history` /
      `get_diff` behaviour rather than removing a reachable one.

## Docs impact

- [ ] `README.md` — no new env var, tool, or option
- [x] `docs/` site pages — `docs/tools/index.md`: what a single-note
      `get_history` / `get_diff` describes
- [x] `docs/design/` — `design.md`: the lineage boundary, its three
      load-bearing properties, and the shape it cannot see
- [x] Inline docstrings — `_lineage`, `_track_records`, `_repo_rel`,
      `_empty_tree`, `get_file_history`, `_range_diff`, `_per_commit_diffs`,
      and the `GitQueryManager` mirrors

## Self-review 3fc0a5b4..c27139b1

Charters: rules, diff, comments, history, conformance (no normative content),
past-PRs (light).
Coverage: subagent dispatch was unavailable this session, so the charters ran
as a single-agent sequence rather than in parallel; the past-PRs charter
checked #1288 only, the most recent PR touching these readers. Charter 3 was
first run only over the regions around the hunks and was re-run afterwards
over `git/query.py` and `managers/git_query.py` in full at HEAD, which is
where the second finding below came from.

- `src/markdown_vault_mcp/managers/git_query.py:104,202` — important/verified —
  charter 3: the manager docstrings mirrored the old single-note semantics
  ("scopes to that single file (rename-following)") and would have gone stale
  against the fix. Resolved in `a787b375`, in the same commit as the source
  change.

A second, post-push round added the one test Codecov's patch report was
missing (`fdf94183`): the empty-tree `hash-object` fallback came across
untested when it was extracted from `_root_commit_diff`, and the single-range
diff now depends on it for every note created inside the diffed range. Patch
coverage on `git/query.py` is 100%.

- `src/markdown_vault_mcp/git/query.py:806` — minor/verified — charter 3: the
  comment above the range diff's endpoint resolution ("Resolve the path-at-ref
  once…") described only the rename-resolved branch after the empty-tree branch
  was added beside it. Resolved in `c27139b1`.

Nothing else survived refutation. Two candidates were struck by executed
checks rather than by reading: that an empty lineage set could filter every
commit away (the walk follows at least as far as the caller's own, at a lower
rename threshold and without the window, so its set is a superset of what the
caller enumerates — and a failed walk returns `None`, not an empty set); and
that a non-ancestor `ref` would newly mis-render `get_diff` (if the note's
birth is reachable from `ref` it is not in `ref..HEAD`, so nothing changes;
if it is not reachable, the note genuinely did not exist there).

## Test note

`tests/test_server.py::TestReadAtRevisionTool::test_reused_name_is_refused_through_the_tool`
failed twice in roughly nine full-suite runs on this branch, only under
`--cov`, with no traceback captured; it is filed as #1296. It is not this
change: building that test's fixture repository and calling the new helper
against it shows the lineage set equal to every commit in the repository, so
`get_history` returns exactly what it returned before, and the
`read(revision=)` path the assertion exercises is untouched here. Four
dedicated repeat runs did not reproduce it.

---


pvliesdonk:
@claude review

PR #1300, opened 2026-09-04
fix(git): tell a caller when its write cannot leave the host

### LINKED ISSUE #1287: [Bug]: write/edit report success while the managed-git push is failing — stranded writes are invisible to the client
### What happened

On a managed-git deployment (`MARKDOWN_VAULT_MCP_GIT_REPO_URL` set, `GIT_PULL_INTERVAL_S=120`,
`GIT_PUSH_DELAY_S=30`), the server's clone diverged from `origin`. From that moment:

- every pull ended in `Git pull: conflict resolution loop exceeded 50 iterations`
  → `conflict resolution failed, skipping`
- every push then failed `non-fast-forward`
- **`write` and `edit` kept returning success**, and `read` kept returning the new content

So the tools reported durable writes for changes that could never leave the host. This ran for
hours before anyone noticed. Three commits accumulated locally; one of them was an agent's only
copy of its work, and would have been lost had the clone been rebuilt.

The relevant log lines, repeating every `GIT_PULL_INTERVAL_S`:

```
ERROR markdown_vault_mcp.git.conflict: Git pull: conflict resolution loop exceeded 50 iterations
WARNING markdown_vault_mcp.git.strategy: Git pull: conflict resolution failed, skipping
ERROR markdown_vault_mcp.git.strategy: Git push failed: command ['git','-C','/data/vault','push','origin'] returned 1
error: failed to push some refs to '<remote>'
hint: use 'git pull' before pushing again.
```

### Why it is worse than a stuck sync

The client has **no way to distinguish a durable write from a doomed one.** `write` returns
`created: true`; a subsequent `read` returns the content, because both are served from the local
clone. Every signal available to the caller says the write succeeded.

That matters most for the deployments this feature is aimed at: automated agents whose only
route to the repository is this server. Ours have no shell and no git of their own, so a failed
close-out is completely silent to them — the agent believes it saved its work and moves on.

`git_sync` does report the failure (`reason: "conflict_resolution_failed"`), but it is opt-in and
after the fact. Nothing in the write path surfaces it.

### A second, smaller inconsistency

`git_sync(direction="pull", dry_run=True)` reported `fast_forward: true, would_apply: true` for
the same divergence that the non-dry-run call refused with `conflict_resolution_failed`. The dry
run appears not to model the local commits, so it predicted a clean fast-forward that was not
achievable. That made the situation read as benign on first inspection.

### Steps to reproduce

1. Run a managed-git deployment against a remote that other writers also push to.
2. Let the clone fall behind while the server has at least one unpushed local commit, with both
   sides having modified the same file (an append-heavy shared file does this quickly).
3. Observe the conflict loop above, then call `write` to any path.
4. `write` returns success and `read` reflects it — but the commit never reaches the remote.

### What would fix it

In rough order of value to us:

1. **Make the write path surface push health.** If the last push failed, or the clone is ahead of
   its upstream by more than the in-flight window, either fail the call or return a warning field
   (e.g. `"remote": "unsynced"`) so a caller can react. Reporting plain success while the remote
   is unreachable is the actual data-loss risk.
2. **Expose the state cheaply** — an `ahead`/`behind`/`last_push_ok` field on `stats`, so an
   operator can alert on it without shelling into the host. We currently detect this with an
   external timer running `git rev-list --count @{u}..HEAD`, which works but needs host access.
3. **Reconsider the 50-iteration cap as a terminal state.** Giving up is reasonable; giving up
   *silently and repeatedly forever* is what turned this into stranded data. Even a one-shot
   `logger.critical` on transition into the failed state, rather than the same WARNING every
   cycle, would have made it visible.
4. **Consider honouring `.gitattributes` merge drivers** if the resolver does not already — we
   have since set `merge=union` on our append-only notes, which removes this conflict class for
   us entirely and may be a reasonable documented recommendation for shared vaults.

Happy to test a patch against a reproduction; we have the failing state captured.

### Version

`ghcr.io/pvliesdonk/markdown-vault-mcp` (`:latest` at the time, digest `sha256:02a03afb…`),
HTTP transport, read-write, managed git.

pvliesdonk:
## Closes / Refs

Closes #1287

## What & why

On a managed-git deployment whose pushes were being rejected, `write` and
`edit` kept returning success and `read` kept serving the new content back,
because both are answered from the local clone. Three commits accumulated on
the host for hours, one of them an agent's only copy of its work. Nothing in
the write path said so.

This adds the missing signal.

- **`git/health.py`** — `SyncHealthTracker`, the one place that remembers
  whether the clone is reaching its remote. Push and pull are tracked as
  independent conditions: a clone can pull cleanly while every push is
  rejected, which is the reported incident's shape, so a successful pull does
  not clear a failed push. Only outcomes that prove the clone cannot send
  (`push_failed`, `non_fast_forward`) or cannot reconcile
  (`conflict_resolution_failed`, `non_fast_forward_with_conflicts`) count;
  `fetch_failed` and `no_remote` are not evidence of stranded commits.
  `since` dates the outage, not the condition that outlives it.
- **Fed from outcomes, under the lock that orders them.** `force_push` and
  `_pull_pipeline` hold the strategy-wide lock across the operation and the
  recording of its result, calling the bodies as `_push_locked` /
  `_pull_locked` (the caller-holds-the-lock contract
  `_force_pull_rebase_fallback` already used); `PushScheduler.do_push`
  records inside the same lock and re-raises for `do_push_safe` to log.
  Recording after releasing it let an older operation publish over a newer
  one — an earlier push's success clearing the failure of the push that ran
  after it. A dry run records nothing: it predicts a pull rather than
  observing one.
- **Every vault-mutating write tool** passes its result through
  `attach_remote_health`, which adds a `remote` object
  (`state` / `reason` / `since` / `detail`) while the clone is unsynced and
  adds nothing otherwise — the `previous_revision` idiom, where absence says
  "nothing to report". Reads take no lock: the strategy-wide lock is held
  across a whole fetch + merge, so reading health under it would make every
  write response block behind a pull.
- **The log marks transitions** (report item 3). Entering the state logs once
  at ERROR (`git_remote_unsynced`), recovery once at INFO
  (`git_remote_resynced`). The per-attempt lines that repeated every cycle in
  the incident move to DEBUG with their git stderr: the rejected push, the
  conflict-resolution loop cap, the "conflict resolution failed" line and the
  sibling-commit failure. Two classes stay loud — an unexpected exception on
  either path (a bug, not a cycle), and any failure that can leave the
  working tree inconsistent (`abort_in_progress_rebase`,
  `restore_upstream_paths`).

`tests/test_write_remote_health.py` partitions `WRITE_TOOL_NAMES`, so a new
write tool has to be classified as carrying the key or exempt.
`tests/test_git_health.py` pins the tracker's state machine, the transition
logging, and that every outcome is recorded while the lock is still held.

The reporter offered to test a patch against their captured failing state:
this is the one, and the `remote` object on a write result is what it adds.

## Reviewer notes

Two decisions worth your disagreement:

1. **No tool docstring mentions the key.** The reviewed client-surface budget
   (`tests/test_client_surface_budget.py`) allows 23,500 UTF-16 units of tool
   descriptions and 23,149 are in use, so one sentence on each of the eleven
   attaching tools (~1,600 units) does not fit, and putting it on only some of
   them would mirror one behaviour in a subset of its call sites — the drift
   class #1006/#1007 spent two PRs removing. The key instead carries its own
   `detail` sentence at the moment it appears, and the shape is documented on
   the tools page, in the git guide, and in `design.md`.
2. **A write is warned, never refused.** Refusing on the first rejected push
   would turn a durability warning into an availability outage for a condition
   the next pull cycle often reconciles, while the local commit does still hold
   the data. The operator-configurable block is #1299.

## Review rounds

A local self-review before the first push fixed four things, the substantive
one being that blanket-demoting the push failure logs would have hidden a
genuine bug on the push path; only the routine rejected-push arm moved.

Round 1 (`91f847e9`) answered three findings, each now carrying a regression
test that fails on the old code:

- outcomes were recorded after the strategy lock was released, so concurrent
  pushes or pulls could publish out of order (Codex P1);
- `since` moved forward when the older of two open conditions closed, making
  a payload understate the outage (Codex P2);
- the pull-side per-cycle lines stayed loud while the guide claimed
  otherwise — and the pull side is what repeated in the reported incident, so
  the class was swept rather than the sentence narrowed (Claude).

`db784b71` then covered the two `except ValueError` fallbacks that the
extraction re-indented into the diff, taking patch coverage back to 100%.

- Refuse writes while the clone is unsynced, or make that configurable: #1299
- Report `ahead` / `behind` / `last_push_ok` on `stats` for operators: #1293
- Fix the dry-run pull that predicts a fast-forward the real pull refuses: #1292
- Document `.gitattributes` merge drivers for shared vaults: #1294

## Local review

- [x] Ran a local code-review pass on the cumulative diff before `gh pr create`,
      and again on the commit each review round produced.
- [x] No commit carries `!`. The MCP surface change is additive (a new optional
      result key), the library surface only gains names, and no operator
      surface moved.

## Docs impact

- [x] `README.md`
- [x] `docs/` site pages — `docs/tools/index.md`, `docs/guides/git-integration.md`
- [x] `docs/design/` — `design.md` (sync-health section, the new facet), `module-map.md`
- [x] Inline docstrings

---


pvliesdonk:
@codex review
@claude review

pvliesdonk:
Round 1 addressed in `91f847e9`. All three findings were real; two of them
changed behaviour, not just prose.

**Codex P1 — outcomes recorded after the lock was released.** Confirmed and
fixed as described. The lock is what orders concurrent pushes (interactive
against deferred) and concurrent pulls (the loop against `git_sync` or a
webhook), so recording outside it let an older operation publish over a newer
one — an earlier push's success clearing the failure of the push that ran
after it, which puts the caller back to being told nothing while its commits
are stranded. `force_push` and `_pull_pipeline` now hold the lock across the
operation *and* the recording; the bodies became `_push_locked` /
`_pull_locked` and carry the caller-holds-the-lock contract
`_force_pull_rebase_fallback` already used. `PushScheduler.do_push` records
both outcomes inside its own lock and re-raises for `do_push_safe` to log.
Three tests in `tests/test_git_health.py` assert the strategy lock is held at
the moment each outcome lands (deferred push, interactive push, pull), so the
invariant is checked rather than described.

**Codex P2 — `since` moved forward on partial recovery.** Reproduced before
fixing (pull fails, push fails, pull recovers → `since` jumped to the push's
timestamp). The outage start is now tracked separately from the conditions
and survives until the clone is actually healthy; a genuinely new outage
after a recovery still gets a new start. Both are pinned by tests.

**Claude — the guide claimed a symmetry the code did not have.** Correct, and
the half that was missing is the one that matters: the lines that repeated in
the reported incident are the *pull* side. Rather than narrow the sentence, I
swept the class — the conflict-resolution loop cap, the "conflict resolution
failed" line, and the sibling-commit failure now sit at DEBUG with the push
line, and the two existing tests that pinned their levels were updated to
assert the tracker's once-per-outage ERROR as the operator-visible signal
instead. What deliberately stays loud: an unexpected exception on either path
(a bug, not a cycle), and any failure that can leave the working tree
inconsistent (`abort_in_progress_rebase`, `restore_upstream_paths`). The
guide and `design.md` now say exactly that.

Gate after the round: 4200 passed, patch coverage 97% (the six lines below
100% are two pre-existing `except ValueError` fallbacks for unparseable
`git rev-list --count` output, which the extraction re-indented into the
diff), ruff/mypy clean, structural gate 100%, `mkdocs build --strict` clean.

---

pvliesdonk:
@claude review
@codex review

PR #1305, opened 2026-09-04
fix(git): treat a note's name as a name, not a glob, wherever git takes a pathspec

### LINKED ISSUE #1303: [Bug]: a note whose name contains glob metacharacters gets another note's history
## What happened?

Expected `get_history` for a note whose name contains `[` to describe that
note. It described a different one: a vault holding both `b1.md` and
`b[1].md` reports `b1.md`'s commits under `b[1].md`.

## Observed

Against `820e359b`, driving `GitQueryManager.get_history` over a two-commit
repository holding both names, where the second commit touches `b1.md` only:

```
HISTORY b[1].md: [('ce9d483', 'edit b1 only'), ('9658b96', 'add both')]
HISTORY b1.md:   [('ce9d483', 'edit b1 only'), ('9658b96', 'add both')]
```

The two notes report identical histories. `edit b1 only` never touched
`b[1].md`.

The same shape at the git level:

```
git log --oneline --follow -- 'b[1].md'
  ce9d483 edit b1 only
  9658b96 add both

git log --oneline --follow -- ':(literal)b[1].md'
  9658b96 add both
```

## Expected

A note's history lists that note's commits. A name containing glob
metacharacters (`*`, `?`, `[`, `]`) describes the note that literally bears
it, not every note the name happens to match.

## Reproduction

```bash
git init -q --initial-branch=main repo && cd repo
git config user.email t@e.com && git config user.name T
printf '# one\n' > 'b1.md'
printf '# bracket\n' > 'b[1].md'
git add -A && git commit -qm "add both"
printf '# one v2\n' > b1.md
git add -A && git commit -qm "edit b1 only"
```

Then `get_history("b[1].md")` through `GitQueryManager`, or the
`get_history` MCP tool on a vault with those two notes.

## Suspected cause

`[verified: ran the two `git log` invocations above against the repository
built by those commits, and the `GitQueryManager.get_history` calls that
produced the output above]` The pathspec reaches git unquoted, so git applies
its default glob matching rather than treating the name literally; the same
invocation with `:(literal)` returns only the note's own commit. I have
verified that the two pathspec forms produce the differing outputs above; I
have not verified which of this project's query call sites are affected
beyond the `get_history` path I ran.

## Open questions / scope

- `[unverified]`: the per-commit `get_diff` walk and the lineage walk may
  share this path.
- Distinct from #1290, which is about how a name comes back *out* of the
  `-z` readers rather than how it goes in as a pathspec.
- Not introduced by #1302 (the rename-threshold fix), which touched the same
  argv without changing how the path is passed.

---

### LINKED ISSUE #1304: [Bug]: a write to a note whose name contains glob metacharacters commits a sibling's uncommitted changes
## What happened?

Expected a write to a note to commit that note. A write to a note whose name
contains glob metacharacters also committed an unrelated sibling note's
uncommitted changes.

## Observed

Against `820e359b`, driving the git write callback over a repository holding
`b1.md` (committed, then edited on disk without committing) and a new note
named `b[1].md`:

```
git log --oneline --name-only
  11d048e write: b[1].md
  b1.md
  b[1].md
  bb4270e seed
  b1.md
```

The commit for `b[1].md` carries `b1.md` as well, and `git status` afterwards
is empty: the operator's in-progress edit to `b1.md` went into the note's
commit. This is the same consequence #894, #1249 and #1273 recorded for other
paths into the auto-commit, reached here through the note's name.

## Expected

A write commits the file it wrote. A note whose name happens to match a
sibling's name as a wildcard pattern does not carry that sibling's work into
its commit.

## Reproduction

1. A git-backed vault holding a note `b1.md`, committed.
2. Modify `b1.md` on disk without committing (an operator edit, a pending
   deferred commit).
3. Write a note named `b[1].md` through the MCP write path.
4. Inspect the resulting commit: it contains both files.

## Suspected cause

`[verified: ran the write callback over the repository described above and
read back the commit contents and `git status`]` The path reaches `git add`
unquoted, and git reads a pathspec as a wildcard pattern unless told
otherwise; `:(literal)` is already used on other query paths in this project
for exactly this reason. I have verified the observed commit contents above;
I have not verified which of the staging, deletion and rename paths are
affected beyond the `write` operation I ran.

## Open questions / scope

- `[unverified]`: the delete and rename staging paths may share this code
  path.
- Distinct from #1303, which records the read-side observation (a note's
  history listing another note's commits) rather than what a write commits.

---

pvliesdonk:
## What this fixes

Git reads a pathspec as a wildmatch pattern unless told otherwise. A vault
accepts a note named `b[1].md` — `Vault.writer.write("b[1].md", …)` creates,
reads and lists it — and from then on that name selected whichever sibling it
matched:

| Call | Before | After |
|---|---|---|
| `get_history("b[1].md")` | `b1.md`'s commits, reported as this note's (#1303) | its own commits |
| `write` to `b[1].md` with an uncommitted edit pending on `b1.md` | commit holds `b[1].md` **and** `b1.md`; the operator's in-progress edit is gone from the working tree (#1304) | commit holds `b[1].md`; the edit is still there, uncommitted |

The second one is the failure #894, #1249 and #1273 each closed on a different
path into the auto-commit, reached this time through the note's own name.

The revision readers already defended against this: `:(literal)` arrived with
#1137 for `get_file_at_ref`, scoped to that walk. This generalises it. `_literal`
moves to `git/_run.py` as `literal_pathspec`, and every command in the package
that takes a **pathspec** builds it there: `log`, `diff`, `show`, `add`,
`commit --only`, `ls-files`, `checkout`.

## The two boundaries, stated rather than left to be discovered

- **`check-ignore` stays raw.** It takes *pathnames*, not pathspecs: it already
  matches literally, and it rejects the magic outright — `fatal: pathspec magic
  not supported by this command: 'literal'`. Wrapping it would have asked about
  a file named `:(literal)b[1].md`.
- **`<rev>:<path>` sites are not in this class.** `_rename_aware_diff_args`'
  blob-spec form and `conflict.py`'s `git show REBASE_HEAD:<path>` are revision
  expressions, resolved by tree lookup rather than wildmatch. Probed both
  bracket and star names: they resolve exactly. (The #1137-documented hazard
  there is an *unmatched* path resolving to something else, which is why the
  content read uses `ls-tree` instead.)

The staging helpers now return their scope in literal form, because that value
exists to be handed back to git as the pathspec limiting `git diff --cached
--quiet` and `git commit --only`. Converting only the `git add` calls would
have left the commit globbing and the sweep intact. The `str(old_path) in
ignored` comparisons keep raw paths, and `conflict.py`'s returned sibling paths
stay raw — they are filesystem paths and user-facing output, not pathspecs.

## Verification

Both bugs were reproduced at product level before any code changed, and each
git command was probed in a scratch repository before being converted —
`--follow`, absolute paths, directory pathspecs, `add`, `add -u`,
`commit --only`, `diff --cached --quiet`, `checkout --ours`, `ls-files`, `rm`,
and `check-ignore` (the one rejection).

Four regression tests, all mutation-checked against a neutered
`literal_pathspec`: write, delete, rename, and history. The write and delete
tests assert the sibling's edit **survives in the working tree**, not just that
it is absent from the commit — the data-safety claim is the point. The rename
test is deliberately included even though the plain-name rename tests already
cover those lines: `_stage_rename` is where the conversion is most intricate,
with literal appends sitting next to raw comparisons.

Full suite 4215 passed, structural gate 100%, `ruff`/`mypy` clean.

## What this deliberately does not do

- **No bracket-name conflict-resolution fixture.** `conflict.py`'s three call
  sites are converted and ride the existing conflict-path tests; building a
  rebase-conflict repository around a glob-named note costs more than the
  remaining risk on three mechanical call sites.
- **#1290 (a carriage return in a name, on the way *out* of the `-z` readers)
  is untouched** — the other direction, a different mechanism.

## Sequencing

Branched off `main` at `c1dd425f`, not on top of #1302. Both touch the
`--follow` argv line in `_history_log_cmd`, so whichever merges second takes a
one-line conflict there; that is expected and trivial.

Milestone left unset on both issues — #1304 is the same sweeps-operator-work
family as #894 / #1273, so v4.2 would be consistent, but that is the owner's
call.

## Self-review c1dd425f..HEAD

Charters: rules, diff, comments, history, conformance. Coverage: full;
single-agent (no subagent fan-out), no past-PR charter.

- `src/markdown_vault_mcp/git/strategy.py:2002` — important/verified — the staging
  helpers' docstrings described their return as "the pathspec that was staged"
  while the form changed to literal, and two consumers document the same value.
  Evidence: `_stage_rename` returns `deletions + pathspec`, both now literal.
  Fixed by naming the form in all four places (producers and consumers).
  Resolved in this branch's HEAD.
- Refuted while reviewing: wrapping `conflict.py`'s `written` / `updated_originals`
  lists themselves rather than at the call site would have leaked `:(literal)`
  into the `conflict_files` a caller is shown; the conversion is at the call
  site, and the lists stay raw. Verified by reading both consumers.
- No history findings: `:(literal)` was introduced narrowly by #1137 for one
  walk, not chosen against for the others.

Closes #1303, closes #1304


pvliesdonk:
@claude review
@codex review

pvliesdonk:
Addressed the one nit in `1bab65fd`: `_tree_entry`'s docstring line had become dense once the helper moved out of this module (an inline `:func:` reference inside a parenthetical). It now reads as prose — the package-wide rule is stated once in `design.md` and on `literal_pathspec` itself, so the aside does not need to restate the reference.

No other changes; both reviews' conversion-coverage traces match what the diff does.

PR #1315, opened 2026-09-04
chore: update copier template to v7.0.2

### LINKED ISSUE #1295: chore: update copier template to v7.0.2
## What

`.copier-answers.yml` records `_commit: v7.0.1`; the template source
(`pvliesdonk/fastmcp-server-template`) has since released **v7.0.2**
(2026-09-02). Adopt the one intervening patch release.

`[verified: read `.copier-answers.yml` at `3fc0a5b4`; `gh release list` on the
template]`

## Template delta (v7.0.1 → v7.0.2)

Five fixes, no breaking changes. Two of them land on surfaces this project
actually has:

- **template#565 — `fix(config)`: comment out `packaging/env.example`, like
  every other env template.** **This is the material one.**
  `packaging/env.example` here carries **82 uncommented assignment lines**,
  among them `FASTMCP_LOG_LEVEL=INFO`, `OLLAMA_HOST=…` and an empty
  `OPENAI_API_KEY=`. Its own header states it installs to
  `/etc/markdown-vault-mcp/env.example` and that the systemd unit sources
  `/etc/markdown-vault-mcp/env` via `EnvironmentFile=`, so an operator who
  copies it to `env` as instructed activates all 82.
  `[verified: `grep -cE '^[A-Z][A-Z0-9_]*=' packaging/env.example` → 82; read
  the file header]`

  The harm upstream recorded (template#290, closed by #565) is not that the
  values are wrong — they are the defaults. It is that an `EnvironmentFile`
  exists to carry *deviations*: a consumer migrating a hand-written file went
  from 1 active + 13 commented to 14 active, and a deliberate `/var/lib/…`
  cache override — needed because the unit runs under `ProtectSystem=strict`
  — was replaced by the default it existed to override.

- **template#569 — `fix(config)`: give the generated wizard spec a route to
  `dockerVolume`/`dockerPath`.** `docs/javascripts/config-wizard/wizard-spec.json`
  here carries `dockerImage` and neither of the other two.
  `[verified: grepped the spec at `3fc0a5b4`]` `[unverified]` whether this
  project's `scripts/gen_config_surface.py` takes the template's generator
  change or diverges from it; I have not compared them.

- **template#563 — `fix(config)`: name the failed config import when the mcpb
  fields guard aborts.** Upstream twin of #1259, whose local half merged as
  PR #1279. `[unverified]` whether the local guard already reports the import
  failure by name after #1279, or only the template's does.

- **template#564 — `fix(ci)`: report diff-cover's coverage percentage, not its
  line count.** `.github/workflows/ci.yml:127-149` builds the patch-coverage
  status description from `diff-cover` output. `[verified: read the workflow]`
  `[unverified]` whether the local copy shows the line-count defect.

- **template#566 — `fix(ci)`: make the Vale style-pack cache actually prevent
  a download.** CI-internal.

## Why stop at v7.0.2

The template is at **v8.0.0** (2026-09-03), which carries two `!` commits —
template#560 `feat!: adopt fastmcp-pvl-core v7 / FastMCP 4` and template#577
`fix!: make compose.yml a working quick start, without a reverse proxy`. That
release is the FastMCP 4 cutover **#1271** already tracks on the v5 milestone,
and adopting it forces a major bump. v7.0.2 is the last template release
adoptable without one.

`[verified: `gh release view v8.0.0` on the template; read #1271]`

## How the fix arrives

Worth knowing before the update, because none of it is visible in a diff:

- `packaging/env.example` is **generated**, not template-rendered — from
  `fastmcp_pvl_core.server_config_surface()` plus `config-presentation.yml`.
  It is `_skip_if_exists` only for update-patch mechanics; it is regenerated
  **unconditionally** by the `_migrations` hook at the end of every
  `copier update`, once the diff has restored `config.py`.
- The fix itself is one flag: `commented: false` → `commented: true` on the
  `packaging/env.example` target in `config-presentation.yml`.
  **`config-presentation.yml` is template-owned** — only
  `config-presentation.domain.yml` is skip-listed — so the flip is re-rendered
  by the update rather than needing a hand edit. This project's copy still
  reads `commented: false` at `config-presentation.yml:467`.
  `[verified: read `copier.yml`'s `_skip_if_exists` and the generated-artifact
  note in the template at v7.0.2; read `config-presentation.yml:458-467` here]`
- The two `examples/*.env` targets keep `commented: false` deliberately; only
  this target and `.env.example` are commented.
- CI runs `scripts/gen_config_surface.py --check` and fails on drift, so the
  regenerated artifacts must be committed in the same change.

## Documentation

template#565 also added a step to `docs/design/config-migration.md.jinja`,
which is template-owned and arrives with the update. `[unverified]` whether
anything else in `docs/` describes the shipped file's comment state.

## Notes

The weekly `copier update` cron is not the route here — see
the `applying-template-updates` skill
(`.agents/skills/applying-template-updates/SKILL.md`)
for the resolution policy and the seeded-once files the diff cannot show.

---

pvliesdonk:
## Closes / Refs

Closes #1295

## What & why

Adopts the one template release between `v7.0.1` and the FastMCP 4 cutover
(`v8.0.0`, which #1271 tracks on v5). Five upstream fixes; two land on
surfaces this project has.

**The material one — template#565.** `packaging/env.example` is now generated
fully commented, like `.env.example`. That file installs to
`/etc/markdown-vault-mcp/env.example`, and `packaging/scripts/postinstall.sh`
copies it to `/etc/markdown-vault-mcp/env` when no env file exists — so its
82 live assignments *became* a fresh install's configuration rather than a
template to edit. Among them was
`MARKDOWN_VAULT_MCP_SOURCE_DIR=/data/vault`, a container path, while
`docs/deployment/systemd.md`'s own "Minimum required" section tells the
operator to set `/var/lib/markdown-vault-mcp/vault`. An `EnvironmentFile`
exists to carry deviations; every line is now a commented default.

The change is comment-prefix-only — 82 removed lines, 82 added, each the
same assignment behind a `# `, no value edited. `[verified: paired the
`-`/`+` lines of the diff programmatically]` A deployed
`/etc/markdown-vault-mcp/env` is never overwritten by the package, so no
existing install changes; this reaches fresh installs only, which is why it
carries no `!`.

Also arriving:

- **template#569** — the wizard hint vocabulary gains `dockerVolume` /
  `dockerPath`. Here that reaches exactly one question:
  `MARKDOWN_VAULT_MCP_BEARER_TOKENS_FILE` now emits `-v <answer>:/etc/markdown-vault-mcp/tokens.toml`
  alongside the container-path override, and a **blank** answer now emits
  neither half — previously it emitted an `-e` naming a file that was never
  mounted, beside `BEARER_TOKEN`, and core prefers a tokens file when both
  are set, so the default bearer output produced a server that refused to
  start. `MARKDOWN_VAULT_MCP_ACL_PATH` gets the same hint upstream but is
  gated out here by `enable_authorization: false`, so that half is a no-op.
  `[verified: the regenerated `wizard-spec.json` validates against
  `wizard-spec-schema.json` and carries one container-path question]`
- **template#563** — the mcpb fields guard now names a failed config import
  instead of blaming the `files:` map. Upstream twin of #1259, whose local
  half merged as PR #1279.
- **template#564** — CI's patch-coverage status reads diff-cover's
  `Coverage:` percentage rather than its `Total:` line count. The local copy
  did carry the defect: an 80-line diff at 50% posted
  `Patch coverage: 80% (minimum 80%)` on a *failing* status.
- **template#566** — `vale-action` v2 → v3, and the style-pack cache becomes
  load-bearing via `sync: false` on an exact cache hit.

**One seeded-once change applied by hand.** `.copier-seeded-changes.md`
reports `.vale.ini` as the only skip-listed file the template changed: its
`Packages =` comment no longer described what CI does. Applied — with one
sentence corrected, see below.

**Post-merge, one thing to watch.** v3 upgrades reviewdog 0.17 → 0.21, and a
`push`-triggered run with a `github-pr-*` reporter previously reported
nothing and exited 0. Pushes to `main` and `release/*` now get the prose gate
they silently skipped. `filter_mode: added` is unchanged, so a merge to
`main` reports only lines that already passed the same gate on the PR — but
watch the first push to `main` before treating a green `main` as routine.
The escape hatch, if it fails on prose that predates this, is
`reviewdog_version: "0.17.0"` on the `Run Vale` step.

- **Does not go to `v8.0.0`.** That release carries two `!` commits — the
  FastMCP 4 / pvl-core 7 cutover and the `compose.yml` rework — and is the
  cutover #1271 already tracks on the v5 milestone. v7.0.2 is the last
  template release adoptable without a major bump.
- **Does not fix the `.vale.ini` wording upstream.** The template's new
  comment ends "Locally, `vale sync` is unconditional and always resolves
  latest", which is false of the mechanism it ships: the `vale-sync`
  pre-commit hook syncs only when a pack directory is *missing*. This copy
  carries a corrected sentence; the template bug is
  pvliesdonk/fastmcp-server-template#586.
- **Does not reconcile the `vale-sync` hook's `exclude:` drift.** This
  project excludes all three internal subtrees there; a fresh v7.0.2 render
  excludes only `docs/superpowers/`, so copier will keep re-offering the
  change. Pre-existing, harmless in effect (the `vale` hook itself is in
  lockstep, and template#529 tracks the exclusion-seam class), and out of
  scope here.
- **Does not touch `server.json` / plugin manifests.** Version lockstep is
  unaffected; no manifest is in the diff.

## Local review

- [x] No commit carries `!`. Assessed against the last stable release: no env
      var, config file, CLI flag, deployment layout, on-disk state format, or
      public import changed. A deployed `/etc/markdown-vault-mcp/env` is never
      overwritten, so no operator has to do anything to upgrade.

### Self-review d4b5e52d..28c574a9

Charters: rules, diff, comments, history, conformance. Coverage: full except
the optional past-PR-reviews charter, skipped — the touched files are
template-owned and re-rendered, so prior review threads on them constrain the
template, not this diff.

One finding, fixed before the push:

- `.vale.ini:56` — important/verified — the seeded-file correction I was
  copying across asserts a local behaviour the shipped tooling contradicts.
  Evidence: the incoming text reads "Locally, `vale sync` is unconditional
  and always resolves latest", while `.pre-commit-config.yaml`'s `vale-sync`
  hook — "vale sync (download style packages if missing)" — runs
  `vale sync` only inside `if [ ! -d ".vale/styles/$p" ]`. Fix: kept the
  substance, replaced the closing sentence with one naming the guarded hook
  and distinguishing it from a hand-run `vale sync`. Resolved in `28c574a9`;
  filed upstream as template#586.

Four template-owned files were compared byte-for-byte against a fresh
`copier copy` of v7.0.2 using this project's own answers — `ci.yml`,
`config-presentation.yml`, `gen_config_surface.py`, `generators.js` are
**identical** to that render, which is what rules out the merge having
displaced local content in any of them. `[verified: `diff` against
`/tmp/.../target-render`]`

## Docs impact

- [ ] `README.md` — nothing describes the shipped file's comment state.
- [ ] `docs/` site pages — `docs/deployment/systemd.md:104` already called
      `/etc/markdown-vault-mcp/env.example` a "Commented example with all
      variables"; the generated file now matches the prose instead of
      contradicting it.
- [x] `docs/design/` — `config-migration.md` gains the template's step for
      rescuing active assignments out of `packaging/env.example` before the
      generator comments them.
- [ ] Inline docstrings — no `src/` change in this diff.

## Gate

`git diff --check`, conflict-marker grep, `uv lock --check`,
`uv sync --locked`, `gen_config_surface.py --check`, `vendor_spa.py --check`,
ruff check + format, mypy, `pytest -x -q` (4222 passed, 19 skipped),
`mkdocs build --strict`, `pre-commit run --all-files`, and
`scripts/structural_gate.sh` (189 lines, 0 violations) all pass locally.

`tests/test_config_wizard_smoke.py` self-skips against a stale `site/`, which
would have skipped every test covering the changed `generators.js`. Rebuilt
the site and re-ran it: 16 passed, including the two new container-path
cases.

---


pvliesdonk:
Closing and abandoning this branch: the update was applied too mechanically and the review of it was not sound. Re-doing #1295 from scratch.

PR #1316, opened 2026-09-04
chore: update copier template to v7.0.2

### LINKED ISSUE #1295: chore: update copier template to v7.0.2

pvliesdonk:
## Closes / Refs

Closes #1295

## What & why

Adopts the one template release between `v7.0.1` and `v8.0.0`. v8.0.0 carries
two `!` commits and is the FastMCP 4 cutover #1271 already tracks on v5, so
v7.0.2 is the last release adoptable without a major bump.

Five upstream fixes. The ones that land on surfaces this project has:

- **template#565** — `packaging/env.example` is generated fully commented,
  like `.env.example`. It installs to `/etc/markdown-vault-mcp/env.example`
  and is a template to copy to `.../env` and edit; an `EnvironmentFile`
  carries deviations from the defaults. 82 assignments, each now behind a
  `# `, no value edited.
- **template#569** — the wizard hint vocabulary gains `dockerVolume` /
  `dockerPath`. Here it reaches one question:
  `MARKDOWN_VAULT_MCP_BEARER_TOKENS_FILE` gains a bind mount, and a blank
  answer now emits neither the mount nor the container-path override.
  `MARKDOWN_VAULT_MCP_ACL_PATH` gets the same hint upstream but is gated out
  by `enable_authorization: false`.
- **template#563** — the mcpb fields guard names a failed config import.
  Upstream twin of #1259, whose local half merged as PR #1279.
- **template#564** — CI's patch-coverage status reports diff-cover's
  `Coverage:` percentage rather than its `Total:` line count.
- **template#566** — `vale-action` v2 → v3; the style-pack cache becomes
  load-bearing via `sync: false` on an exact hit.

`.copier-seeded-changes.md` reports `.vale.ini` as the only skip-listed file
the template changed. Its corrected comment is applied verbatim.

Zero conflict markers, no new `.copier-answers.yml` keys.

Two notes for the reviewer, neither a defect:

- This commit edits `.vale.ini`, which is what the CI cache key hashes, so the
  first run re-syncs the unpinned style packs. That is the documented way to
  bump them.
- `vale-action` v3 upgrades reviewdog 0.17 → 0.21. A `push`-triggered run with
  a `github-pr-*` reporter previously reported nothing and exited 0, so pushes
  to `main` and `release/*` now get the prose gate they were skipping.
  `filter_mode: added` is unchanged. The escape hatch, if it ever fails on
  prose that predates this, is `reviewdog_version: "0.17.0"` on the `Run Vale`
  step.

- **Does not adopt v8.0.0** — the FastMCP 4 / pvl-core 7 cutover, tracked by
  #1271 on v5.
- **Does not reconcile the `vale-sync` hook's `exclude:`** — this project
  excludes all three internal doc subtrees there, a fresh v7.0.2 render
  excludes only `docs/superpowers/`. Pre-existing and untouched by this diff;
  template#529 tracks the exclusion-seam class.

## Local review

- [x] No commit carries `!`. Nothing an operator or a library consumer of the
      last stable release must change to upgrade: a deployed
      `/etc/markdown-vault-mcp/env` is never overwritten by the package.

### Self-review d4b5e52d..b364d0f5

Charters: rules, diff, comments, history, conformance.
Coverage: full except the optional past-PR-reviews charter, skipped — the
touched files are template-owned and re-rendered, so prior threads on them
constrain the template, not this diff.

**No findings.** Every hunk in this diff is machine-produced — rendered by
`copier update` from v7.0.2, or written by `scripts/gen_config_surface.py`
during the update's migration hook. What each charter checked:

- **rules** — the AGENTS.md gates pass (below); `server.json`,
  `.claude-plugin/plugin/.claude-plugin/plugin.json` and
  `.claude-plugin/plugin/.mcp.json` are absent from the diff, so manifest
  version lockstep is untouched; `docs/design/config-migration.md` arrived
  with the update.
- **diff / history** — every changed file was compared against a fresh
  `copier copy` of v7.0.2 using this project's own `.copier-answers.yml`.
  Eight are byte-identical to that render, `.vale.ini` among them. The seven
  that diverge are the project's own by construction (`.copier-answers.yml`,
  the generated `wizard-spec.json` and `packaging/env.example`, the
  hook-written `.copier-seeded-changes.md`) or carry domain content
  (`AGENTS.md`, `mkdocs.yml`, `.pre-commit-config.yaml`) — and every one of
  them diverges *less* now than at `d4b5e52d`, which is what rules out the
  merge having displaced local content.
- **comments** — the one mechanical lockstep `.vale.ini` documents holds
  across all four sites: `Packages =`, `BasedOnStyles`, the `ci.yml` cache
  `path:` list, and the `for p in` list in `.pre-commit-config.yaml` all name
  `Google proselint write-good ai-tells`. The new `sync:` expression
  references `steps.vale-cache`, and the cache step declares `id: vale-cache`.
- **conformance** — `wizard-spec.json` validates against
  `wizard-spec-schema.json`, including the new `dockerVolume` key.

## Docs impact

- [ ] `README.md` — no user-facing change to document.
- [ ] `docs/` site pages — none affected.
- [x] `docs/design/` — `config-migration.md` gains the template's step for
      rescuing active assignments out of `packaging/env.example`.
- [ ] Inline docstrings — no `src/` change in this diff.

## Gate

`git diff --check`, conflict-marker grep, `uv lock --check`, `uv sync
--locked`, `gen_config_surface.py --check`, `vendor_spa.py --check`, ruff
check + format, mypy, `pytest -x -q` (4238 passed, 3 skipped), `mkdocs build
--strict`, and `pre-commit run --all-files` all pass locally.
`scripts/structural_gate.sh` reports no Python source changes and skips.

---


pvliesdonk:
@claude review

PR #1342, opened 2026-09-05
docs(git): state what the remote must permit, and how to read a refusal

### LINKED ISSUE #1337: [Docs]: the git guide covers credentials but not what the remote must permit, nor how to see why a push was refused
### What happened?

Bringing up a managed git vault against a corporate GitLab hit two remote-side refusals in a row. Both are ordinary first-run conditions, neither is covered by the git integration guide, and neither is diagnosable from the log — so each cost a long detour before the actual git error was found.

### Observed

The guide's managed-git setup section gives `GIT_REPO_URL` / `GIT_USERNAME` / `GIT_TOKEN` and a "Provider Username Reference" listing the HTTPS username per provider (`docs/guides/git-integration.md:55-62`, `:360-368`). It does not state what the remote must *permit*.

Two refusals encountered on first bring-up:

1. The default branch was **protected** on the GitLab project, so the token could not push to it. Writes committed locally and accumulated; the `remote` key correctly reported `unsynced`, with `reason: "push_failed"`.
2. No usable token was present for the clone, so `fetch` failed and the GitLab push webhook returned `503 {"error":"pull not applied","reason":"fetch_failed"}` on every delivery.

In both cases the log line carries no cause at the default level. The text that names the actual problem —

```
remote: GitLab: You are not allowed to push code to protected branches on this project.
```

— is reachable only by calling `git_sync(direction="push")` over MCP and reading `PushResult.hint`, or by reproducing the push by hand inside the container. The guide does not mention `git_sync` as the diagnostic route, and its troubleshooting section does not cover a clone that commits but never replicates.

### Expected

The managed-git section states the remote-side prerequisites an operator must satisfy before the vault can replicate — at minimum that the token needs push access to the target branch, and that a protected default branch will refuse it — and points at the one call that surfaces the underlying git error.

### Reproduction

Configure a managed git vault whose remote refuses the push (protect the default branch against the token's identity). Write a note, wait for `push_delay_s`, then read the container log at the default level: the vault is reported unsynced with no cause.

Observed against `4.2.0rc0` (core 6.0.0) and `docs/guides/git-integration.md` at `a4d28b1b`.

### Suspected cause

`[verified: read the guide]` The guide covers credentials (which username, which token variable) but not authorization (what that token must be allowed to do on the remote), and its GitLab section covers only the webhook credentials.

### Open questions / scope

- Whether the same prerequisites want stating for GitHub (branch protection and rulesets refuse a push the same way) and Bitbucket.
- `[unverified]`: the `fetch_failed` webhook path may deserve the same treatment as the push path; that is about the missing log detail rather than the docs, tracked in #1330.

Related: #1330 (the push-side observability gap this compounds).

---

pvliesdonk:
## Closes / Refs

Closes #1337

## What & why

The guide gave `GIT_REPO_URL` / `GIT_USERNAME` / `GIT_TOKEN` and a
per-provider username table. Together those say who the server authenticates
*as*, and nothing about what that identity is allowed to do. Bringing a vault
up against a corporate GitLab hit two remote-side refusals in a row; both are
ordinary first-run conditions, and each cost a long detour before the actual
git error was found.

Two additions:

- **What the remote must permit** (managed-mode section): the token needs push
  access to the tracked branch, and protected branches or rulesets refuse
  independently of the token's own scopes. Where to look on GitLab and GitHub,
  and the symptom both share — writes succeed, `read` serves them back, and
  commits accumulate in a clone that is not replicating.
- **`git_sync(direction="push")` is the diagnostic route** (`git_sync`
  section): it runs a real push and returns the remote's own sentence in
  `push.hint`. With the reason `git push --dry-run` is *not* a substitute: a
  dry-run push negotiates with the remote but never runs its pre-receive
  hooks, which is exactly where a protected branch rejects, so it reports
  success against a remote that would refuse the real push. That also explains
  why a `dry_run` `git_sync` returns `dry_run_unsupported` on the push leg
  rather than a prediction.

The dry-run point is the one worth reading closely: it is the trap that
produced the detour, and the guide previously stated the `dry_run_unsupported`
behavior without saying why it exists.

- **Say anything about what the log now contains.** #1330's fix (PR #1341,
  open) makes the push failure log its cause at `WARNING`, and #1341 carries
  the guide edit describing that. Keeping it out of here means this diff is
  reviewable against `main` on its own, with no claim that is only true once
  another PR lands. The two edits touch different sections of the same file;
  whichever merges second may need a trivial rebase.
- **Enumerate Bitbucket's protection model.** The issue raises it as an open
  question. GitLab and GitHub are the two verified against; the general
  statement covers the rest without inventing per-provider UI paths.
- **Document the `fetch_failed` webhook path.** #1337 flags it `[unverified]`
  and routes it to #1330 as a log-detail question rather than a docs one.

## Local review

      Checked specifically for forward references to #1341's behavior (none),
      and verified the `push.hint` field name against
      `git/types.py:214` and the tool reference rather than from memory.
- [x] `mkdocs build --strict` passes; the new cross-reference anchor resolves.
- [x] No commit carries `!`. Documentation only; no code changed.

## Docs impact

- [ ] `README.md` — the deployment prerequisites belong in the git guide.
- [x] `docs/` site pages — `docs/guides/git-integration.md` and
      `docs/releases/4.2.md`.
- [ ] `docs/design/` — no design decision changed; this documents behavior
      that already existed.
- [ ] Inline docstrings — no code touched.

---

### DIFF
--- a/docs/guides/git-integration.md
+++ b/docs/guides/git-integration.md
@@ -68,6 +68,35 @@ Behavior:
 - If `SOURCE_DIR` is empty at startup, the server clones `GIT_REPO_URL` into it.
 - If `SOURCE_DIR` is already a git repo, the server verifies `origin` matches `GIT_REPO_URL`.
 - Each write tool call is committed (see [What a commit covers](#what-a-commit-covers)); the accumulated commits are pushed after the configured idle delay.
+
+### What the remote must permit
+
+`GIT_USERNAME` and `GIT_TOKEN` decide who the server authenticates *as*. They
+say nothing about what that identity is allowed to do, and a token that
+authenticates perfectly can still be refused at push time. Both conditions
+below are ordinary first-run states rather than mistakes, and neither
+is visible from the credentials.
+
+**The token needs push access to the branch the vault tracks.** A read-only
+token clones and pulls without complaint, so a vault can run for hours,
+committing locally, before anything reveals that nothing is replicating.
+
+**A protected default branch refuses the push.** GitLab protected branches,
+GitHub branch protection, and GitHub rulesets all reject a push that does not
+meet their conditions, whatever the token's own scopes say. Either grant the
+token's identity permission to push to that branch, or point the vault at a
+branch it may write:
+
+- GitLab: Settings → Repository → Protected branches, or use a token whose
+  role is listed under **Allowed to push and merge**.
+- GitHub: Settings → Branches (or Rules → Rulesets); a bypass entry for the
+  app or account behind the token.
+
+The symptom is the same either way: writes succeed, `read` serves them back,
+and the commits accumulate in the clone without reaching the remote. See
+[When the clone stops reaching its remote](#when-the-clone-stops-reaching-its-remote)
+for what that looks like, and the next section for how to read the actual git
+error.
 - Periodic pull uses fast-forward-only updates.
 
 Two mechanisms sit alongside the periodic loop, both described below: a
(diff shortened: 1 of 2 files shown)


pvliesdonk:
@claude review
@codex review

PR #1344, opened 2026-09-06
fix(git): stop advising an unrelated note when a revision read refuses a copy

### LINKED ISSUE #1336: [Bug]: the revision-read copy refusal advises reading an unrelated note
### What happened?

`read(path, revision=)` correctly refuses when git records the note as a *copy* rather than a continuation. But the refusal message then advises reading a different note — the copy source git happened to score — which in the observed case had nothing to do with the content that lived at that path.

The refusal itself is right. The concern is the advice, because the documented next step after a revision read is to write the content back.

### Observed

Sequence on a `4.2.0rc0` git vault:

1. `write("probe/b1.md", ...)` → commit `66b5f12`
2. `delete("probe/b1.md")`
3. `write("probe/b1.md", <unrelated content>)` — a new note reusing the freed name
4. `read("probe/b1.md", revision="66b5f12")`

```
Error: '0-Inbox/_probe42/b1.md' appears as a copy of '0-Inbox/_probe42/t1-revision.md'
rather than as a continuation of it, so the note now at that path did not exist at
that revision. Read '0-Inbox/_probe42/t1-revision.md' at '66b5f12...' instead if the
content you want is the one it was copied from.
```

`t1-revision.md` is an unrelated note the caller never mentioned. What actually lived at `probe/b1.md` at `66b5f12` was the *first* `b1.md`, which the message does not mention. Both notes were short and shared frontmatter scaffolding (`title`, `type`, `description`, an OKF `generated` block), which is apparently enough for git to score them as a copy.

The sibling `A` branch of the same walk ends with "use 'get_history' to find a revision this note actually has"; the `C` branch offers no such fallback.

### Expected

A refusal that names another note is only useful if that note is plausibly what the caller wanted. Where it is an artefact of similarity scoring, pointing at `get_history` — as the neighbouring branch does — carries no risk of the caller restoring unrelated content.

### Reproduction

As the numbered sequence above: two short notes with similar frontmatter, one reusing a freed name.

Observed on `4.2.0rc0` (core 6.0.0).

### Suspected cause

`[verified: read the source]` `_path_at_ref` (`git/query.py:1725`) formats `paths[0]` from git's `C` name-status record straight into the advice sentence. `[unverified]` The likelihood of a spurious `C` record may have risen with #1302, which pinned every walk to `--find-renames=30`; short notes sharing generated frontmatter clear 30% similarity easily. I have not verified that the threshold is what produced this record.

### Open questions / scope

- `[unverified]`: `get_history` and `get_diff` use the same walk and may surface the same copy records, though neither hands the caller content to write back.

---

pvliesdonk:
## Closes / Refs

Closes #1336

## What & why

`read(path, revision=)` refuses correctly when git records the note as a copy rather than a continuation, but the `C` branch of `_path_at_ref` then advised reading the copy source. That source is git's similarity match — in the observed case an unrelated note — and the documented next step after a revision read is to write the content back, so the advice invited restoring unrelated content. The refusal still names what git found (it explains the refusal) and now ends the way the sibling `A` branch already did: use `get_history`.

Decision recorded on #1336 before coding. Not done, deliberately: the rename threshold (`--find-renames=30`) is untouched — the cause of the spurious `C` record is unverified, and lowering it would trade this for missed lineage.

## Self-review 11b0ada..0aba481

Charters: rules, diff, comments, history. Conformance: no normative content. Past-PRs: skipped (single message change; no prior review threads on this branch of `query.py`).
Coverage: full.

- `git/query.py`, `tests/test_git_revision.py`, `docs/releases/4.2.md` — important/verified — the comment, test docstring and release note stated the *cause* of the spurious copy record ("clear the rename threshold easily") as fact; #1336 marks that cause `[unverified]`. Resolved in `0aba481`: all three now state only what was observed.

## Verification

- `uv run pytest -x -q`: 4260 passed, 3 skipped (pre-amend run; the amend changed prose only, `tests/test_git_revision.py -k copy_refusal` re-run: 1 passed).
- `ruff check --fix` → `ruff format` → `ruff format --check`: clean. `mypy src/ tests/`: clean.
- `scripts/structural_gate.sh`: 0 violations. `pre-commit run --all-files`: all passed (Vale included).

## Docs impact

- [x] `docs/releases/4.2.md` — paragraph in "History that describes the right note".
- [ ] `README.md` / `docs/tools/index.md` — no contract change; the refusal list there does not quote the message. `grep -rn "copied from\|appears as a copy" docs/ README.md` finds no mirror.
- [ ] `docs/design/` — no mirror of the message.
- [x] Inline comment on the branch says why.

No commit carries `!`.

### DIFF
--- a/docs/releases/4.2.md
+++ b/docs/releases/4.2.md
@@ -110,6 +110,17 @@ The end state is uniform: `get_history`, `get_diff`, and the new
 following, the same threshold, and the same lineage boundary at the note's
 creation.
 
+One refusal message changed after the release candidate. `read(path,
+revision=)` correctly refuses when git records the note as a copy rather
+than a continuation, but the refusal then named the copy source as the thing
+to read
+([#1336](https://github.com/pvliesdonk/markdown-vault-mcp/issues/1336)).
+That source is git's similarity match (in the case that surfaced this, an
+unrelated note that merely shared frontmatter scaffolding), and the
+documented next step after a revision read is to write the content back. The
+refusal still names what git found, and now points at `get_history` the way
+the neighbouring refusal already did.
+
 ## One commit per tool call
 
 This one arrived from the field.
--- a/src/markdown_vault_mcp/git/query.py
+++ b/src/markdown_vault_mcp/git/query.py
@@ -1722,11 +1722,20 @@ def _path_at_ref(raw: str, cur_rel: str, ref: str) -> str:
                 "'get_history' to find a revision this note actually has."
             )
         elif status.startswith("C") and paths[-1] == tracked:
+            # The copy source is git's similarity match, not a note the caller
+            # asked about: in the case that surfaced this (#1336) it was an
+            # unrelated note that merely shared frontmatter scaffolding.
+            # Naming it as the thing to read invites restoring unrelated
+            # content, and the documented next step after a revision read is
+            # to write it back, so this branch points at 'get_history' the
+            # way the 'A' branch above does. The source is still named,
+            # because it is what git said and it explains the refusal.
             raise ValueError(
                 f"{tracked!r} appears as a copy of {paths[0]!r} rather than as a "
                 "continuation of it, so the note now at that path did not exist "
-                f"at that revision. Read {paths[0]!r} at {ref!r} instead if the "
-                "content you want is the one it was copied from."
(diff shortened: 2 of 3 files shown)


pvliesdonk:
@claude review
@codex review

PR #1346, opened 2026-09-06
fix(git): report a configured commit claim the token does not carry

### LINKED ISSUE #1331: [Bug]: a configured GIT_COMMIT_*_CLAIM that the token does not carry falls back silently
### What happened?

`MARKDOWN_VAULT_MCP_GIT_COMMIT_NAME_CLAIM` and `_GIT_COMMIT_EMAIL_CLAIM` are configured on a `4.2.0rc0` OIDC deployment, but every commit is authored by the static fallback identity. Nothing at any log level says the configured claims are not being applied.

This is not a regression of #1218 — that fix (resolution at the tool edge) is working. The claims simply are not present in the token, and that outcome is indistinguishable from success.

### Observed

Compose sets:

```
MARKDOWN_VAULT_MCP_GIT_COMMIT_NAME:        Work Vault Agent
MARKDOWN_VAULT_MCP_GIT_COMMIT_EMAIL:       vault-agent@<host>
MARKDOWN_VAULT_MCP_GIT_COMMIT_NAME_CLAIM:  name
MARKDOWN_VAULT_MCP_GIT_COMMIT_EMAIL_CLAIM: email
```

Every commit from an authenticated MCP write:

```
author: Work Vault Agent <vault-agent@<host>>
```

The OKF provenance stamped by the *same writes* records the authenticated subject:

```yaml
generated:
  by: human:9d8a647f-b95b-4807-bc24-ea67c68018a0
```

So one `Principal` resolved a subject (`kind="human"`) while `display_name` and `email` fell back — silently. Startup logs no warning; no per-write log mentions the claims.

### Expected

When a claim key is configured but absent from the resolved principal, that is worth one line. The operator configured attribution deliberately and has no signal that it is inert; the failure looks exactly like a deployment that never configured it.

### Reproduction

1. Set `GIT_COMMIT_NAME_CLAIM` / `_EMAIL_CLAIM` to claims the IdP does not put in the **access token**.
2. Authenticate and write a note.
3. Inspect the commit author and the container log.

Observed on `4.2.0rc0` (core 6.0.0), auth mode `multi`, Authelia 4.39.20.

### Suspected cause

`[verified: read the source]` `resolve_mcp_principal` (`_identity.py:143`) reads the configured keys via `_claim_value` and returns `None` for an absent claim, with no logging. `GitWriteStrategy` then falls back to the static identity. The one existing startup warning (`git/strategy.py:487`) is explicitly suppressed when a claim *is* configured (`and not self._commit_name_claim`), so configuring a claim removes the only nearby diagnostic.

`[verified: read the IdP config]` In this deployment the claims are absent from the access token: none of Authelia's `claims_policies` defines an `access_token:` section, and the OIDC client has no `claims_policy` at all, so the access token carries only standard JWT-profile claims plus `sub`. This explains this instance, not the general case.

### Open questions / scope

- Whether the same silence applies when a claim is present but empty.
- `[unverified]`: the OKF provenance path resolves from the same `Principal` and may want the same signal.

Related: #1218 (closed — the thread-context cause), #565 (the feature).

---

pvliesdonk:
## Closes / Refs

Closes #1331

## What & why

On an OIDC deployment with `GIT_COMMIT_NAME_CLAIM` / `_EMAIL_CLAIM` configured, every commit was authored by the static fallback identity, and nothing at any level said why: the identity provider simply did not put those claims in the access token. The fallback is correct; what was missing is any signal that the operator's configured attribution is inert. The one nearby startup warning (`git/strategy.py:487`) is suppressed exactly when a claim *is* configured.

`resolve_mcp_principal` now resolves each configured claim through `_resolve_claim`, which logs `git_commit_claim_unusable field=… claim=… state=absent|empty|not_a_string` once per process per `(field, key)` — per field, so both fields configured to the same claim each get their line (the open finding from the closed #1341's review) — and only for an authenticated caller: with no token there is nothing to be missing. `configure_identity_claims` clears the memory, so a reconfiguration gets a fresh answer. Decision recorded on #1331 before coding.

Not done: the OKF provenance path (it stamps the subject via `get_subject()`, which is unaffected by the claim keys — that is why the issue saw `human:<subject>` next to a static commit author); the startup warning keeps its condition.

## Probe (`get_subject`/`get_claims` patched, three writes then an unauthenticated one)

```
WARNING markdown_vault_mcp._identity: git_commit_claim_unusable field=display_name claim=name state=absent detail=commits fall back to the static commit identity
WARNING markdown_vault_mcp._identity: git_commit_claim_unusable field=email claim=email state=empty detail=commits fall back to the static commit identity
principal: Principal(subject='9d8a647f', display_name=None, email=None, kind='human')
unauthenticated: Principal(subject=None, display_name=None, email=None, kind='local')
```

Two lines for three writes; none for the unauthenticated caller.

## Self-review 11b0ada..HEAD

Charters: rules, diff, comments, history. Conformance: no normative content. Past-PRs: the closed #1341's one open finding (dedup keyed on the claim alone) is fixed here with a test that configures both fields to the same key.
Coverage: full.

- No blockers or importants survived refutation.
- Minor, justified: two concurrent first writes can race on the `(field, key)` set and log the same pair twice. Benign duplicate; a lock for a once-per-process diagnostic is not worth its cost.

## Verification

- `uv run pytest -q`: 4252 passed, 19 skipped — 16 of the skips were `tests/test_config_wizard_smoke.py` skipping because the local `site/` was stale against the regenerated `wizard-spec.json`; after `mkdocs build` those 16 pass. The other 3 skips are the usual Python 3.13 symlink gates.
- ruff check/format, `mypy src/ tests/`, `scripts/structural_gate.sh` (0 violations), `pre-commit run --all-files` (Vale included), `scripts/gen_config_surface.py --check`: clean.

## Docs impact

- [x] `docs/design/design.md` — identity section: the diagnostic, its dedup key, and why it lives at the resolution site.
- [x] `docs/releases/4.2.md` — paragraph after the #1218 fix.
- [x] Config help for `GIT_COMMIT_NAME_CLAIM` gained one sentence; the generated surfaces (`docs/configuration.md`, `wizard-spec.json`, `.env.example`, `examples/*.env`, `packaging/env.example`, `server.json` descriptions) are regenerated from it. No version stamp changed.
- [ ] `README.md` — no config or tool surface changed.
- [x] Docstrings: `_resolve_claim`, `_claim_state`, `resolve_mcp_principal`.

No commit carries `!`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -205,7 +205,7 @@
 # MARKDOWN_VAULT_MCP_GIT_COMMIT_NAME=markdown-vault-mcp
 # Git committer email for auto-commits.
 # MARKDOWN_VAULT_MCP_GIT_COMMIT_EMAIL=noreply@markdown-vault-mcp
-# OIDC claim key used as the commit author name (e.g. name); overrides GIT_COMMIT_NAME per request when an OIDC token is present. The claim is resolved when the tool call arrives and carried to the background commit, so it applies on every write.
+# OIDC claim key used as the commit author name (e.g. name); overrides GIT_COMMIT_NAME per request when an OIDC token is present. The claim is resolved when the tool call arrives and carried to the background commit, so it applies on every write. A configured claim the token does not carry is reported once at WARNING and the static identity is used.
 # MARKDOWN_VAULT_MCP_GIT_COMMIT_NAME_CLAIM=
 # OIDC claim key used as the commit author email (e.g. email); overrides GIT_COMMIT_EMAIL per request when an OIDC token is present. Resolved and carried the same way as the name claim.
 # MARKDOWN_VAULT_MCP_GIT_COMMIT_EMAIL_CLAIM=
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -374,7 +374,7 @@ starting points live under `examples/`: `obsidian-readonly.env`,
 | `MARKDOWN_VAULT_MCP_GIT_PUSH_DELAY_S` | `30.0` | No | Seconds of write-idle time before pushing; 0 pushes only on shutdown. |
 | `MARKDOWN_VAULT_MCP_GIT_COMMIT_NAME` | `markdown-vault-mcp` | No | Git committer name for auto-commits; set this in Docker where git config user.name is empty. |
 | `MARKDOWN_VAULT_MCP_GIT_COMMIT_EMAIL` | `noreply@markdown-vault-mcp` | No | Git committer email for auto-commits. |
-| `MARKDOWN_VAULT_MCP_GIT_COMMIT_NAME_CLAIM` | (none) | No | OIDC claim key used as the commit author name (such as name); overrides GIT_COMMIT_NAME per request when an OIDC token is present. The claim is resolved when the tool call arrives and carried to the background commit, so it applies on every write. |
+| `MARKDOWN_VAULT_MCP_GIT_COMMIT_NAME_CLAIM` | (none) | No | OIDC claim key used as the commit author name (such as name); overrides GIT_COMMIT_NAME per request when an OIDC token is present. The claim is resolved when the tool call arrives and carried to the background commit, so it applies on every write. A configured claim the token does not carry is reported once at WARNING and the static identity is used. |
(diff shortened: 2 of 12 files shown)


pvliesdonk:
@claude review
@codex review

PR #1364, opened 2026-09-06
fix(rename): say when update_links does not apply to an attachment

### LINKED ISSUE #1338: [Bug]: renaming an attachment ignores update_links, leaving embeds dangling
### What happened?

`rename(old, new, update_links=True)` on an **attachment** returns
`updated_links: 0` and rewrites nothing. A note embedding that attachment
keeps pointing at the old path, which no longer exists.

This has always been the behaviour, but it was invisible until now: a
wikilink to an attachment never resolved in the first place (#1333), so
there was no working link for a rename to break. With #1333 fixed, the
embed resolves, and renaming the attachment silently un-resolves it.

### Observed

Vault with `attachment_extensions: [png]`, `Images/pic.png` present, and
`note.md` containing `![[Images/pic.png]]`:

```python
r = v.writer.rename("Images/pic.png", "Images/renamed.png", update_links=True)
# rename result updated_links: 0
# note body after: "# N\n\n![[Images/pic.png]]\n"
```

The file moved; the embed did not. `get_broken_links` now reports it,
correctly — nothing in the vault is at `Images/pic.png` any more.

The note branch of the same call does rewrite links, so the two halves of
one documented parameter behave differently with no signal to the caller.

### Expected

Unclear which of two things, and that is the open question below: either
`update_links=True` rewrites links to a renamed attachment the way it does
for a note, or it says it cannot. Silently reporting `updated_links: 0` for
a call that did have links to update is the part that is wrong either way.

### Reproduction

1. Vault with `png` in `MARKDOWN_VAULT_MCP_ATTACHMENT_EXTENSIONS` and an
   attachment `Images/pic.png`.
2. A note containing `![[Images/pic.png]]`.
3. `rename("Images/pic.png", "Images/renamed.png", update_links=True)`.
4. Read the note — the embed still names the old path.

Observed on `main` at `9e93b546` plus the #1333 fix, with the probe above.

### Suspected cause

`[verified: read the source and ran the probe above]` `DocumentManager.rename`
branches on `is_note(old_path)` (`managers/document.py:1102`). The note
branch collects `self._fts.get_backlinks(old_path)` before the move and
rewrites them; the attachment branch (`:1136`) calls
`self._artifacts.move(old_path, new_path, if_match)` and never reads
`update_links` at all.

`[unverified]` Whether `move_folder`, which moves attachments as part of a
subtree, has the same gap. I have not verified it.

### Open questions / scope

- Whether an attachment rename *should* rewrite embeds, or whether the
  honest answer is that `update_links` does not apply to attachments and the
  call should say so.
- `[unverified]`: `move_folder` moves attachments too and may share this
  code path.

Related: #1333 (the fix that makes this reachable).

---

pvliesdonk:
## Closes / Refs

Closes #1338. Refs #1333 (the fix that made this reachable), #1359 (attachments as graph nodes, v5).

## What & why

`rename(old, new, update_links=True)` on an attachment answered `updated_links: 0` and nothing else. The attachment branch of `DocumentManager.rename` never read the flag, so — in the issue's words — "the two halves of one documented parameter behave differently with no signal to the caller". Since #1360 the link graph is notes-only, so the `0` is honest: no reference to an attachment is a tracked link, and there is nothing to rewrite. The silence was the bug.

**Decision (owner, 2026-09-05, quoted on the issue):** the attachment branch of `rename` *says* `update_links` does not apply instead of a silent `0`; rewriting embeds on attachment rename stays #1359.

- `RenameResult` gains `hint: str | None = None` (`types.py`). The attachment branch sets it only when `update_links=True` (`managers/document.py`): references to attachments are not tracked as links, so none were rewritten; notes embedding the old path still name it.
- The MCP `rename` result carries the `hint` key only when set — absent, not null — the same shape as the warn-only `remote` key and `PushResult.hint`.
- `updated_links` stays `0` for attachments; a note rename, or an attachment rename without the flag, has no `hint`.
- `move_folder` is unaffected: it has no `update_links` parameter and rewrites through its `.md`-only `md_map` (`_plan_folder_move`); attachments inside the moved prefix have had no tracked links since #1360. This closes the `[unverified]` in the report.

**Client-surface budget.** The `rename` description's added clause pushed the reviewed total to 58,064 > 58,000 (`tests/test_client_surface_budget.py`; input schemas sit at 27,594/27,600, so the `update_links` arg description could not grow). The description now carries "update_links does not apply (attachment references are not tracked as links)" and drops "— omitting this leaves those links broken", which the `update_links` arg description still says verbatim (both clauses came from #214's `05a757ca`; one of two duplicated warnings moved, none left the surface). `Returns:` text does not count toward the description (FastMCP keeps only the first text section) and is where the `hint` is documented. Total after: 57,991.

## Probe (library, the issue's exact inputs, `attachment_extensions=["png"]`)

```
rename result: RenameResult(old_path='Images/pic.png', new_path='Images/renamed.png', updated_links=0, hint='update_links does not apply to attachments: references to attachments are not tracked as links, so none were rewritten. Notes embedding the old path still name it; edit them to point at the new path.')
note body after: '# N\n\n![[Images/pic.png]]\n'
no-flag: RenameResult(old_path='Images/renamed.png', new_path='Images/pic.png', updated_links=0, hint=None)
```

Note branch, for contrast (`other.md` holds `[[note]]`):

```
note rename: RenameResult(old_path='note.md', new_path='moved.md', updated_links=1, hint=None)
other body after: 'See [[moved]].\n'
```

- Rewrite embeds when an attachment is renamed, or resolve/backlink attachments: #1359.
- Change `updated_links` for attachments, or add a `hint` to `move_folder` (it has no flag to not-apply).
- Bump `INDEX_SEMANTICS_VERSION`: no stored row changes.

## Self-review 7aabfe66..9b01e083

Charters: rules, diff, comments, history. Conformance: no normative content. Past-PRs: not walked (the closed #1341 bundle's #1338 hunks were not consulted; the decision on the issue was the input).
Coverage: full. Comments and history ran as parallel read-only agents; rules and the diff charter I walked myself.

- `facets/writer.py` `WriterFacet.rename` docstring — minor/verified — still described `update_links` with no attachment carve-out; pre-existing since #1360 but the same class as the docstrings this PR changes, so swept here. Resolved in `9b01e083`.
- History: the deleted `test_update_links_ignored_for_attachments` (`68bb8870`) pinned observed behaviour, not a decision ("verifies update_links=True is silently ignored"); its two assertions survive in the replacement test. No prior `hint` on a write result was ever reverted. The `rename` description had never been budget-trimmed before.
- Rules, diff: clean.

## Verification

- TDD: three library tests and one tool test failed first (`AttributeError: 'RenameResult' object has no attribute 'hint'`, `KeyError: 'hint'`); the two "omits hint" tool tests guard the `None`-stripping and would fail on a bare `asdict`.
- `uv run pytest -x -q --cov=src/markdown_vault_mcp`: 4392 passed, 3 skipped; every changed line covered (`writer.py` 100%, `types.py` 100%, `document.py` misses are pre-existing lines).
- ruff check/format, mypy (231 files), pre-commit `--all-files` (Vale clean after two rewrites), `scripts/structural_gate.sh` (0 violations, 100%), `mkdocs build --strict` exit 0.

### DIFF
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -1829,6 +1829,7 @@ class RenameResult:
     old_path: str
     new_path: str
     updated_links: int = 0  # number of source docs updated (update_links=True)
+    hint: str | None = None  # why nothing was rewritten (attachment, #1338)
 
 # --- Index types ---
 
@@ -2125,8 +2126,9 @@ itself resolves `[[Figure 1.png]]` to the file and shows it in its graph
 ([`reference/obsidian-markdown.md`](reference/obsidian-markdown.md),
 "Extension, non-markdown targets, embeds"); making attachments graph nodes
 here — resolution, `exists`, backlinks, rename-rewrite — is the v5 feature
-#1359. Until then `rename` of an attachment reports `updated_links: 0`
-with no signal; making it say that `update_links` does not apply is #1338.
+#1359. Until then `rename` of an attachment with `update_links=True`
+reports `updated_links: 0` with a `hint` saying the flag does not apply
+(#1338).
 
 Two consequences follow. The allowlist is now an input to what a note's
 stored rows are, so it is recorded as build provenance alongside
@@ -2483,7 +2485,13 @@ count reflects source documents successfully rewritten. Link style is
 preserved: vault-root-relative links are rewritten as vault-root-relative;
 source-directory-relative links (such as `../notes/target.md`) are rewritten
 with the correct new relative path from the source file's directory.
-`update_links` is silently ignored for attachments (non-`.md` files).
+`update_links` does not apply to attachments (non-`.md` files): the link
+graph is notes-only, so no reference to an attachment is tracked and there
+is nothing to rewrite. Passing the flag for an attachment sets
+`RenameResult.hint` saying so; `updated_links` stays `0`, and the MCP tool
+carries the `hint` key only when it is set (#1338). `move_folder` has no
+such flag and rewrites through its `.md`-only path map, so it needs no
+equivalent. Rewriting embeds on attachment rename is the v5 feature #1359.
 
 **`move_folder()` behavior**: a dedicated tool (not an overload of `rename`)
 that moves an entire folder subtree to a new prefix in one call. It is the
--- a/docs/releases/4.2.md
+++ b/docs/releases/4.2.md
@@ -407,6 +407,18 @@ What this deliberately does not do is make attachments part of the graph:
 rewrites the embeds pointing at the renamed file are the v5 feature
 [#1359](https://github.com/pvliesdonk/markdown-vault-mcp/issues/1359).
 
+**Renaming an attachment says when `update_links` does not apply.** The
+attachment branch of `rename` never read the flag, so a call that asked
+for a rewrite answered `updated_links: 0` and nothing else, and "the two
+halves of one documented parameter behave differently with no signal to
+the caller"
+([#1338](https://github.com/pvliesdonk/markdown-vault-mcp/issues/1338)).
+The count was honest; the silence was the bug. The result now carries a
+`hint` when `update_links=true` is passed for an attachment: references to
+attachments are not tracked as links, so nothing was rewritten and the
+notes embedding the old path still name it. A note rename, or an attachment
+rename without the flag, carries no `hint` key at all.
+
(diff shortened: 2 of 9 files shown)


pvliesdonk:
@claude review

PR #1371, opened 2026-09-06
fix(okf): stale on the stale_after date, and a bare verified mapping is one entry

### LINKED ISSUE #1357: [Bug]: okf.py departs from OKF v0.2 on the staleness boundary and the bare-mapping `verified` shorthand
### What happened?

Two places where the OKF read layer and the OKF v0.2 field reference disagree, surfaced while reconciling `scripts/check_references.py` with `okf.py` on #1356. `[verified: read both; the field reference quoted verbatim from the open-knowledge-format plugin's references/field-reference.md]`

- **Staleness boundary.** `derive_stale` returns true iff `stale_after` is strictly before today, and `docs/design/design.md` (§ OKF read semantics) says the same. The field reference says: "`stale_after: 2026-09-23` — absolute date. Stale when `today >= stale_after`." So on the `stale_after` date itself the server reports the note as current while a spec-following consumer reports it stale.
- **`verified` shorthand.** `derive_trust_tier` returns `unverified` unless `verified` is a list. The field reference says: "A single verifier may be written as a bare mapping without the list dash." and "Consumers must treat a bare mapping as a one-element list." A note with `verified: {by: human:x, at: ...}` is human-reviewed per spec and unverified per this server.

### Observed

- `src/markdown_vault_mcp/okf.py`, `derive_stale` (`raw < today`) and `derive_trust_tier` (`if not isinstance(verified, list): return TRUST_UNVERIFIED`).
- `docs/design/design.md`, the OKF read-semantics paragraph ("`stale_after` strictly before today").

Observed on `main` at ec78d51.

### Expected

The read layer follows the spec on both points, or the design doc records why it deliberately does not. I have not verified whether the strict-before boundary was a deliberate choice; `docs/design/okf.md` § 3 states it without a rationale.

### Open questions

- `[unverified]`: `okf_validate` may report staleness with the same boundary and share this code path.

pvliesdonk:
## Closes / Refs

Closes #1357.

## What & why

The OKF v0.2 field reference says a note is "stale when `today >= stale_after`" and that consumers "must treat a bare mapping as a one-element list" for `verified`. The read layer did neither, and neither departure was deliberate — #968's body and commit say only "date comparison on `stale_after`", and `docs/design/okf.md` § 3 states strict-before with no rationale.

**Decision (on the issue):** follow the spec on both. The `verified` half is four sites, not two, and one of them loses data: `derive_trust_tier` (bare mapping read as `unverified`), `append_okf_verification` (a bare mapping was *replaced* by the new entry, dropping the existing attestation), the `okf_verify` tool's `verified_count` (bare mapping counted as 0), and the `derive_stale` boundary. One helper, `okf.verified_entries`, does the coercion so the readers cannot drift again. `okf_validate` computes no staleness (the decision comment said otherwise; corrected there) — the consumers are the `stale` annotation, the `stale` filter, `stats.okf.stale_count` and ranking, all through `derive_stale`, so there is no second boundary.

No `INDEX_SEMANTICS_VERSION` bump: staleness and tier are derived at annotation time from the stored frontmatter blob; no stored row changes.

## Breaking-change classification, made deliberately

`fix:`, not `fix!:`. AGENTS.md's refinement 1 ("if the old behaviour is gone, treat it as breaking") is aimed at a tool whose *contract* changes; this corrects a derived advisory field to the value the spec it implements defines, the same class as #1333 and #1334 changing what `get_broken_links` returns in this cycle, both `fix:`. What an operator sees: a note whose `stale_after` is today reads `stale: true` from upgrade day (the Upgrading section says so). If you read it the other way, retitling is enough.

## Conformance (executed, `today = 2026-09-06`)

| `stale_after` | stale |
|---|---|
| `2026-09-05` / `2026-09-06` / `2026-09-07` (date or string) | true / true / false |

| `verified` | tier | `okf_verify` result |
|---|---|---|
| `{by: human:alice, at: …}` (bare mapping) | human-reviewed | `[alice, peter]`, count 2 (was `[peter]`, count 1) |
| `{by: process:ci}` | machine-confirmed | prior kept, count 2 |
| `[…]`, scalar, absent | unchanged | unchanged |

- Change how `verified: {}` / `[{}]` read (machine-confirmed, pre-existing for the list form; the docstring now says "entries that are not mappings are ignored", which is what the code does).
- Touch `stale_after` timezone semantics (server-local date; the open question in `okf.md` § 10 stands).

## Self-review c78cf215..c87017db

Charters: rules, diff, comments, history, conformance (executed matrices above). Coverage: full; one read-only agent ran rules/comments/history/conformance, the diff charter I walked myself.

- AGENTS.md breaking-change policy — important — the commit carried no `!` and no recorded reason; the classification above is the resolution, written down rather than left implicit.
- `docs/releases/4.2.md` — minor/verified — claimed `okf_validate` derives staleness; it does not (`stats` does). Resolved in `c87017db`, and the issue's decision comment corrected.
- `examples/okf/prompts/*` — minor — two prompt texts still said "in the past"; now "has arrived". Resolved in `c87017db`.
- `derive_trust_tier` docstring — minor — "malformed entries are ignored" overstated (only non-mapping entries are). Resolved. Test renamed to what it pins (`test_scalar_verified_is_replaced`).
- History: `git log -S` on both departures points only at #968; nothing was reverted.

## Verification

- TDD: six tests failed first (boundary ×2, bare-mapping tier ×2, append, tool count).
- `uv run pytest --cov`: 4527 passed, 3 skipped; `okf.py` and `_server_tools/writer.py` at 100%.
- ruff, mypy (232 files), pre-commit `--all-files` (Vale), structural gate 100%, `mkdocs build --strict` exit 0, client-surface budget test (the three "reached" edits fit: +2 chars in input schemas).

### DIFF
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -1382,10 +1382,12 @@ what is implemented today:
   `_server_tools/_common.py` mirroring `attach_conventions`): when active,
   `search` hits, whole-document `read`s, and `get_context` carry an `okf`
   key — `type` (when declared), `status` (absent ⇒ `stable` per spec),
-  `stale` (`stale_after` strictly before today, server-local date), and
+  `stale` (today on or after `stale_after`, server-local date), and
   `trust_tier` (`human-reviewed` if any `verified[].by` has the `human:`
   prefix, else `machine-confirmed` if `verified` is non-empty, else
-  `unverified`). Reads carry the full `sources` list; search hits carry
+  `unverified`; a bare `verified` mapping is a one-element list, as the
+  spec's single-verifier shorthand requires, at every site that reads
+  the field — tier, the `okf_verify` append and its count, #1357). Reads carry the full `sources` list; search hits carry
   `sources_count`. Section reads omit the key (no frontmatter to derive
   from). When inactive, no payload changes at all — the non-OKF baseline
   is byte-identical.
--- a/docs/design/okf.md
+++ b/docs/design/okf.md
@@ -153,11 +153,14 @@ server's ranking math — layer 1 surfaces signal and lets the agent judge.
 **Computed per-note properties** (derived at annotation time from the stored
 frontmatter blob; not persisted):
 
-- `staleness`: `stale` iff `stale_after` < today (date-only comparison,
-  server-local date); absent field → not stale.
+- `staleness`: `stale` iff today ≥ `stale_after` (the spec's rule, date-only
+  comparison, server-local date — the named date is the first stale day,
+  #1357); absent field → not stale.
 - `trust_tier`: `human-reviewed` if any `verified[].by` has the `human:`
   prefix; else `machine-confirmed` if `verified` is non-empty (all
-  verifiers non-human); else `unverified`. Exact tier algorithm lives in
+  verifiers non-human); else `unverified`. A bare `verified` mapping is
+  read as a one-element list (spec shorthand; `okf.verified_entries` is the
+  one reader, #1357). Exact tier algorithm lives in
   the field-mapping table (§1 spec-stability caveat) with table-driven
   tests.
 - `okf.status`: the raw `status` value, defaulted to `stable` when absent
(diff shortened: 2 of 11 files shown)


pvliesdonk:
@claude review

@codex review

PR #1375, opened 2026-09-07
fix(okf): stale_after as an instant with an offset, the bare date kept

### LINKED ISSUE #1373: [Bug]: stale_after is compared as a date; OKF v0.2 (amended 2026-08-21) makes it an instant, and a quoted conformant value is ignored
### What happened?

`derive_stale` compares `stale_after` as a date. The OKF v0.2 spec, since its in-place amendment of 2026-08-21, defines `stale_after` as "an absolute instant. A concept is stale when `now >= stale_after`", with an explicit UTC offset, and asks consumers to be strict about values that have none. #1357 aligned the boundary with the July text (`today >= stale_after`, a date), which the amendment superseded before #1357 was filed.

### Observed

`[verified: direct calls to derive_stale on main at 2fcfa994, today = 2026-09-23]`

| `stale_after` as written | parsed by YAML as | reported stale on 2026-09-23 | under `now >= stale_after` |
|---|---|---|---|
| `2026-09-23T00:00:00Z` | datetime | true | true from 00:00Z |
| `'2026-09-23T00:00:00Z'` (quoted) | str | **false** (treated as absent) | true |
| `2026-09-24T00:30:00+02:00` (= 2026-09-23T22:30Z) | datetime | **false** | true from 22:30Z on the 23rd |
| `2026-09-23T12:00:00+02:00` | datetime | true all day | true from 10:00Z |
| `2026-09-23` (bare date) | date | true | a strict consumer ignores it |

So a spec-conformant quoted instant is silently not stale; an aware instant loses its offset (`.date()` on the naive-local value); and a bare date, which the amendment's reference agent ignores, is honoured here.

### Expected

Either the instant rule, with an offset-aware comparison and the quoted string form accepted, or a recorded decision that this server stays lenient on bare dates (an Obsidian author writes `2026-12-31`, and the upstream issue #242 that motivated the amendment is still open) while still reading a conformant instant correctly. Silently ignoring the quoted conformant form is wrong either way.

### Reproduction

A note with `stale_after: '2026-01-01T00:00:00Z'` on an OKF bundle; `read` reports `okf.stale: false`.

### Suspected cause

`[verified: read the source]` `derive_stale` handles `date`, `datetime` (via `.date()`, discarding tzinfo) and `date.fromisoformat` on strings, which rejects a datetime string. Sources and quotes: `docs/design/reference/okf-v0.2.md`, § "Lifecycle" and § "Where this project departs".

### Open questions / scope

- The design question `docs/design/okf.md` § 10 leaves open ("`stale_after` timezone semantics") is now answered by the spec: an explicit offset on the value, compared against `now`.
- Related: the write-side stamps (dates where the spec wants datetimes), filed alongside; #1357 (the date-boundary fix this supersedes in part).

---

pvliesdonk:
## Closes / Refs

Closes #1373. Refs #1357 (the date-boundary fix this extends), #1372 (the write side, next PR), the reference page `docs/design/reference/okf-v0.2.md` (#1374).

## What & why

OKF v0.2 was amended in place on 2026-08-21: `stale_after` is "an absolute instant … stale when `now >= stale_after`", with an explicit offset, and the amendment asks consumers to ignore an offset-less value. #1357 had aligned the reader with the *July* text (a date, `today >= stale_after`), and a spec-conformant quoted instant (`'2026-01-01T00:00:00Z'`) was silently not stale.

**Decision (on the issue):** honour both texts, strict only where neither allows the value.

| `stale_after` | rule |
|---|---|
| bare date (YAML date or `YYYY-MM-DD` string) | stale from that server-local day on — the July rule, kept deliberately (Obsidian authors write dates; upstream #242 is still open) |
| instant with an offset (YAML timestamp, or a quoted ISO string, `Z` included) | stale when `now >= stale_after`, compared as instants |
| datetime without an offset | ignored, one debug line — allowed by neither text |

`derive_stale` takes an aware `now` instead of a date; `derive_annotation` / `matches_okf_filters` take `now` (no external caller passed `today`). No `INDEX_SEMANTICS_VERSION` bump: derived at annotation time. `docs/design/okf.md` § 10's open "timezone semantics" question closes with this.

## Conformance (executed by the review, `now = 2026-08-07T10:00+02:00` = 08:00Z)

| input | stale |
|---|---|
| `date(2026,8,7)` / `"2026-08-07"` / `date(2026,8,8)` | true / true / false |
| `"2026-08-07T08:00:00Z"` / `"…08:00:01Z"` | true / false |
| `"2026-08-07T02:00:00-08:00"` (= 10:00Z, later than now) | false — the offset decides, not the calendar day |
| `"2026-08-07 08:00:00Z"` (space separator, what PyYAML re-dumps) | true |
| `"2026-08-07T08:00:00"` / naive `datetime` | false (ignored) |
| PyYAML `stale_after: 2026-08-07T08:00:00Z` → aware `datetime` | true |

- Write stamps as datetimes: #1372, next.
- Flag an offset-less datetime in `okf_validate`: it is ignored with a debug line; a finding type is not in this issue.
- Change the bundle's own reference pages, whose `stale_after` stays a calendar date because the template-owned checker requires one (fastmcp-server-template#603).

## Self-review 23d97ede..12e323d7

Charters: rules (incl. the logging standard for the two `logger.debug` lines), diff, comments, history, conformance (the matrix above, plus a YAML round trip and a JSON index round trip proving the SQL-backed stats/filter path agrees with the live path). Coverage: full; one read-only agent, the diff charter walked by me as well.

- `okf.py` `_parse_stale_after` — structural gate — eight returns; the offset check split into `_aware_or_none`. Resolved before the first push.
- `docs/releases/4.2.md` upgrade bullet — minor — said listings "may gain entries" but not that a note with an offset-less datetime (previously truncated to its date) now loses `stale: true`. Resolved in `12e323d7`.
- Reference page observation stamp still `2026-09-06` under a bullet describing today's code. Resolved in `12e323d7`.
- Recorded intent: #1371's three boundary cases survive verbatim; the one removed case (naive datetime → stale) is the deliberate flip. `20260101` / `2026-W32-5` strings, which 3.11's relaxed `date.fromisoformat` used to accept by accident, are now ignored — the decision scopes the string form to `YYYY-MM-DD`.

## Verification

- TDD: the extended parametrisation failed first (signature, then eight value rows).
- `uv run pytest --cov`: 4536 passed, 3 skipped; `okf.py` 100%.
- ruff, mypy (232 files), pre-commit `--all-files` (Vale), structural gate 100%, `mkdocs build --strict` 0, `check_references.py` 0.

### DIFF
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -1383,7 +1383,9 @@ the spec itself says, dated and sourced, in
   `_server_tools/_common.py` mirroring `attach_conventions`): when active,
   `search` hits, whole-document `read`s, and `get_context` carry an `okf`
   key — `type` (when declared), `status` (absent ⇒ `stable` per spec),
-  `stale` (today on or after `stale_after`, server-local date), and
+  `stale` (a bare `stale_after` date: the server-local day has reached it;
+  an instant with an offset: `now >= stale_after`; an offset-less datetime
+  is ignored — #1357, #1373), and
   `trust_tier` (`human-reviewed` if any `verified[].by` has the `human:`
   prefix, else `machine-confirmed` if `verified` is non-empty, else
   `unverified`; a bare `verified` mapping is a one-element list, as the
--- a/docs/design/okf.md
+++ b/docs/design/okf.md
@@ -64,7 +64,8 @@ metadata but attaches no meaning to `type`, `status`, `stale_after`, `sources`,
   Actor convention `human:<id>` / `process:<id>` / `<tool>/<version>`.
   Derived tiers: **unverified → machine-confirmed → human-reviewed**.
 - **Lifecycle:** `status: draft|stable|deprecated` (default `stable`),
-  `stale_after: YYYY-MM-DD`.
+  `stale_after`: a `YYYY-MM-DD` date (the July 2026 text) or an ISO 8601
+  instant with an offset (the text since 2026-08-21).
 - **Bundle marker:** `okf_version: "0.2"` allowed only in the bundle-root
   `index.md` frontmatter.
 - **Reserved files:** `index.md` (progressive-disclosure directory listing)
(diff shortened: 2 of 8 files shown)


pvliesdonk:
@claude review

PR #1376, opened 2026-09-07
fix(okf): stamp generated.at and verified[].at as UTC instants

### LINKED ISSUE #1372: [Bug]: OKF write stamps generated.at and verified[].at as bare dates; the spec requires ISO 8601 datetimes with an offset
### What happened?

The OKF write layer stamps `generated.at` and `verified[].at` as bare dates. Both texts of OKF v0.2 define those keys as "an ISO 8601 datetime", and since the spec's in-place amendment of 2026-08-21 every timestamp-valued key is "an ISO 8601 datetime with an explicit UTC offset, for example `2026-06-30T14:00:00Z`".

### Observed

`[verified: direct calls on main at 2fcfa994]`

```
apply_okf_write_stamp("---\ntype: X\n---\n# T\n", actor="markdown-vault-mcp/4.2", today=date(2026, 9, 23))
  -> generated: {at: '2026-09-23', by: markdown-vault-mcp/4.2}
append_okf_verification("---\ntype: X\n---\n# T\n", subject="peter", today=date(2026, 9, 23))
  -> verified: [{at: '2026-09-23', by: human:peter}]
```

Both take `today: date` and write `today.isoformat()`. The `okf_verify` tool and the write enricher pass `date.today()`.

The spec's July text (§5.2) already said "`generated.at`: An ISO 8601 datetime"; the August amendment (knowledge-catalog PR #323) adds the offset requirement and says of consumers: "Consumers are expected to be strict rather than lenient about a value that arrives without an offset" — the reference agent ignores such a value and the Dataplex demo's storage rejects a bare date at write ("`Text '2026-12-31' could not be parsed at index 10`"). A bundle this server has stamped would fail that push.

Sources and quotes are on the reference page `docs/design/reference/okf-v0.2.md` (§ "Trust", § "Where this project departs").

### Expected

Stamps are ISO 8601 datetimes with an explicit offset, so a bundle written by this server is accepted by a strict v0.2 consumer.

### Reproduction

Any content write on a vault with `MARKDOWN_VAULT_MCP_OKF_WRITE=true`; read the note's `generated.at`.

### Suspected cause

`[verified: read the source]` `apply_okf_write_stamp` and `append_okf_verification` in `okf.py` take a `date` and call `isoformat()` on it; nothing in the write path holds a time or an offset.

### Open questions / scope

- Which offset to write (server-local, or UTC `Z`) is not decided by the spec, which requires only that one be present.
- `[unverified]`: whether python-frontmatter's dumper round-trips a `datetime` string as written, or re-serialises it as `2026-06-30 14:00:00+00:00` the way PR #323 found PyYAML does — the stamp may need to be written as a quoted string.
- The read side of the same amendment (`stale_after` as an instant) is a separate observation, filed alongside.

---

pvliesdonk:
## Closes / Refs

Closes #1372. Refs #1373 (PR #1375, the read side — same `okf.py`, different functions; whichever merges second rebases its release-notes hunk), #1357, the reference page `docs/design/reference/okf-v0.2.md`.

## What & why

Both texts of OKF v0.2 define `generated.at` and `verified[].at` as "an ISO 8601 datetime"; the 2026-08-21 amendment adds "with an explicit UTC offset" and shows that a strict consumer rejects a bare date at write. The server stamped `today.isoformat()` — a bare date — in both places.

**Decision (on the issue):** UTC, `Z`, seconds precision, written as a *string* — the spec's own example form, `2026-09-07T05:32:29Z`. A string because a `datetime` value is re-dumped by the YAML emitter as `2026-06-30 14:00:00+00:00`, the corruption the amendment PR names. One helper, `okf_timestamp`, spells it for both writers; it refuses a naive value, since `astimezone` would silently read one as host-local time and invent an offset the caller never stated. `apply_okf_write_stamp` / `append_okf_verification` take `now` (aware); the enricher and `okf_verify` pass `datetime.now(UTC)`. The log's `## YYYY-MM-DD` headings stay the server's local day (§9, unchanged by the amendment) — recorded in `okf.md` § 6 as two rules meaning what they each say.

Accepted consequence, in the Upgrading notes: `generated.at` moves on every content write, which is what "the content's last meaningful change" means; the old once-per-actor-day stability was a side effect of the date form, never a recorded decision (the review read the feature commit and PR #984 to check).

## Conformance (executed by the review)

- `+02:00` input → `2026-09-07T05:32:29Z`; microseconds truncated; UTC unchanged; naive → `ValueError`.
- Raw text after stamping: `at: '2026-09-07T05:32:29Z'` (quoted); `frontmatter.loads` → `str`; re-`dumps` → identical. Same for `append_okf_verification`.
- Matches the accepted-row shape of the amendment's storage table; nothing we emit resembles its two rejected rows.

- Read `generated.at` anywhere (nothing does), so no `INDEX_SEMANTICS_VERSION` bump and no reader change.
- Rewrite stamps already on disk; a bare-date stamp stays until the note is next written.
- Change the log heading to UTC.

## Self-review 23d97ede..36b30bd2

Charters: rules (incl. the import-surface guard — `okf` submodule names are not root-exported, no snapshot change), diff, comments, history, conformance (executed). Coverage: full; one read-only agent, the diff charter walked by me as well.

- `okf_timestamp` — important/verified — accepted a naive `datetime` and read it as host-local time (would stamp `03:32:29Z` for a naive `05:32:29` on this host). Resolved in `36b30bd2`: refused with `ValueError`, pinned.
- `docs/guides/okf.md` example frontmatter — minor — still showed `at: 2026-08-01` beside prose saying `at` is an instant; the only such example in `docs/`/`examples/` (swept). Resolved.
- Log heading (local day) vs stamp (UTC) can name different days around local midnight — minor — deliberate, now written down in `okf.md` § 6. Resolved.
- Test name still said "same day" for what is now a same-instant guard. Renamed.
- History: `git log -S"today.isoformat()"` → the feature commit and this one; no idempotency or anti-churn rationale was ever recorded, so nothing deliberate is weakened.

## Verification

- TDD: the pure-transform tests failed on the signature, then on the values; the naive-guard test failed before the guard.
- `uv run pytest`: 4529 passed, 3 skipped (round 1: OKF files 204 passed); ruff, mypy (232 files), pre-commit `--all-files` (Vale), structural gate 100%, `mkdocs build --strict` 0, `check_references.py` 0, client-surface budget test passes with the `<UTC instant>` docstring edit.

### DIFF
--- a/docs/design/okf.md
+++ b/docs/design/okf.md
@@ -363,7 +363,11 @@ and none of it is implied by vault declaration.
   OKF consumers should calibrate to that.
 - **Convention maintenance:** on successful writes, append a `log.md` entry
   (newest-first, `## YYYY-MM-DD` section, `**Update**:`-style bullet) and
-  refresh the affected folder's `index.md` listing. Both are guaranteed
+  refresh the affected folder's `index.md` listing. The section heading is
+  the server's *local* calendar day, as §9 has it and as a reader of the
+  log expects, while the same write's `generated.at` is a UTC instant
+  (#1372); around local midnight the two can name different days, which is
+  the two rules stating what they each mean, not a defect. Both are guaranteed
   versions of what the advisory layer asks the agent to do. Generated
   `index.md` content derives from the same data as `get_toc`.
   **Status:** shipped in phase 5b (see §9). Maintenance runs only for
--- a/docs/design/okf.md
+++ b/docs/design/okf.md
@@ -315,7 +315,9 @@ Everything here changes bytes or write outcomes; all of it is operator-gated
 and none of it is implied by vault declaration.
 
 - **Provenance stamping:** writes through `write`/`edit` set/update
-  `generated: {by, at}`. Actor string: authenticated identity when available
+  `generated: {by, at}`, `at` the UTC instant of the write in the spec's
+  example form (`2026-06-30T14:00:00Z`, a string; #1372). Actor string:
+  authenticated identity when available
   (`human:<subject>` via the existing access-token dependency), else
   `markdown-vault-mcp/<version>` as a tool actor. Existing `generated`
   values are overwritten (it describes the current bytes); `sources` are
@@ -326,7 +328,7 @@ and none of it is implied by vault declaration.
   body also invalidate (the spec ties verification to the concept, not the
   body alone); rename does not. This is the highest-value enforcement:
   it is exactly the invariant an advisory-only setup eventually misses.
-- **`okf_verify` tool:** appends `{by: human:<subject>, at}` to `verified`,
+- **`okf_verify` tool:** appends `{by: human:<subject>, at}` (`at` a UTC instant, #1372) to `verified`,
   promoting the note's tier. `destructiveHint=False`, `idempotentHint=False`.
 
   The authenticated subject is *whose token* is in play, not evidence that a
--- a/docs/design/reference/log.md
+++ b/docs/design/reference/log.md
@@ -2,6 +2,7 @@
 
 ## 2026-09-07
 
+- `okf-v0.2.md`: the write-stamp departure removed for #1372 — `generated.at` and `verified[].at` are now UTC instants in the spec's example form, written as strings; the `generated`/`verified` claims gain their pins and the open offset question closes.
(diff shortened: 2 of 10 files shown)


pvliesdonk:
@claude review

PR #1379, opened 2026-09-07
fix(scanner): read a markdown destination in every CommonMark spelling

### LINKED ISSUE #1353: [Bug]: an inline link destination is taken as raw text up to the first ), so CommonMark's other destination spellings resolve to nothing
### What happened?

`_RE_INLINE_LINK` (`scanner.py`) captures an inline link's destination as `[^)\n]+` — everything up to the first `)`. CommonMark defines several destination spellings that this reads wrongly, so a link written in any of them is stored as a target that cannot exist and reported broken.

### Observed

`[verified: direct calls to extract_links on main at 0a473a59]`, source note `sub/src.md`:

| written | stored `raw_target` | stored `target_path` |
|---|---|---|
| `[x](<my note.md>)` | `<my note.md>` | `sub/<my note.md>` |
| `[x](a\(b\).md)` | `a\(b\` | `sub/a\(b\` |
| `[x](a(b).md)` | `a(b` | `sub/a(b` |
| `[x](note.md "title")` | `note.md "title"` | `sub/note.md "title"` |
| `[x](a&amp;b.md)` | `a&amp;b.md` | `sub/a&amp;b.md` |

The pointy-bracket form is CommonMark's spelling for a destination containing spaces, which makes it the natural way to link a note titled with spaces without percent-encoding. Reference-style definitions already strip a trailing title (`_collect_reference_definitions`), so the same title on a reference link resolves and on an inline link does not.

### Expected

A destination written in any CommonMark spelling names the same file the literal spelling does.

### Reproduction

Notes `sub/my note.md` and `sub/src.md`, the latter containing `[x](<my note.md>)`; `get_outlinks("sub/src.md")` reports `target_path="sub/<my note.md>"`, `exists=false`.

Observed on `main` at `0a473a59`.

### Suspected cause

`[verified: read the source]` The destination group is a single negated class; nothing in `_extract_inline_links` interprets angle brackets, backslash escapes, balanced parentheses, a trailing title, or entity references. I have not verified whether the same holds for the destination of a reference definition beyond the title stripping it already does.

### Open questions / scope

- Percent-encoded destinations are a sixth spelling of the same class and are tracked separately in #1332; this issue is the other five.
- `[unverified]`: `compute_new_raw_target` / `apply_link_replacement` (the rename path) may share this shape, since they search the file for the stored `raw_target`.

Surfaced by the research step for #1332.

---

pvliesdonk:
## Closes / Refs

Closes #1353. Refs #1332 (percent-decoding, the sixth spelling), #1333 (attachment classification now reaches these spellings), #1334 (no line ending inside the parentheses, kept), #1105 (rename fidelity), #1365 (this bumps `INDEX_SEMANTICS_VERSION` 7 → 8 under the rule as written).

## What & why

An inline link's destination was everything up to the first `)`, so the five spellings CommonMark §6.3 defines beyond the literal one stored targets no file can have: `[x](<my note.md>)` → `sub/<my note.md>`, `[x](note.md "title")` → `note.md "title"`, `[x](a(b).md)` → `a(b`, and `\(` or `&amp;` stayed as typed. The pointy form is how CommonMark spells a destination with a space, so a note titled with spaces could not be linked without percent-encoding.

**Decision (on the issue, with one correction posted from the review):** parse by §6.3's grammar with the reference page's rows as the test set.

- `_parse_destination` reads what follows `](`: optional spaces/tabs; the `<…>` form (no line ending, no unescaped `<`/`>`; spaces and `)` allowed) or the plain form (parentheses escaped or balanced, **three levels**, the depth the spec's examples reach — a bounded-depth regex whose alternatives are mutually exclusive per character, so a pathological line is one linear pass: 4000 unclosed openers on one line 0.40 s on `main`, 0.012 s here); then an optional title in any of the three quotings, escapes honoured, separated by spaces/tabs only (a NBSP is a destination character, Ex. 507); then `)`. The scan resumes after the whole link, so `[b](y)` inside a title is not a second link.
- One decoder for extraction and the rename path, `decode_markdown_destination` in `utils/links.py`: brackets off; backslash escapes and character references in **one pass** (escape leftmost, so `\&#46;` is a literal `&`); references only as §2.5 has them — a valid HTML5 name with its `;` (all 2125 checked) or a numeric reference, with U+0000, surrogates and > U+10FFFF → U+FFFD; then `decode_link_target`'s percent layer with its refusals; then the fragment split at the first `#` that is neither escaped nor the `#` of a well-formed `&#…;` (`split_markdown_fragment`, shared with the rename path).
- The anchor test is made **on the spelling as written**: `#h` and `<#h>` are anchors; `%23x.md`, `\#x.md`, `&#35;x.md` are files named `#…`. (Round 1 caught that testing the decoded form dropped `[a](%23x.md)` — a regression against `main`.)
- Reference definitions get the same escape-aware title stripping and the same decoder.
- `raw_target` stays as written, title excluded (`<my note.md>`, `a\(b\).md`); `target_path` is the decoded form. Attachments in any spelling are excluded (#1333).

**The forced consequence: the rename path.** A `<…>` link that now resolves is a backlink a rename rewrites; `compute_new_raw_target` compared the destination with its brackets on and would have written `[x](new.md#frag>)`. It now takes the brackets off for the comparison, decodes escapes and entities as well as percent-escapes, splits the fragment with the same rule, and puts the brackets back: pointy stays pointy, plain stays plain. No escaping is introduced where the author used none — with one bounded exception the review found: a new name containing `#` (or `<`/`>` inside the pointy form) would come back *re-pointed* (read as a fragment or anchor, the backlink silently gone) rather than *broken*, so exactly those characters are escaped on rewrite (`\#y.md`, `<new\>x.md>`).

**Departures, pinned and documented** (`design.md` "Link Extraction", the reference page's departure entries): three-level parenthesis cap; spaces and tabs accepted in a plain destination (only spaces/tabs trimmed at its edges); no line ending inside the parentheses (#1334); link text still not bracket-balanced; wikilinks untouched.

## Conformance (executed by the review against spec.json, 0.31.2)

All 51 examples in "Links", "Backslash escapes" and "Entity and numeric character references" containing `](`: **27 agree** with the spec's `href`, **20 out of scope** (every one on the departure list, or pre-existing inline-precedence rules outside §6.3: raw HTML, autolinks, images, empty destinations), **1 disagreed** (Ex. 507, NBSP — fixed). Entity name regex matches all 2125 semicolon names; every ASCII punctuation escape (Ex. 12) matched.

- Balance brackets in link text (`[a [b] c](x)`), parse raw HTML or autolink precedence, accept an empty destination, or accept a destination wrapped onto its own line: all recorded departures, unchanged.
- Escape a new name with a space or unbalanced paren on rename (comes back broken, visibly — the pre-existing class).

## Self-review, two rounds, 9c543b33..513b9a99

Charters: rules, diff, comments, history, conformance (executed — the table above, the rename round-trip on seven shapes, 20 KB pathological inputs). Coverage: full; two read-only agents in round 1 (conformance; rules/comments/history/rename) and one in round 2, the diff charter walked by me as well.

**Round 1** (on `d5e7e593`):
- blocker/verified — the anchor test ran on the decoded destination, dropping every file named with a leading `#` (`%23x.md`, `\#x.md`, `&#35;x.md`); `%23` is a regression against `main`. Resolved: `is_anchor_destination` judges the written spelling; pinned ×7.
- blocker/verified — `&#xD800;` produced a lone surrogate in `target_path`, which UTF-8 refuses: one such note would abort the whole index build (`UnicodeEncodeError` is not caught in `_build_from_notes_inner`). Resolved: surrogates → U+FFFD per §2.5; pinned.
- important/verified — `_RE_TRAILING_TITLE` used `\s`, so a NBSP separated a "title" (Ex. 507). Resolved: `[ \t]`; pinned.
- important/verified — the rename path split the fragment on a plain `#` while extraction did not (`x&#46;md` misjudged as relative). Resolved: `split_markdown_fragment` shared; pinned.
- low ×3 — a backslash escaped a line ending inside `<…>`; DEL not a control character; `finditer` resumed inside a parsed title and extracted a phantom link. All resolved and pinned. Doc nits (`decode_percent` name, the departure entry's header) resolved. The issue's "any depth" wording corrected on the issue.

**Round 2** (on `34535c51`): all seven fixes re-executed and confirmed; the scan loop shown to be strictly increasing and to miss no nested link. Lows, all resolved in `513b9a99`: escapes and references decoded in two passes let `\&#46;` re-read as a reference (now one pass); Unicode-wide `.strip()` trimmed a NBSP the grammar keeps (now `strip(" \t")` at all three sites); a rename producing a `#` or `>` in a name came back re-pointed (now escaped, pinned, documented); the anchor rule and the §2.5 invalid-code-point rule were stated nowhere in the docs (now in `design.md` and the reference page, pinned).

## Verification

- TDD: the new file failed at import, then per feature; each round's tests failed before their fix.
- `uv run pytest`: 4604 passed, 3 skipped; `utils/links.py` 100%, scanner's new code covered (remaining misses pre-existing).
- Timing, the #1334 inputs: 0.295 s / 4.28 s / 0.128 s (unchanged shape).
- ruff, mypy (233 files), pre-commit `--all-files` (Vale), structural gate 100%, `mkdocs build --strict` 0, `check_references.py` 0.


pvliesdonk:
@claude review

PR #1406, opened 2026-09-08
fix(okf): the maintained log.md stacks a frontmatter block per write

### LINKED ISSUE #1391: [Bug]: OKF_WRITE log.md append stacks one more frontmatter block per write when required_frontmatter is set
### What happened?

With `OKF_WRITE` enabled on a vault whose `required_frontmatter` includes `title`, every MCP `write` / `edit` / `append` into a folder prepends one more `---\ntitle: Log\n---` block to that folder's `log.md`. The blocks accumulate one per write; the bullets themselves are appended correctly.

### Observed

A live vault folder's `log.md` after 18 writes: 18 identical frontmatter blocks stacked at the top, then the `# Log` heading and 18 bullets (abridged):

```markdown
---
title: Log
---

---
title: Log
---

(... 16 more identical blocks ...)

# Log

## 2026-09-07

- **Update**: wrote `3-Resources/Research/_conventions.md`
- **Update**: wrote `3-Resources/Research/sbom-accuracy-completeness-and-binding.md`
(...)
- **Update**: edited `3-Resources/Research/selective-disclosure-and-zero-knowledge-over-sboms.md`
```

`read("…/log.md")` still reports `frontmatter.title == "Log"` because the parser takes the first block as metadata and treats every later block as body, so the file stays indexed and `okf_validate` does not flag it.

### Expected

One frontmatter block, regardless of how many writes the folder has received.

### Reproduction

Real `Vault`, no git, current `main` (`390d4e8b`):

```python
root = tmp / "vault"; (root / "guides").mkdir(parents=True)
(root / "index.md").write_text('---\nokf_version: "0.2"\ntitle: Root\n---\n# Root\n')
col = Vault(source_dir=root, read_only=False, okf_mode="on", okf_write=True,
            required_frontmatter=["title"])
col.index.build_index()
for n in "abc":
    col.writer.write(f"guides/{n}.md", f"---\ntitle: {n}\n---\n# {n}\n")
    wait_for_writer_drain(col)
print((root / "guides/log.md").read_text().count("title: Log"))   # 3
```

Output: three blocks, three bullets. Not reproducible on a vault with no `required_frontmatter`, where `log.md` is written without frontmatter.

### Suspected cause

`[verified: traced in code and confirmed by the reproduction above]` `ConventionMaintainer._append_log` reads the current `log.md` through `DocumentManager.read()` and passes `existing.content` to `append_okf_log_entry` as if it were the body alone; its own docstring says "`read()` hands back the body alone". `NoteContent.content` is documented in `types.py` as "Raw markdown body including frontmatter", and that is what the read returns. The maintainer then writes the result back with `frontmatter=` set, so the serializer emits a new block above text that still contains the previous one.

`[verified: git tag --contains 65083123]` The read-modify-write with frontmatter carry-over arrived in #1196 (the fix for #1174) and is in v4.1.0 and v4.2.0-rc.0.

`[verified: read tests/test_okf_convention.py]` The unit-test double's `read()` returns a body-only `content`, and the real-vault two-write test asserts that the title and both bullets are present, not that the block appears once, so the suite is green with the defect.

### Open questions / scope

- Deployed vaults already carry stacked blocks; the file above is one. Whether a fix should also normalise those on the next append, or leave repair to the operator, is a decision for the fix.
- `[unverified]`: `generate_index` writes a freshly built body rather than the read-back text, so it does not appear to share this path; not confirmed by a test.

---

pvliesdonk:
## Closes / Refs

Closes #1391. Refs #1403 (the repair of files already stacked, deliberately not here).

Re-implementation from scratch after the earlier attempt was pulled and the issue reopened. Same destination, different route: this one fixes the read/write contract mismatch and stops.

## What & why

`DocumentManager.read` returns the whole file in `NoteContent.content`, block included (`managers/document.py:386` builds it from `_read_text_utf8`), while `write` puts a `frontmatter=` mapping *above* the body it is handed (`managers/document.py:663`). The OKF convention maintainer's `log.md` append fed the read text straight back, so on any vault whose log carries frontmatter — a `required_frontmatter` seeding `title`, or a hand-seeded block — every enforced write serialised one more identical block above the last. One live folder had accumulated eighteen.

`scanner.strip_frontmatter_block` takes the opening block off before the append; the frontmatter is still carried across as the mapping, which is what #1174 needs. The helper delegates detection to `python-frontmatter`, so it agrees with `parse_note` about where the block ends — same handlers, same offset-zero anchoring, an unterminated block read as no frontmatter at all — and returns the remainder **unnormalised**, unlike `frontmatter.parse`, which strips the body at both ends and would silently de-indent a leading code block. The separator newlines come off with the block because `frontmatter.dumps` re-inserts them; keeping them added a blank line per rewrite.

Only the block that *opens* the text comes off. An identical block further down is body text, because telling one stacked by the defect apart from one a log quotes on purpose needs a rule about body content this layer does not have.

**The class was swept.** `_append_log` is the only read-modify-write that carries frontmatter across a rewrite. The other three candidates are correct as they stand and are untouched: `okf_migrate.generate_index` reads only `existing.frontmatter` and writes a freshly built body (`managers/okf_migrate.py:195-207` — this confirms the issue's `[unverified]` note); `okf_migrate.seed_log` returns early when the log exists (`:244`); `okf_migrate.convert_links` writes the rewritten raw text back *without* `frontmatter=` (`:128`), which is why it never stacked. `DocumentManager.edit` goes through `_finalize_content_write` and never re-serialises a block.

**The unit-test double is part of the fix.** It returned a body-only `content`, which is what let the defect stay green through #1196. It now serialises the block back in as the real read does; with the source fix reverted, both it and the new real-vault test fail.

- **Remove blocks already on disk.** A 4.1.0 or 4.2.0-rc.0 vault keeps what it has, in the body; the log stops growing. #1403.
- **Parse or inspect body text.** That is the rule #1403 needs and this fix does not invent.
- **Touch `generate_index`, `seed_log`, or `convert_links`** — verified correct above, so changing them would be churn.
- **Bump `INDEX_SEMANTICS_VERSION`.** No stored-row derivation changed: the helper is not on the indexing path, only on the write path. Per the once-per-release policy (#1365) the current value already covers this release in any case.
- **Move the release-notes watermark.** `notes-range-end: 60111e3f` stays: the squash SHA is not knowable here, and advancing it to a SHA that never lands on `main` would make the next notes refresh skip real research. The added paragraph is folded prose, not research output.

## Local review

- [x] No commit carries `!`. Nothing on the operator surface changes; the public library interface only gains `scanner.strip_frontmatter_block` (`tests/public_import_surface.txt` is unchanged — the package root is what it guards).

## Docs impact

- [ ] `README.md` — no user-facing surface changed (no env var, tool, or flag)
- [x] `docs/` site pages — `docs/releases/4.2.md`, one paragraph. The defect shipped in v4.1.0 (`git tag --contains 65083123` → `v4.1.0`), so it is a real fix against this release's range start, not intra-range churn.
- [x] `docs/design/` — `docs/design/okf.md` said `DocumentManager.read` "returns the body without frontmatter". It does not; corrected, with the split now stated as the maintainer's own work.
- [x] Inline docstrings — `strip_frontmatter_block` (new), `_append_log`, `append_okf_log_entry`'s `text:` parameter, `NoteContent.content`.

## Self-review fa94472b..52a382fb

Charters: rules, diff, comments, history, conformance (no normative content — the log's `# Log` / `## YYYY-MM-DD` shape is unchanged), past-PRs (the pulled attempt's Codex findings). Coverage: full.

Four findings, all resolved in `52a382fb`:

- `src/markdown_vault_mcp/okf.py:919` — important/verified — docstring rot the diff introduced: `append_okf_log_entry`'s `text:` parameter said "The current `log.md` **file text**", but its only production caller now hands it the body. Exactly the misreading being fixed, one call away. Fix: the parameter and return are documented as the body, pointing at the helper for callers holding raw text.
- `src/markdown_vault_mcp/types.py:266` — important/verified — `NoteContent.content` read "Raw markdown body including frontmatter", the self-contradictory phrasing the defect grew from. Fix: "the note's raw file text, frontmatter block included", with the rewrite hazard named.
- `tests/test_okf_convention.py` — important/verified — an untested behaviour change: taking the block off is unconditional, but putting one back is the policy's call, so on a vault with no `required_frontmatter` a hand-seeded *empty* `---\n---` now disappears on the next append. It carried nothing and leaving it in the body is what stacks, so this is intended cleanup — now pinned by `test_an_empty_block_is_dropped_on_an_unconfigured_vault` rather than left to be discovered. A block **with** keys is still carried across (#1174's case, still pinned).
- `tests/test_scanner.py` — minor — an identity assertion (`is text`) over-specified "returned unchanged"; equality says what is meant.

Coverage: `_okf_convention.py` 100%, `types.py` 100%, `okf.py` 98% (misses are unrelated functions), and every added `scanner.py` line covered. Structural gate 100% over 189 changed lines, 0 violations. Full suite 4614 passed. `mkdocs build --strict` and Vale clean.

The pulled attempt's two surviving Codex findings both reproduce against a naive version of this helper and are pinned here: the empty block (`test_an_empty_block_is_still_a_block`, plus the real-vault `test_an_empty_seeded_block_is_replaced_not_stacked_above`) and body normalisation (`test_the_body_is_not_normalised`). Its earlier three belonged to the healing loop this revision does not have.

---

### DIFF
--- a/docs/design/okf.md
+++ b/docs/design/okf.md
@@ -467,11 +467,23 @@ carry. Both generators and the `OKF_WRITE` maintainer write through it:
   trade one defect for another.
 
 The `OKF_WRITE` maintainer needs this for a second reason: `_append_log` is a
-read-modify-write and `DocumentManager.read` returns the body without
-frontmatter, so the log's frontmatter has to be carried across the rewrite
-explicitly. Since the maintainer runs after *every* content write, the
+read-modify-write, so the log's frontmatter has to be carried across the
+rewrite explicitly. Since the maintainer runs after *every* content write, the
 alternative is not a one-time gap but frontmatter stripped again after each
 save — including frontmatter an operator seeded by hand.
+
+The split is the maintainer's own work, and getting it backwards is #1391.
+`DocumentManager.read` returns the *whole file* in `NoteContent.content`,
+block included, while `DocumentManager.write` puts a `frontmatter=` mapping
+above the body it is handed. Feeding the read text straight back therefore
+serialises a second block above the one still in it, once per write; a live
+folder accumulated eighteen. `scanner.strip_frontmatter_block` takes the
+block off first, delegating detection to `python-frontmatter` so it agrees
+with `parse_note` about where the block ends, and returning the remainder
+unnormalised. Only the block that *opens* the text comes off: an identical
+block further down is body, because separating "stacked by the defect" from
+"quoted on purpose" needs a rule about body content that this layer does not
+have. Repairing files the defective releases already stacked is #1403.
 
 ### Export
 
--- a/docs/releases/4.2.md
+++ b/docs/releases/4.2.md
@@ -386,6 +386,22 @@ in the spec's example form, `2026-09-07T05:32:29Z`, written as a string so
 the YAML emitter cannot re-spell it; stamps already on disk are left as they
 are until the note is next written.
 
+**A maintained `log.md` stops collecting frontmatter blocks.** With
+`OKF_WRITE` on and a `required_frontmatter` configured, every write into a
+folder added one more identical `---\ntitle: Log\n---` block to the top of
+that folder's `log.md`; one live folder had accumulated eighteen
+([#1391](https://github.com/pvliesdonk/markdown-vault-mcp/issues/1391)). The
+per-write log append reads the file, adds its bullet and writes it back with
+the frontmatter carried across. But the read hands back the whole file, block
+included, and the write puts the carried block above whatever body it is
+given. The append now takes the opening block off the text first, so the file
+keeps exactly one however many writes the folder receives. Nothing reported
+the defect because the bullets were always correct and the file stayed
+indexed throughout. Blocks a 4.1.0 or 4.2.0-rc.0 vault already wrote stay
+where they are, in the body: removing them means telling a stacked block
+apart from one a log quotes on purpose, and that is
+[#1403](https://github.com/pvliesdonk/markdown-vault-mcp/issues/1403).
+
 **A link with any URI scheme is external.** The external-link filter was an
 allowlist of four prefixes, so a `file:`, `ftp:`, `obsidian:` or `zotero:`
 destination was resolved against the source note's folder and reported as a
--- a/src/markdown_vault_mcp/_okf_convention.py
+++ b/src/markdown_vault_mcp/_okf_convention.py
@@ -33,6 +33,7 @@ from markdown_vault_mcp.okf import (
     ReservedFrontmatterPolicy,
     append_okf_log_entry,
 )
+from markdown_vault_mcp.scanner import strip_frontmatter_block
 
 if TYPE_CHECKING:
     from collections.abc import Callable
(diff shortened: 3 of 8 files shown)


pvliesdonk:
@claude review

PR #1426, opened 2026-09-08
docs(design): OKF administration under more than one author

### LINKED ISSUE #1393: [Feature]: one maintainer for the OKF reserved files — split maintenance from OKF_WRITE stamps, derive log.md from git history

When a vault is synced by git and authored from more than one client, `index.md` and `log.md` can have two maintainers: this server under `OKF_WRITE`, and an external client (an Obsidian plugin, a second server, a human). Both insert at the same place, the top of the newest `log.md` section and the whole `index.md` listing, so concurrent edits conflict on rebase `[verified: real-git probe, 2026-09-07: same-day log bullets on both sides → conflict; both append to the same section → conflict; adjacent index inserts → conflict]`. Sequential edits do not conflict but the server's regeneration overwrites the other maintainer's body wholesale, so the files flip format on alternate writes, one commit each.

Nothing in the configuration or documentation states the constraint the owner set on 2026-09-07: "*only one should maintain the OKF administration*". The single flag `OKF_WRITE` bundles two things with different risk: per-note provenance stamping, which lives inside the note being written and cannot conflict on its own, and cross-file reserved-file maintenance, which is the conflict source. An operator who wants stamps while an Obsidian plugin owns the administration has no setting for that.

`log.md` has a second problem in a git-managed vault. The append model records only this server's writes, so the file is a partial history that omits every change made by another clone. The OKF field reference says of `log.md`: "In a git-hosted bundle `log.md` largely duplicates the commit log; it earns its keep when bundles travel without version-control history." The server already derives a `log.md` from git history in `okf_seed_log`.

### Desired outcome

- The documentation states that exactly one party maintains the reserved files, and what happens to hand edits when the server is that party. The owner's position: "If we choose 'server managed' then clobbering is fine and maybe even preferred."
- The operator can declare the server as maintainer of the reserved files separately from enabling provenance stamps.
- In a git-managed vault, `log.md` is derived from history (the owner: "We might be able to get that information from commit messages, with an honest question on whether (in a git managed environment) we should not use those as a source of truth anyway") rather than appended to, or is left out of the working tree and produced at export.

### Implementation ideas

Observations that bear on the design, not a prescription:

- A `.gitattributes` line `log.md merge=union` merged both probed conflict shapes cleanly and correctly, with no duplicated date heading `[verified: real-git probe]`. It depends on the other client's git honouring merge attributes; Obsidian mobile's isomorphic-git is `[unverified]` on that point.
- A history-derived `log.md` has to skip commits that touch only reserved files, or it logs itself, and a typical Obsidian auto-commit message ("vault backup: <timestamp>") carries no usable entry text; the changed-file list does.
- `OKF_WRITE` defaults to off today, so the maintainer is already opt-in; the missing piece is the split, not the gate.

Related (2026-09-07 multi-author reserved-files analysis): #1391, #1392, #1394, #1395, #1396

---

### LINKED ISSUE #1424: [Feature]: an obsidian-git reference page under docs/design/reference/

Several decisions in the OKF ownership design rest on how obsidian-git behaves — default commit subject `vault backup: {{date}}`, `syncMethod: "merge"`, no rebase on mobile, isomorphic-git honouring only a `mergeDriver` callback and not `.gitattributes` — and those facts are cited inline in the design with a fetch date, not recorded as a reference page `[verified: fetched from the plugin's src/constants.ts and README and the isomorphic-git merge docs on 2026-09-08]`. Per the `researching-references` skill, a change that depends on an external tool's behaviour is closed with a reference entry under `docs/design/reference/`, dated and sourced, with a `stale_after`.

### Desired outcome

A reference page for obsidian-git (desktop and mobile) covering commit subject defaults, sync method, merge behaviour and attribute handling, in the OKF v0.2 bundle shape the other reference pages use, and the design's inline citations replaced by links to it.

### Implementation ideas

- `[unverified]` whether obsidian-git on desktop, which shells out to the system git, honours `.gitattributes` `merge=union`; the design records this as untested.

Design: `docs/design/okf-ownership.md` (PR to follow on branch `docs/okf-ownership-analysis`).

---

pvliesdonk:
## Closes / Refs

Closes #1424
Refs #1425 (epic), #1393, #1394, #1405, #1411

## What & why

Adds `docs/design/okf-ownership.md`: the design, approved by the owner on 2026-09-08, for how the server administers OKF frontmatter and the reserved `index.md` / `log.md` when a vault is also edited elsewhere and synced by git. One `OKF_WRITE` ladder (`off|stamp|maintain|own`), read-side derivation of void `verified` entries, one ingest event, a curated `log.md` (one keyed entry per note per day from declared intent, else a diff summary, else the file list), `okf_intent` with an in-band nag, protected reserved paths, reconciliation at `own`, and a path-by-path table of safe defaults. Written for future reference; the epic and its fifteen children carry the implementation.

An independent review (read-only agent, 2026-09-08) raised four blocking and four advisory findings, addressed in `ef495d67`. The Codex (7) and Claude (4) reviews on this PR are addressed in `8023d484`; the response comment lists each. Also touched: one pointer paragraph each in `docs/design/okf.md` and `docs/design/design.md` so the authoritative chain reaches this design, and the reference bundle's `index.md` / `log.md` for the new page.

- No code, config, or user-facing docs change: #1412–#1424.
- Does not rewrite `docs/design/okf.md` §6 or `docs/guides/okf.md` (only a pointer is added to `okf.md`); both still describe the append model and the stale "up to three commits": #1423, and the guide follows the implementation.
- The obsidian-git reference page is included (`docs/design/reference/obsidian-git.md`, closes #1424) after Codex asked for it in this change rather than deferred.
- Does not close #1393; proposed as superseded by #1412 and #1416 in the epic, the owner's call.

## Local review

- [x] Ran a local code-review pass on the cumulative diff before `gh pr create` (an independent design review; the diff is one internal design document, excluded from Vale and the published site).
- [x] No commit carries `!`.

## Docs impact

- [ ] `README.md`
- [ ] `docs/` site pages
- [x] `docs/design/`
- [ ] Inline docstrings

---


pvliesdonk:
@claude review the design 
@codex review the design

pvliesdonk:
All eleven findings addressed in `8023d484`. Per finding:

**Codex**

- **Trailer vs committer (§5.1, P1).** Correct: "committer *or* trailer" left the renamed-committer case open. The trailer is now the discriminator for every new commit; the committer identity classifies only commits that predate the trailer. #1415 and #1413 updated to match.
- **`human:<author>` from a git author (§5.1, P1).** Correct, and it also broke situation C. Reconciliation no longer derives an actor from git at all: a stamp the external writer changed is left alone; a stamp it left untouched is removed together with `verified`. §3 table, §4.1, §5.1–5.3, §10 item 7 and the defaults table rewritten; #1420 updated.
- **Local unkeyed log entries (§6.2, P1).** Correct for a served clone edited directly. The take-upstream policy is now gated on a decidable check: every line added to `log.md` between the merge base and the local head is keyed; otherwise today's sibling stays. No list merger in either branch. #1416 updated.
- **Automatic diffs need their own opt-in (§10.1, P1).** Agreed; summarised log entries get a dedicated setting, off by default, with a configured backend being consent only to the explicit tool. §6.2, §10 items 4 and 5, the defaults table; #1418 updated.
- **Conventions protection vs posture (§10, P2).** Fixed: reserved files follow the ladder, the conventions file's protection is #1411's decision and posture-independent; §7, §10 item 6 and the table now agree.
- **Link from the authoritative chain (P1).** Added a pointer paragraph in `docs/design/okf.md` (header, naming §6 as superseded once the children land) and in `docs/design/design.md` (OKF Read Semantics).
- **External premises need a reference page (P1).** Written in this change: `docs/design/reference/obsidian-git.md` (README, `constants.ts`, both git managers, isomorphic-git `merge`/`pull`, Obsidian properties and templates pages; 22 sourced claims, 4 left `[unverified]` and named), wired into the bundle's `index.md` and `log.md`, `check_references.py` clean. §11 now points at it. Closes #1424.

**Claude**

- **Instructions budget.** Correct: it is the 1,536-unit `instructions` budget, and the maximal surface measures 1,507 today, so 29 units of headroom. §4.2 and §7 corrected; the 23,500 figure is kept only where it applies (tool descriptions).
- **`matches_okf_filters` does not handle `type`.** Correct; `type` is deliberately not an OKF filter dimension and lives in the `document_tags` path. §8 and #1421 retargeted.
- **Quotes not in the cited source.** Correct: "largely duplicates the commit log ... earns its keep" is the plugin skill's digest, and "derived, never stored" was a paraphrase. §2 now quotes `okf/SPEC.md` §3, §5.1, §5.3 and §9 verbatim and labels the digest reading as an interpretation.
- **Fourteen vs fifteen; #1405 folded in.** PR body says fifteen; §10.2 now says "items 1 to 8 ... #1405 is related, not a child", matching the epic.

@claude review
@codex review

pvliesdonk:
@claude review the design 
@codex review the design

pvliesdonk:
Round two addressed in `17087eb4`. All six Codex findings and Claude's held; Claude's is Codex's first.

- **Comparison base still by committer (§5.1; Codex P1, Claude).** The one remaining instance. "Last external content change" is now the last commit *without the trailer* (committer only for pre-trailer history), matching §10 item 1.
- **Pure renames (§5.1, P1).** Correct, and `test_rename_preserves_verified` pins the contract. The base now also requires the diff to change the note's blob, read through `--name-status` with rename detection as the history walks already do (#1297); a rename with no blob change is not a content change. #1413 and #1420 updated.
- **Reconcile `generated` and `verified` independently (§5.1, P1).** Correct; the all-or-nothing rule contradicted the §3 table. Two rules now, one per family: each is removed when the external content change left it untouched, and a family the writer maintained is kept. #1420 updated.
- **Mechanical fallback in the knowledge log (§6.2, P2).** Agreed: a file list labelled `**Update**` was §6.1's rejected model by another route. The fallback is now a *placeholder* under a distinct word (`**Change**: undescribed change to ...`), at most one per (note, day), asserting only that bytes changed, pending until intent or summary replaces it; a "no line rather than a placeholder" setting is named as worth offering. #1416 updated.
- **Pending reconstruction omits foreign commits (§6.2, P2).** Correct. The pending set is now derived from every content commit in the window, server and foreign alike, classified by the trailer, so a foreign commit ingested before a restart is still pending after it. #1417 updated.
- **Placeholder indistinguishable after restart (§6.2, P2).** Correct. The entry's source is part of its key (the bold word tells curated from placeholder), so the nag can list placeholders from the file alone. #1416 updated.

@claude review
@codex review

pvliesdonk:
Round three addressed in `82931bcb`. All four held, and two of them changed the design rather than a sentence, so the reasoning is written out here and in §5.1 / §6.2.

- **Instances at different postures (§5.1, P1).** The trailer encoded *who* committed when every reader asks *what was already done*. It now records the effective OKF level, `Vault-OKF: none|stamp|maintain|own`, and each consumer skips a commit only if that level did its work: reconciliation and the comparison base skip `stamp` and above, regeneration skips `maintain` and above, and a server commit at `none` is external for both purposes. Legacy commits classified by committer count as `maintain`. A uniform posture becomes a recommendation rather than a precondition. #1415, #1413, #1420 updated.
- **Negative dispositions (§6.2, P1).** Correct: `kind: none` and a summarizer "no" left no trace and a restart would re-derive them as pending and eventually write the placeholder the human declined. They are now persisted where the keys are, as a keyed HTML comment under the date heading in `log.md`: invisible to a reader, present to the derivation, in every clone. The state directory would be per-clone and git notes add a ref obsidian-git never pushes. #1417 updated.
- **Lower boundary (§6.2, P1).** Correct, and it revisits the watermark I rejected earlier for the wrong reason. The first draft of this fix put an `accounted-through` marker line in the root `log.md`; I withdrew it before pushing because two instances would each advance it to a different sha and conflict on that line on every concurrent pass. The boundary is instead read from the graph: every accounting pass is a commit carrying `Vault-Operation: okf_accounting`, the newest such commit reachable from `HEAD` is the boundary, and the candidates are the commits above it. Git's range semantics keep the scan bounded and still catch a branch merged late, which a date watermark misses. Activation is the first pass's commit, empty if nothing needs writing; history before it is `okf_seed_log`'s job. #1417, #1416 updated.
- **Non-git independence (§5.3, P1).** Correct and inconsistent with the git rule from round two. The index holds the note's previously indexed frontmatter, so whether the change touched `generated` and, separately, `verified` is a comparison of the prior row with the new bytes; both families now reconcile independently without git too. #1420 updated.

Also: a keyed negative comment counts as keyed in the `log.md` conflict check, so a clone holding only server-written lines still takes the lossless branch.

@claude review
@codex review

pvliesdonk:
Round four addressed in `f8295ff0`. Both held.

- **Placeholders across the boundary (§6.2, P1).** Correct: the placeholder's own accounting commit becomes the boundary, `A..A` is empty, and the range alone can no longer see the pair. The pending set is now the union of two sources: the range above the boundary for unaccounted commits, and the placeholder lines read from the `log.md` files themselves, within a nag window (candidate default seven days, so Monday's foreign change pulled on Tuesday is still nagged, and a vault whose agents never declare does not grow an endless list). Older placeholders stay as the honest record and stop being nagged; that cut-off is written down as a decision. The set is computed at accounting time and on intent, held in memory, re-derived on start, so decorating a result is a lookup. #1417 updated.
- **Added-lines-only classifier (§6.2, P1).** Correct on both counts: `append_okf_log_entry` adds unkeyed `# Log` and `## date` lines, so a server-only conflict would have taken the sibling; and a hand deletion adds nothing and would have passed vacuously, after which take-upstream throws the deletion away. The check now classifies the whole local patch since the merge base, every line, by shape: added keyed bullets and comments, added date headings, the title and seeded frontmatter of a new file, blank-line additions and removals (the helper trims trailing blanks), and a removed keyed line re-added under the same key. Anything else takes the sibling. #1416 updated.

@claude review
@codex review

pvliesdonk:
Round five addressed in `8367eda3`, and from this push on the branch gets a local `code-review` pass before every push; its report is below the replies.

**Round five (Codex).** Four of the five were one defect: §10 restated rules from §5.1 and §6.2 and drifted from them. Rather than patch the four sentences, §10 now states decisions and pointers only and says why. The fifth was real on its own: `okf_intent` could not select a (note, day) pair. It now takes targets as `path` or `path@YYYY-MM-DD`; a bare path resolves to the note's single pending day and is refused, naming the days, when more than one is pending, so today's intent never overwrites an older placeholder; the nag carries `{path, day, state}` items whose `path@day` form the tool accepts back. #1417 updated.

## Self-review cec05c84..8367eda3

Charters: rules, diff, comments, history, conformance (`check_references.py --strict` and the OKF plugin validator on `docs/design/reference`, both clean), past-PRs (the five rounds above). Coverage: full; two read-only charters ran as subagents, code claims verified against `src/` and `tests/` at HEAD (all held).

- `docs/design/okf-ownership.md` §6.2 — blocker/verified — the self-accounting rule ("a commit whose `log.md` diff adds a bullet accounts for itself") and the pending predicate ("no keyed line covers that sha") gave opposite verdicts for a human commit that adds an unkeyed bullet, so such a note would be nagged forever. Resolved in `8367eda3`: the self-accounting rule is part of the pending predicate and stated nowhere else.
- §6.2 — important/verified — the two pre-model filters produced "no entry" with no persistence, so a layout-only or frontmatter-only change stayed pending until some later pass moved the boundary. Resolved: the filters are pure functions of a commit's diff and run inside the predicate; only judgements (the model's "no", `kind: none`) are persisted. The three commit concepts (reconciliation, follow-on, accounting) are now one: an accounting pass makes one commit that carries everything it owes and is the boundary; a pass with nothing to write makes none, except activation.
- §4.2 / §7 — important/verified — the breaking-change test was applied to two of the ladder's three changes and not to the reserved-file refusal at the aliased `maintain`, which the shipped `OKF_WRITE=true` accepts today. Resolved: the ladder with the refusal is a `!` under refinement 1 and ships in 5.0 with the planned breaking changes; the aliases are an upgrade convenience, not the reason it is non-breaking. Epic #1425, #1412 and #1419 updated.
- `docs/design/okf.md` pointer — important/verified — it vouched for all of §6 while §6's "up to three commits" cost note is stale (#1423). Resolved: the pointer carves that note out.
- §10 item 4 — important/verified — "appended never overwritten" contradicted the recompute-by-key rule. Resolved: "appended and recomputed by key, never regenerated wholesale".
- Minor, resolved together: foreign entries are dated by author date (a rebase keeps it); §4.2 carried the same sentence twice with two different budgets.

@claude review
@codex review

pvliesdonk:
Closing: the ownership part of this design landed in #1433 (docs/design/okf.md §6.0); the remaining areas (change event, provenance, log curation, untyped notes) are tracked by their issues on the okf-design milestone, and this branch stays as reference material.

PR #1429, opened 2026-09-09
docs(design): simplify multi-author OKF ownership (replaces #1426)

### LINKED ISSUE #1393: [Feature]: one maintainer for the OKF reserved files — split maintenance from OKF_WRITE stamps, derive log.md from git history

### LINKED ISSUE #1424: [Feature]: an obsidian-git reference page under docs/design/reference/

pvliesdonk:
## Closes / Refs

Closes #1424
Refs #1425, #1393, #1416, #1417, #1418

Competing replacement for #1426. This branch preserves its history and reference research, then rewrites the proposal; #1426 remains open for comparison.

## What & why

The multi-author OKF design accumulated a log-accounting protocol whose progress boundary could lose pending work after a restart and whose conflict classifier could discard a human correction. This proposal separates ownership, provenance evidence, index maintenance, and explicit log curation, reducing the ownership document from 866 to 419 lines and updating the authoritative design pointers.

It retains the recorded `OKF_WRITE` ladder, curated knowledge history, explicit intent, protected reserved paths, and separate model opt-ins. It proposes removing automatic placeholders, negative-disposition comments, activation/accounting commits, patch-shape ownership inference, and the mandatory second commit. Log conflicts retain siblings. Unknown legacy or cold-build provenance is distinguished from demonstrated staleness.

The document distinguishes recorded owner choices from proposed changes and lists the epic children whose earlier issue bodies need reconciliation. Adoption of these revised product choices is the decision for this PR.

- Implement configuration, tool, ingest, or provenance changes: #1425 and its children.
- Specify exact intent key/retry semantics: #1416, #1417.
- Design background summary delivery or claim durable reminders: #1418.
- Implement generated git subjects: #1405.
- Rewrite the published guide or the implemented OKF layer's commit-count note: #1423.
- Close #1393 or rewrite the epic's existing issue bodies as though the revised proposal were already adopted: #1425.

## Local review

- [x] No commit carries `!`; this PR changes no runtime operator or library surface.

Charters: rules, diff, adjacent commitments, history, and prior PR reviews (inline comments and conversation comments). The inherited dated reference was read and structurally validated, not independently re-researched; its unverified observations remain marked. No new formal schema.

Resolved findings from #1426's final state:

- Legacy matching committer identity does not establish that OKF stamping or maintenance occurred (§5.1).
- Index maintenance no longer advances a global boundary past delayed log work; the automatic accounting protocol is removed (§6.2–6.4).
- A human can edit keyed prose in the same shape as a server recompute; conflicts retain both versions (§6.3).

Self-review range: `cec05c84..dd755403` (merge base with `origin/main` through the replacement commit).

Validation: `uv run pytest -x -q` and the full coverage run each passed with 4,628 passed / 3 skipped; total coverage 97.45%, with no executable patch lines. Ruff check/fix, format and format-check passed in order; mypy passed; all-files pre-commit passed. The structural gate reported no Python source changes. Strict reference validation, local document links, and `git diff --check` passed.

## Docs impact

- [ ] `README.md`
- [ ] `docs/` site pages
- [x] `docs/design/`
- [ ] Inline docstrings

Internal design and reference changes only; the implemented user-facing behavior is unchanged.

---


pvliesdonk:
Review of the replacement against #1426's final state (`8367eda3`), by the author of #1426.

**Verdict: this is the better base.** Five review rounds on #1426 each surfaced a consequence of the previous round's mechanism, which is the signature of a design carrying machinery it cannot keep consistent. §6.4's diagnosis ("defects in the mechanisms, not missing sentences in their summaries") is correct, and I would rather have written it. The reference bundle is byte-identical to #1426, every section reference resolves, and the pointer edits in `okf.md` and `design.md` are accurate.

**Corrections here that #1426 lacked, all real:**
- A legacy commit matching the committer identity is *unknown capability*, not `maintain`; `OKF_WRITE=false` commits carry the same identity (§5.1).
- "Edit, then re-verify" versus "re-verify, then edit" inside one pulled range cannot be told from the endpoints; the assessment must walk the range (§5.1). #1426 compared endpoints.
- A cold non-git build has no prior state and must not delete supplied claims (§5.3).
- Markdown stripping is not a semantic filter: a changed link destination leaves the visible words unchanged (§6.1).
- The suppressed-intent flag is not by itself permission to bypass protected paths; the internal path must be deliberate (§7). #1426 reused the flag.

**Three product reversals the owner is being asked to accept, stated plainly because they reverse positions from the 2026-09-08 discussion:**
1. *The log becomes selective.* No entry is written for a change nobody describes (§6.2). This drops the placeholder and, with it, the owner's earlier half-rule "if a commit has changes but does not include log.md then it must be written". A foreign Obsidian commit that no agent ever declares intent for, on a vault without the summariser opt-in, is never in `log.md`. Defensible ("a log can be selective without pretending to be exhaustive"), but it is a change of what `log.md` promises, not a simplification of how it is kept.
2. *The nag becomes optional and advisory* ("If result reminders are added ...", §6.2). Session-tracking-and-nag was the owner's own proposal and the primary channel of the approved design. Making it a bounded hint with no derivation specified is honest about what can be guaranteed, but it should be the owner's choice, not a side effect.
3. *`log.md` conflicts keep the sibling* (§6.3), reinstating the #1399 position. Consistent with dropping the classifier; it means #1395's `log.md` half stays as today.

**Two losses to restore before this replaces #1426:**
- The safe-defaults matrix (every optional path × every behaviour, with its default) was an explicit owner request on 2026-09-08 ("run through all these paths and have safe defaults"). §4.2 covers defaults in prose and §10's acceptance table covers many paths, but the path × behaviour table is gone. It is short and worth carrying over.
- The epic #1425 body and the child issues (#1412–#1422) were reconciled to #1426's final state and now contradict this proposal on the trailer semantics, the second commit, placeholders, negative comments and the boundary. §10 names the reconciliation as required but does not do it; whoever adopts this PR owns that pass, and it should happen before any child is picked up.

**One open classification.** §4.1 leaves the ladder-plus-protection breaking assessment as "must be assessed"; #1426's §4.2 and the epic body already conclude `!` under refinement 1 (the shipped `OKF_WRITE=true` accepts a reserved-file write that `maintain` refuses). Either conclusion here or a pointer to the epic's would close it.

Nothing in the document is factually wrong against the code or the references as far as I can check; the claims it makes about the repository are the ones #1426's two local charters verified. If the owner accepts the three reversals, adopt this, close #1426, restore the matrix, reconcile the issues.

---

pvliesdonk:
Closing: the ownership part of this design landed in #1433 (docs/design/okf.md §6.0); the remaining areas (change event, provenance, log curation, untyped notes) are tracked by their issues on the okf-design milestone, and this branch stays as reference material.

PR #1433, opened 2026-09-09
docs(design): OKF ownership as three independent switches, not a ladder

pvliesdonk:
## Closes / Refs

Refs #1412, #1425, #1419, #1392, #1431, #1432, #1434

## What & why

Second of the small per-surface design PRs (owner, 2026-09-09: rethink the ownership ladder, do not recycle it; a design says *what* to build, the implementation plan says how to get there). Amends the authoritative OKF design in place: `docs/design/okf.md` gains §6.0 and its §6 bullets name their owning switch, the §2 config table gains the rows it was missing, and `design.md` gets decision 25.

The design: the three writes the server can make into a shared bundle (stamp its own notes, maintain `index.md`/`log.md`, repair other people's notes) collide with different things, and an operator may want any combination. The ladder both earlier drafts carried (`off|stamp|maintain|own`) assumed each was a superset of the last and so forbade real cases, such as a git-hook-stamped vault that still wants the listing maintained. So: three independent booleans, each default off, none implying another. `OKF_WRITE` stays as it is; a proposed `OKF_MAINTAIN` owns the reserved files; a proposed `OKF_RECONCILE` repairs external notes. The switches govern what the server does on its own; explicit operations on non-reserved notes answer to read-only mode and `OKF_MODE` as today, and the one explicit operation a switch forbids is a client write to a reserved file the server maintains.

Things found not clearly written down, per the owner's rule that those become fix candidates: `OKF_VERIFY` was absent from the design's config table (fixed here); `stats` and `config://vault` do not report the write state the design said they expose (#1432); the instruction snippet tells agents to update the reserved files while the server regenerates them, and the guide reads as a survival promise (#1431); `READ_ONLY` with any OKF write setting is silent and undocumented (#1434).

- Plan the transition for deployments running `OKF_WRITE=true` today, which maintain the reserved files and would stop under `OKF_MAINTAIN`'s default; that, and whether it or the reserved-file refusal carries a `!`, is decided on the implementing PRs under the breaking-change policy: #1412, #1419, epic #1425.
- Name the settings finally or implement them: the `config-contract` skill and #1412.
- Design the ingest event, provenance evidence, log curation, reconciliation or untyped-note handling: the next per-surface PRs (#1414, #1413, #1416/#1417, #1420, #1421/#1422).
- Change `docs/guides/okf.md`, `docs/configuration.md` or `README.md`: they describe shipped behaviour and follow the implementation.

## Local review

- [x] Ran a local code-review pass on the cumulative diff before `gh pr create`, and again on the cut section before the final push.
- [x] No commit carries `!`.

## Review history

Nine bot rounds. The decision (three switches) did not change after the first; every finding was in a sentence around it that over-claimed or mis-described code, and the last rounds found that the section was carrying release classification, which is implementation planning. §6.0 was cut from 162 to 71 lines and holds only the decision, the boundary principle, the rejected alternatives with design reasons, and pointers to the issues that carry the implementation constraints.

## Docs impact

- [ ] `README.md`
- [ ] `docs/` site pages
- [x] `docs/design/`
- [ ] Inline docstrings

---


pvliesdonk:
@claude review
@codex review

pvliesdonk:
All three addressed in `9a53631d`, each thought through for what else it touches.

- **§6's heading and lead (Codex, P2).** Correct, and the heading was the real artefact: every bullet in §6 inherited "gated by `OKF_WRITE=true`" from it, so placing independent switches underneath contradicted the section's own premise. §6 is now "the write-side switches"; the lead says each behaviour names its owner, and each bullet does (stamping, invalidation and `okf_verify` on `OKF_WRITE`; convention maintenance on `OKF_MAINTAIN`; a new bullet gives `OKF_RECONCILE` an owner in this section and points at #1420). The consequence that reaches code is the tool tags: `okf-enforce` today means "off when `OKF_WRITE` is off", so §6.0 now states that tool gating follows the owning switch, one tag per switch, with a maintenance-only tool (an intent tool) tagged to `OKF_MAINTAIN` and the migration tools keeping `okf` + `write` only. Phasing gains row 5c.
- **"Never writes them" versus the migration tools (Codex, P2).** Correct, and it exposed a gap in the protection rule rather than only a wording problem: `okf_generate_index` and `okf_seed_log` are client-invoked writes to reserved paths, so #1419's refusal has to admit them explicitly. The row now says "no automatic maintenance", and a new boundary states that the two tools stay available at both values of the switch and become the sanctioned client path once the refusal exists, admitted because they are the server's own generators (the maintainer's own code), not because of the suppression flag they run under, which also covers `okf_verify` and must not become a general bypass. #1419 updated to carry that rule.
- **"Two of them" (Claude).** Correct: the ladder's four rungs are four of the eight combinations. Fixed.

@claude review
@codex review

pvliesdonk:
Addressed in `844a0e1d`. Correct: "set" is not a value, `false` is one, and a literal reading rejects a normal `OKF_MODE=off` deployment or forces tracking which variables were supplied. The rule is now about the *effective* value: rejected only when a switch is effectively true, where `OKF_MAINTAIN` unset inheriting a false `OKF_WRITE` is false and conflicts with nothing; the read-only warning uses the same distinction.

The consequence worth writing down, now in §6.0: the effective values are computed once, after inheritance and before inertness, and that single derivation feeds start-up validation, the read-only warning (#1434), the reporting keys (#1432) and the gating of tools and maintenance. Four consumers of one function cannot disagree; four re-derivations would, which is the class of defect the earlier design PRs kept producing.

@claude review
@codex review

pvliesdonk:
Addressed in `6667adae`. Both reviews found the same real contradiction, and it was mine: "one derivation feeds four consumers" while the reporting paragraph computed after inertness. There are two projections, not one, and the design now names them and assigns each consumer one.

- **Configured posture**: the three booleans after inheritance and nothing else, a pure function of configuration computed once at start-up. It feeds the `OKF_MODE=off` validation, the read-only warning, and tool registration, because a tool registered under a switch cannot disappear when a declaration is momentarily absent; activation stays a per-call check, as the maintainer does today.
- **Runnable posture**: configured ∧ bundle active ∧ vault writable, evaluated when asked because activation can flip mid-session. It gates stamping, maintenance and repair per call and is what "the instance actually does" means.
- **Reporting** follows the split the two surfaces already embody: `config://vault`, the configuration resource, reports the configured posture beside the `okf_mode`/`okf_active` it already carries; `stats.okf` reports the runnable posture beside `mode`/`declared_version`. An operator who sees a switch configured on and running off reads the reason off the same payloads. #1432 updated to that shape.

Each projection has one derivation and the second is defined in terms of the first, which is the property the previous sentence was reaching for and mis-stated.

@claude review
@codex review

pvliesdonk:
Addressed in `90074311`. Correct: the migration tools are registered under the `okf` tag, hidden only when `OKF_MODE=off`, respect read-only mode, and check no activation (`server.py` tag pass; no `state().active` in their handlers or manager). That is by design, not omission: generating the index of a vault that is not yet declared is the migration ratchet's first step, and my bullet would have broken it. The boundary now says exactly that, and #1419 carries the same wording so the refusal's admission rule is not built on the wrong gate.

One consequence written down while there, from the previous Claude review's trace: `stats.okf` is omitted entirely on an inactive bundle, as today, so the runnable posture is reported only when there is one and inactivity itself is read from `config://vault`. Without that sentence an implementer of #1432 could reasonably add an all-false `okf` section to `stats` on inactive vaults and change a shipped shape.

@claude review
@codex review

pvliesdonk:
Addressed in `19fb2d89`. Correct: §7's ratchet is audit, declare, enrich, then the mechanical transforms, so "the ratchet's first step" was the wrong reason for a right rule. The gate claim stands (no activation check; hidden only at `OKF_MODE=off`; read-only respected). The reason now given is the one §7 actually states: the tools are one-shot transforms invoked deliberately rather than ongoing enforcement, and an operator runs them before declaring a vault (the root `index.md` the declaration goes into may not exist yet) or under `OKF_MODE=on` on a bundle they cannot declare. No other sentence in §6.0 rested on the ratchet ordering.

@claude review
@codex review

pvliesdonk:
Both addressed in `9b739c48`, and they share a root I had not stated.

- **`OKF_RECONCILE` off "never touches another party's note" (Codex, P2).** Correct, and the same class as the earlier migration-tools finding: I kept writing each switch's "no" as a prohibition on the server, when the switches govern only what the server does *on its own*. Rather than patch a third cell, §6.0 now opens its boundaries with the principle once: stamping as a side effect of a write, maintenance and repair as reactions to changes are the switches' domain; an explicit, client-invoked operation, an agent editing any note, `okf_convert_links` rewriting links in notes another party authored, the two reserved-file generators, is governed by read-only mode and `OKF_MODE` only. A "no" in the table means the server initiates nothing, not that the operation is unreachable. The reconcile row now says "no automatic repair".
- **The `OKF_MODE=on` clause (Claude).** Correct: under `on`, `OkfState.active` is unconditionally true, so that case never needed the no-activation-check rule and did not illustrate it. Dropped; the pre-declaration `auto` case, where the bundle really reports inactive, is the one that motivates the rule, and the sentence now says so.

@claude review
@codex review

pvliesdonk:
Not a patch this time. The owner called the pattern: six rounds of findings, every one in a sentence I had added *around* an unchanged decision, each describing code I had not verified at that moment. So `b6ae016a` cuts §6.0 from 162 lines to 86: the decision, the one boundary principle, the compatibility argument, the rejected alternatives, and pointers. The implementation constraints moved to the issues that own them and sit next to the code they constrain: the configured/runnable projections and reporting surfaces to #1432, tool tags and validation to #1412, the read-only warning to #1434; #1419 and #1414 already carried theirs. Then one local review of the cut section, with a single charter (every sentence about current code verified against the tree; every cross-reference checked), before this push. It found two things, both addressed in `aff0ab61`:

- **Blocker, and it predates every round:** #1392's regeneration and #1419's refusal both fire at the same shipped setting and are both escapable only by `OKF_MAINTAIN=false`, and I classified them oppositely without a criterion. The criterion is now stated once: what the last stable release's documentation promised, which is how the policy is applied to any `fix:` that changes behaviour. The guide promises the server keeps the reserved files current and tells an operator who wants otherwise to turn `OKF_WRITE` off, so the regeneration restores a promise and is a fix; the switch shipping no later is the escape the fix owes because today's costs the stamps, not the ground of the classification. The guide also promises a hand write to `index.md` is left alone, so the refusal removes a promise and is the one `!`. Decision 25 says the same.
- **Important:** "an agent's edit of any note is governed by read-only mode and `OKF_MODE`" was false; ordinary `write`/`edit`/`append` carry only the `write` tag and are never gated by `OKF_MODE`. Corrected.

@claude review
@codex review

pvliesdonk:
Both addressed in `dc6044ff`, and the first changed what the section is rather than what it says.

- **"Restores a promise" versus the guide's "runs only for content writes" (Codex, P1).** Codex is right that the guide does not promise regeneration after a pull, and the owner's reading of why this argument exists at all is the deeper point: §6.0 was carrying release classification (`!`, 4.x versus 5.0, milestone sequencing), which is implementation planning, and the guide's own text is ambiguous enough ("keeps current" and "runs only for content writes" in one paragraph) that any verdict pre-decided here over-claims. The policy already places that decision on the implementing pull request against the stable release at that time. So the design now records the *facts* an implementer classifies against, quoted and attributed (three from the guide, one from the configuration reference), states that it does not decide the classification, and keeps the one thing it does decide: `OKF_MAINTAIN` is the escape both #1392 and #1419 need, and it keeps the stamping posture where today's escape does not. Decision 25, phasing row 5c, #1412, #1419 and the epic no longer assert a `!` or a release.
- **"Any note" versus the refusal (Codex, P2).** Correct; the boundary principle now exempts explicit operations on *non-reserved* notes, names the one explicit operation a switch does forbid (a client write to a reserved file while the server maintains it, #1419), and names the migration tools as admitted generators.

@claude review
@codex review


Follow-up `1afc6598`: the fourth quote carried backticks the configuration reference does not; it now reads exactly as `docs/configuration.md` has it.

pvliesdonk:
`4a2ed230`, on the owner's rule that a design says *what* to build and hedging for compatibility or intermediate stages is the implementation plan's business. Consequences: `OKF_MAINTAIN` defaults to off, not to `OKF_WRITE`'s value, which was a compatibility hedge dressed as a default and the ladder's mistake in a smaller form (now listed under Rejected with that reason); the Compatibility paragraph is gone entirely; the rejected "stamp by default" alternative carries its design reason, §2's trust model, instead of an upgrade reason. The transition for shipped `OKF_WRITE=true` deployments, and any `!`, now lives on #1412, #1419 and the epic as implementation. The PR body is rewritten to describe this final state, since it becomes the squash commit.

@claude review
@codex review

pvliesdonk:
Both addressed in `dbb01603`; both were contradictions between sentences of §6.0 that I should have caught in my own pass.

- **`okf_verify` in the exemption (P2).** Correct: it is `OKF_WRITE`'s own tool, hidden when the switch is off (table row, `server.py` tag pass), so it cannot also be an operation the switches may not forbid. The exemption now names ordinary `write`/`edit`/`append` and the `okf_*` migration tools only, and the sentence that follows lists the two explicit operations that are the switches' own: the reserved-file write refusal, and `okf_verify` existing only under `OKF_WRITE`.
- **"Any indexed change" (P2).** Correct, and it contradicted the #1414 pointer three paragraphs down. The row now triggers on any indexed change to a non-reserved note and says in place that a reserved-file change never triggers its own regeneration.

Swept the rest of the section for the same shape ("any", "alone", "only") before pushing; nothing else over-claims against another sentence.

@claude review
@codex review

pvliesdonk:
Addressed in `718b2d80`. Correct: `WRITE_PROTECT_EXISTING` raises `DocumentExistsError` on a blind write to an existing note, and `if_match` conditions a write too, so "read-only mode alone" over-claimed. The class is the point: every time the sentence enumerated what governs an ordinary write it missed one (`OKF_MODE` last round, write protection now). Its job was only to say that no OKF switch is among those policies, so it now says that: an ordinary write answers to the vault's general write policies, named as examples and delegated to `design.md`, and to no OKF switch.

@claude review
@codex review

PR #1435, opened 2026-09-09
docs(design): the vault change event

pvliesdonk:
## Closes / Refs

Refs #1414, #1425, #1392, #1394, #1413, #1416, #1420

## What & why

Third of the per-surface design PRs. Adds "Change Event" to `docs/design/design.md` beside the index lifecycle it belongs to, decision 26, and a one-line pointer from `okf.md` §6.0. It says what to build and nothing about how to get there.

The vault raises one event whenever the index has applied a change to the documents it holds, on every route: the server's own write path, an incremental reindex (pull, watcher, boot, the `reindex` tool), or a full build. It carries the paths that entered, changed or left (leaving for any reason: deletion, a newly matching exclude pattern, newly missing required frontmatter), moves as the route knows them, the origin of the change (own through the write path, observed by a scan), and whether a full build ran. Everything that keeps something derived from vault state current subscribes to it and to nothing else: the reserved-file maintainer, log curation, provenance assessment, the second-maintainer warning. It is not a git event, not the write callback, not a filesystem event; embeddings are a derived representation that converges on its own route and raise nothing.

- Specify the event's implementation: threading, delivery, locks, how the tracker grows per-path lists, the boot-sequence and warm-restart traps. Those are the owner's recorded facts on #1392/#1395 and live on #1414.
- Design the consumers: #1392, #1416, #1413/#1420, #1394.
- Touch `src/` or user-facing docs.

## Local review

- [x] No commit carries `!`.

## Self-review 5214f37d..6bd1bceb

Charter: an independent read-only agent verified every sentence about current code against the tree (own-write route, pull wiring, watcher handler, boot sequence, `ReindexResult` counts, the purge helper, the write callback's call sites, rename knowing old and new, the tracker reporting a rename as removal plus addition), the PR #1429 quotation verbatim, decision 26 and the #1414 body against the section, and every boundary word against what exists. Coverage: full.

- `design.md` Change Event — important/verified — "whatever route" and "every change … exactly one event" reached past the six routes: `build_embeddings` mutates the vector store, which `design.md` counts as part of the index, on its own route. Resolved in `6bd1bceb` by stating the scope the design is entitled to: the event is about documents entering, changing or leaving; embeddings are a derived representation and raise nothing.
- Change Event — minor/verified — `reindex(force=True)` runs the full build, not an incremental reindex. Resolved: filed under the full build.
- No what-versus-how residue found.

## Docs impact

- [ ] `README.md`
- [ ] `docs/` site pages
- [x] `docs/design/`
- [ ] Inline docstrings

---

### DIFF
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -849,6 +849,58 @@ chunks embedded by a different model are invisible to the chunk-set diff
 (the persisted provider/model fingerprint check normally catches this case
 at load time and forces the rebuild automatically).
 
+### Change Event (proposed, #1414)
+
+**Status:** design, 2026-09-09; not implemented. This says what to build;
+the constraints the first attempt (PR #1400, closed) established are on
+#1414, beside the code they constrain.
+
+The vault raises one event whenever the index has applied a change to what
+it holds, whatever route the change took: the server's own write path, an
+incremental reindex (a pull, the file watcher, the boot reindex, the
+`reindex` tool), or a full build. Everything that keeps something derived
+from vault state current subscribes to this event and to nothing else: the
+reserved-file maintainer (#1392), log curation (#1416), provenance
+assessment (#1413, #1420), the second-maintainer warning (#1394). None of
+them subscribes to git, to the filesystem, or to the tool that happened to
+write.
+
+The event says:
+
+- **Which paths entered, changed, or left the index.** "Left" covers every
+  reason a row goes: deleted on disk, newly matching an exclude pattern,
+  newly skipped for missing required frontmatter. Today's
+  `ReindexResult` reports those as counts.
+- **Moves, as the route knows them.** The own write path knows a rename as
+  a pair; a scan reports a removal and an addition.
+- **The origin of the change:** *own*, made through the server's write
+  path, or *observed*, found by a scan. A consumer that must not react to
+  the server's own work reads this. Who authored the bytes of an observed
+  change is not this event's to say; that is the provenance design's.
+- **Whether a full build ran.** A full build names no delta and means
+  everything may have changed, so consumers refresh everything. The cold
+  start, where the pull before the first build is absorbed by that build
+  and the boot reindex then reports no change, is the case this covers.
+
+Guarantees: the event is raised after the index reflects the change, so a
+consumer reads a consistent index; every change the index applies appears
+in exactly one event; a consumer's own writes go through the ordinary
+write path and raise their own *own* events, and the rule that a change to
+a reserved file never triggers its own regeneration is the consumer's
+(`okf.md` §6.0), not the event's.
+
+What it is not: not a git event, since it carries no commits and exists on
+vaults without git; not the write callback (`on_write`, the git-commit
+hook), which fires per write to commit bytes and never sees an observed
+change; not a filesystem event.
+
+Rejected: a git-commit-based event, which misses vaults without git, the
+watcher and the boot build, and whose unit is wrong anyway ("even another
+server's correctly stamped write can require a local index projection to
+be refreshed", PR #1429); one hook per route (pull, watcher, write), three
+sources that drift; `ReindexResult` as the event, which is a count summary
+returned to the caller of one route, not a delivery to subscribers.
+
 ### Embedding Convergence (#665)
 
 Since #1157 the whole embedding lifecycle — the cold build, this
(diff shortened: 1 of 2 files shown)


pvliesdonk:
@claude review
@codex review

pvliesdonk:
Addressed in `87cc499f`, with the remedy deliberately not the one proposed.

The finding is right about the sentence: "subscribes to this event and to nothing else" conflated the trigger with the inputs. Log curation learns *that* a note changed from the event; what it then writes draws on a declared intent, the note's content and git history, none of which the event carries or should. The section now says that once: the event is the trigger, not the material, and a consumer's other inputs are that consumer's design to name.

The remedy is wrong for the design, and the section now says why rather than leaving it implicit. The event has the index's granularity, not the operation's: several writes to one path before the index applies them are one change, a full build names none, and the event does not say which tool wrote, how many times, or what the operation was. The per-write `log.md` bullet in today's `okf.md` §6 is the shipped behaviour that the log design (#1416) supersedes (one curated entry per note per day, from intent or a summary, per the owner's 2026-09-08 decisions), so extending the event with non-coalesced write records, or keeping an operation-level feed beside it, would build the event around the model being retired. A consumer that needs per-operation facts is not a consumer of this event.

Also folded in Claude's two wording nits (the odd mid-sentence wrap; "own *own* events").

@claude review
@codex review

pvliesdonk:
Addressed in `94bba13c`. The race is real: the write path writes the file and marks the path dirty, and a later job reads whatever is then on disk (`managers/document.py` `_finalize_content_write`; `index_writer.py` `drain_dirty_paths` / `process_dirty_paths`), so the design's origin was defined by the route while what consumers need is defined by the bytes. Origin is now defined by the bytes the index applied: *own* when they are the bytes the server's write path wrote, *observed* otherwise. A server write overtaken on disk before indexing therefore raises one *observed* event and no *own* one, because its bytes never became index state; the exactly-once guarantee holds, since the event reports index-state transitions and there was exactly one. How the implementation tells the two apart (retaining the written hash is the obvious way) is #1414's, not the design's.

@claude review
@codex review

pvliesdonk:
Both addressed in `97f5ed2b` by removing the field they are about, on the owner's question whether the design was arranging implementation. It was: two consecutive findings on the edges of "origin" (a race, then deletion, with renames and coalescing next) are the signature of a label that is a mechanism rather than a decision. So the question became whether the event needs to say who made a change at all, and no consumer does: the reserved-file maintainer regenerates from the state it finds; provenance assessment derives authorship from git and the note under its own design (#1413, #1420); the second-maintainer warning compares what it finds with what the maintainer last wrote (#1394); log curation takes "who" from a declared intent or history (#1416). The event now says what changed and carries no origin, the guarantee sentence no longer promises an *own* label, decision 26 and #1414 match, and the race and deletion questions have no field left to be about.

@claude review
@codex review

pvliesdonk:
Addressed in `38e7dbb1`. Correct: `IndexManager.reindex` logs the excluded purge and returns `deleted=len(changes.deleted)`, so the purge never reaches `ReindexResult`, and "today's result reports those as counts" was false for that reason. The sentence described today's code, which the design does not need to do, and it is deleted rather than corrected; the precise fact, that excluded removals surface nowhere in the result and the event has to learn of them where the purge happens, is on #1414 where the implementation reads it.

This one is a self-review failure as well as an over-reach: the pre-PR charter verified "`ReindexResult` carries counts only" and not that sentence at its own precision. The charter for the next section verifies each code sentence individually or the sentence goes.

@claude review
@codex review

pvliesdonk:
Addressed in `a07b76d0`, and this time the whole section went through the discipline before the push rather than the one sentence.

- **The "left" list (Codex, P2).** Correct: the enumeration after "every reason" missed `parse_error`, `encoding_error` and `internal_error`. The list is gone; "left" is now defined by pointing at the Skip Tombstones contract, which is the contract that says when the index stops serving a note: its row is gone (deleted, or newly excluded by pattern) or has become a tombstone (any surfaced deterministic skip). A transient `OSError` leaves the row untouched and so is not a leave, verified against `managers/index.py`.
- **Same pass, my own findings against the section.** Two sentences described today's code where the decision did not turn on it, the same class as last round: the cold-start absorption mechanics behind the full-build flag, and which route knows both ends of a rename. Both cut; the flag's rationale is only that a full build names no delta, and moves are "when the route that reported the change knows both ends". "Documents" is pinned to the markdown notes the index holds, with attachments never indexed (`tracker.py` walks `**/*.md`; `fts_index.py`), because the word was carrying an unstated scope.
- **Pre-push charter, sentence by sentence.** An independent agent verified each of the ten code and contract claims at its own precision (routes, notes only, embeddings touch no note, coalescing, per-write maintenance, the tombstone contract's completeness including the OSError case, the write callback's call sites, the #1429 quotation, decision 26, #1414), every boundary word, and the absence of mechanism. Zero findings.

@claude review
@codex review

pvliesdonk:
Closed at the owner's decision. The objective facts this draft established are recorded on #1414; the branch `docs/change-event` (`a07b76d0`) stays as reference material.

PR #1451, opened 2026-09-11
fix(okf): okf_validate reports index.md frontmatter the format forbids

### LINKED ISSUE #1396: [Bug]: okf_validate does not flag extra keys on the root index.md or frontmatter on folder indexes
### What happened?

`okf_validate` reports a bundle as conformant while its root `index.md` carries frontmatter keys other than `okf_version`, and while folder `index.md` files carry frontmatter the server did not put there.

### Observed

`[verified: read src/markdown_vault_mcp/okf.py _audit_file at 390d4e8b]` For the root `index.md` the audit only records that it was seen; for any other file it flags `okf_version` as `misplaced` and, for `log.md`, checks heading shape. No check reads the root index's other keys or a folder index's frontmatter.

The OKF v0.2 field reference on `index.md`: "No frontmatter, with exactly one exception: a bundle-root `index.md` may carry `okf_version`." The skill validator errors on "a root `index.md` carrying keys other than `okf_version`" and on "frontmatter in a non-root `index.md`".

A concrete path to the state: the git conflict resolver injects `conflict_with` / `conflict_date` into a conflicting root `index.md` (separate issue), and the audit stays green afterwards.

### Expected

The audit surfaces both conditions as non-blocking format findings in the conformance category. This does not add a rejection gate or change the concept-only conformance ratio.

### Reproduction

Root `index.md` with `okf_version: "0.2"` plus any second key; run `okf_validate`; the report contains no finding for it.

### Suspected cause

`[verified: as above]` The rule is not implemented in `_audit_file`.

### Open questions / scope

Format validation applies the same role rules to authored and generated indexes. Server-seeded invalid fields remain diagnosable; operator concept-field requirements do not excuse them. #1438 covers the admission/placeholder chain and its transition.

The index_frontmatter defect remains. PR #1409 is closed as superseded; its policy-based exemption is not the desired end state. Stage 1 of #1425. The original observations above retain their recorded baseline; this update changes the target, not the reported evidence.

Keep in v4.2 as a non-mutating conformance diagnostic. Apply the same role rule to authored and server-generated index files, including invalid server-seeded fields; explain those findings rather than exempting them. Generation, seeding, search admission, configuration and stored content remain unchanged. PR #1409 is closed as superseded; a fresh fix addresses this narrowed issue. The admission/seeding redesign belongs to #1438 and #1442 after v4.2. No implementation from the closed PR is a prerequisite.

### Investigation outcome

The 11 September de-risking comment records executed parser/admission and report-compatibility probes. Land #1407 and #1408 before the complete #1396 validation; preserve generator behavior, admission, old report construction and concept-count semantics. The comment records the remaining JSON-frontmatter lenience question explicitly.

---

pvliesdonk:
## Closes / Refs

Closes #1396. Refs #1395 (the conflict resolver that injects the motivating keys), #1407 and #1408 (reader defects this diagnostic sits on top of, deliberately untouched), #1438 and #1442 (the admission/seeding redesign).

A fresh, diagnostic-only fix per the 11 September triage. Nothing from the closed #1409 is carried over: no policy-derived exemption, no change to `ReservedFrontmatterPolicy`, seeding, admission or configuration.

## What & why

OKF v0.2 §8: *"Index files contain no frontmatter, with one exception: a bundle-root `index.md` MAY carry an `okf_version` key."* §11 rule 3 makes reserved-file structure a conformance matter. The audit only noted that the root index was seen (`okf.py:_audit_file`) and never read a folder index's block, so a root index carrying `conflict_with` from a git conflict, or a folder index with a title, audited clean.

A new conformance finding, `index_frontmatter`, reports **any** block on a folder `index.md`, and on the root anything but a lone `okf_version`. The rule is one helper, `_index_frontmatter_conforms`:

- **Presence comes from the block, keys from the parse.** `scanner.strip_frontmatter_block` (the #1391 splitter, reused unchanged) decides whether a closed leading block exists. So an empty, comment-only, sequence/scalar or malformed-YAML block, all of which parse to `{}`, is still reported. It agrees with `frontmatter.parse` on leading blank lines and on an unclosed `---` / body thematic break (neither is a block).
- **Same rule for authored and generated indexes.** On a vault with `required_frontmatter`, the seeded `title` on a generated index is reported. Generation, seeding and admission are unchanged: `test_generated_index_is_audited_like_any_index` asserts the file is still indexed *and* reported.
- **The concept ratio keeps its meaning.** Reserved files still `return` before `total_notes`, so `conformant_notes / total_notes` and every other finding count are unchanged. A malformed root block shows up here only, not under `unparseable_frontmatter`.
- **Deliberate overlap:** a folder index declaring `okf_version` is both `index_frontmatter` and `misplaced_okf_version`. They describe different violations.

**JSON lenience, decided explicitly (the issue's open edge):** the rule judges keys and does not check syntax. Keys come from the same `python-frontmatter` parse every other audit rule and the declaration probe use, so a JSON root block holding only `okf_version` passes, just as a JSON-typed note already counts conformant under `missing_type`. YAML-only syntax conformance would be a separate rule over the whole audit, not this one. Stated on the attribute docstring and in `docs/design/okf.md` §4, and pinned by the `json` case.

**Report compatibility:** `index_frontmatter` is appended last with an empty default, so `OkfAuditReport` keeps accepting v4.1.0's positional arguments (`test_report_still_constructs_without_the_newer_finding`). The MCP payload only gains a key.

**`log.md` is out of scope, checked against the primary source.** I re-read §8/§9 at the reference page's pinned commit `62432a09` (identical at `open-knowledge-format@0b87c52c`): §9 says nothing about frontmatter. The reference page claimed "none in `log.md`" and "reserved files carry none by §8". Both are corrected, the silence is recorded on the §9 claim, and `docs/design/reference/log.md` records the entry. `test_log_frontmatter_is_not_an_index_finding` pins it.

## Acceptance cases I could not pin here

- **BOM-prefixed index:** the audit reader uses plain `utf-8`, so a BOM hides the block from *every* rule (#1407). Presence and keys read the same string, so #1407's reader fix makes this finding fire with no change here. Adding a BOM test now would pin the defect.
- **Malformed JSON index:** still aborts the whole audit before this rule runs (#1408), as it does on `main`. Presence detection itself does not raise on it (probed), so this PR adds no new crash path. The docs say "YAML that does not parse", not "unparseable", for that reason.

- Change what the policy seeds, what the index gate admits, or whether it reaches generated reserved files: #1438, #1442.
- Fix the BOM-blind readers: #1407.
- Stop malformed JSON frontmatter aborting the audit: #1408.
- Remove the `conflict_with` keys the resolver injects into a root index: #1395.
- Rework the OKF guide's wider maintenance/publication claims: #1449. This PR adds only the one guidance sentence the new finding needs.
- Add a `log.md` frontmatter rule: the format has none.

## Local review

- [x] Ran a local code-review pass on the cumulative diff before `gh pr create`. One finding, fixed before push: the docs originally said an "unparseable" block counts, but a malformed JSON block aborts the audit (#1408). Now "YAML that does not parse". Three over-long prose lines reflowed.
- [x] No `!`. `OkfAuditReport` gains a defaulted field; no env var, config, tool or import changes meaning. A new key in an MCP tool result is additive.

**Review round 1 (@claude, 2026-09-12) — one finding, fixed in `13c08068`.** The pack's own `examples/okf/templates/index.md`, which `examples/okf/README.md` step 1 tells the reader to copy to their vault root, carried `title` and `description` beside `okf_version` — so a vault built exactly as the pack instructs reported a finding on its own root index. Worse, I had trimmed the *synthetic* declaration string in `test_example_templates_pass_okf_validate` rather than checking the real file, which is the reported-half fix: that test exists to guard the shipped templates and never read them. The fixture now copies the real `templates/index.md`, which is what makes this class of drift fail the suite (verified: it fails against the old template, passes against the new one). The template keeps `okf_version` alone and says why, including that a vault with `MARKDOWN_VAULT_MCP_REQUIRED_FIELDS` gets those fields written back by the generators. I swept the rest of the class: `docs/design/reference/index.md` (this repo's own bundle root) already declares `okf_version` alone, the `docs/guides/okf.md` snippet is conformant, and the PARA and Zettelkasten packs ship no index template.

Each branch of the rule is mutation-checked. With presence read from keys instead of the block, the empty/comment/sequence/malformed cases fail. Without the root check, the folder-`okf_version` case fails.

Gates, run locally on `13c08068`: `pytest` full suite (4633 passed, 2 skipped), `ruff check`/`format --check`, `mypy src/ tests/`, `scripts/structural_gate.sh` (100%), `mkdocs build --strict`, Vale, `scripts/check_references.py`.

## Docs impact

- [ ] `README.md` (does not enumerate audit findings)
- [x] `docs/` site pages: `docs/tools/index.md`, `docs/guides/okf.md` (operator guidance: generated indexes now show findings; leave required fields in place), `docs/releases/4.2.md` (paragraph and Upgrading bullet, watermark not moved, following #1406)
- [x] `docs/design/`: `design.md`, `okf.md` §4 and "Generated reserved files", `reference/okf-v0.2.md`, `reference/log.md`
- [x] Inline docstrings: `OkfAuditReport`, the `okf_validate` tool docstring (wire `description=` unchanged)
- [x] `examples/okf/prompts/okf-migrate-vault.md` (finding listed; never propose removing a required field)
- [x] `examples/okf/templates/index.md` — the shipped root-index declaration now carries `okf_version` alone (review round 1)

---


pvliesdonk:
@claude review

pvliesdonk:
Confirmed and fixed in `13c08068` — thank you, this was the reported-half fix I made and did not check.

**The finding stands as written.** `examples/okf/templates/index.md` shipped `title` and `description` beside `okf_version`, and `examples/okf/README.md` step 1 tells the reader to copy that file to their vault root, so a bundle built exactly as the pack instructs reported an `index_frontmatter` finding on its own root index. And the guard that should have caught it, `test_example_templates_pass_okf_validate`, never read the shipped file: I trimmed its synthetic `_DECLARED_INDEX` constant instead, which is precisely the drift its own docstring claims to prevent.

**What changed**

- `examples/okf/templates/index.md` now carries `okf_version` alone. Its comment says why (an index file carries no other frontmatter, and `okf_validate` reports one that does), that the `# My Knowledge Bundle` heading is what names the bundle — the indexer falls back to the first H1 for a title, so nothing is lost by dropping `title` — and that a vault setting `MARKDOWN_VAULT_MCP_REQUIRED_FIELDS` gets those fields written back by the generators, so a reader does not delete them by hand.
- `_build_vault_from_templates` copies the real `examples/okf/templates/index.md` instead of a synthetic string, for all three packs (the OKF pack ships the only declaration template; the PARA and Zettelkasten packs send the reader there to turn OKF on). Verified both directions: the test fails against the old template and passes against the new one, so this class of drift now fails the suite.

**Class swept, not just the instance.** Every `index.md` this repo ships or documents: `docs/design/reference/index.md` (this repo's own bundle root) already declares `okf_version` alone; the `docs/guides/okf.md` declaration snippet is conformant; the PARA and Zettelkasten packs ship no index template, and their READMEs point at the OKF pack. `docs/tools/index.md` and friends are site pages, not bundle indexes.

Your reading that the template was already non-conformant before this PR is right — §8 forbade it and nothing checked. This is the PR that makes it visible, so it is the PR that fixes it. The docs-impact checklist now lists `examples/okf/templates/index.md`.

Full suite green on `13c08068` (4633 passed, 2 skipped), structural gate 100%, lint and mypy clean.

pvliesdonk:
@claude review

PR #1453, opened 2026-09-12
fix(okf): a malformed JSON-looking block is a finding, not an aborted audit

### LINKED ISSUE #1408: [Bug]: a malformed JSON-looking block crashes okf_validate instead of being a finding
### What happened?

`okf_validate` raises `json.JSONDecodeError` and returns no report at all when any audited file opens with a `{` on its own line and the block that follows is not valid JSON.

### Observed

`[verified: probed the pinned python-frontmatter]` `frontmatter.loads` detects a leading `{`-delimited block as JSON frontmatter and calls `json.loads` on it. Malformed JSON raises `json.JSONDecodeError`, which is a subclass of `ValueError`, not of `yaml.YAMLError`:

```python
frontmatter.loads('{\n  not json,\n}\n# body\n')
# json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes
```

`[verified: read src/markdown_vault_mcp/okf.py _audit_file]` The audit's guard is `except yaml.YAMLError`, so the exception escapes `_audit_file`, escapes `audit_bundle`, and fails the whole tool call. One note takes down the audit of the entire vault, and the message names no path.

The same shape reaches `scanner.parse_note` and `DocumentManager.read`, whose guard is `except (UnicodeDecodeError, OSError, yaml.YAMLError)`.

### Expected

A file the parser cannot read is a finding, not an outage: the audit already has `unparseable_frontmatter` for exactly this, and `read` already degrades to `None` with a logged warning.

### Reproduction

```python
(vault / "note.md").write_text("{\n  not json,\n}\n# body\n", encoding="utf-8")
audit_bundle(vault)   # raises instead of reporting
```

A well-formed JSON block parses cleanly and is not affected; the failure needs the block to be malformed.

### Suspected cause

`[verified: as above]` The guards name the YAML handler's exception only, while `python-frontmatter` ships a JSON handler in its default handler list and picks it by the opening delimiter.

### Open questions / scope

- Whether the fix widens the guards to `ValueError` (which covers both, since `yaml.YAMLError` is not one) or names `json.JSONDecodeError` alongside is a decision for the fix.
- `[unverified]` Whether the audit and `read` are the only guards phrased this way; `okf.py`, `scanner.py`, `conventions.py` and `git/conflict.py` all call `frontmatter.loads` and were not each traced.
- Surfaced while implementing #1396, which touches `_audit_file`'s parse. Deliberately not fixed there: an audit that crashes and an audit that under-reports are different defects with different fixes.

Keep in v4.2. Malformed frontmatter must not abort the entire audit. Preserve the existing tolerant behavior of the affected read paths, with a path-identifiable finding or existing failure result. Verify malformed JSON-looking input and continued reporting for other files. Do not turn this into a parser or schema redesign.

---

pvliesdonk:
## Closes / Refs

Closes #1408. Refs #1454 (filed from this work), #1396 and #1407 (the two reader defects surfaced alongside it, both merged), #802 (the `parse_error` / `internal_error` split this restores for a second dialect).

## What & why

`python-frontmatter` picks a handler from the block's opening delimiter and lets that handler's own error escape. YAML raises `yaml.YAMLError`; the JSON handler raises `json.JSONDecodeError`; a TOML handler, registered whenever `toml` is importable in the consumer's environment, raises `toml.TomlDecodeError`. **Every guard in this package named only the first.**

So a file opening with `{` and holding invalid JSON raised straight through `_audit_file`, through `audit_bundle`, and out of `okf_validate`: one note took down the report for the entire vault, naming no path. The same shape escaped `DocumentManager.read`, left `read(section=...)` raising instead of its user-facing `ValueError`, broke the OKF detection probe, lost the conventions resolver's raw-text fallback, and made the indexer record the file as `internal_error` rather than the `parse_error` skip #802 introduced.

**The fix is one parse entry point, not nine wider guards.** `scanner.parse_frontmatter` is now the package's only call into `frontmatter.loads`, and re-raises every non-YAML handler failure as `MalformedFrontmatterError`. A YAML failure still propagates as the `yaml.YAMLError` it already was.

`MalformedFrontmatterError` **subclasses `yaml.YAMLError`**, and that is the load-bearing choice: the failure it describes — *this block did not parse* — is the same one in any dialect, so the nine existing guards, and any downstream consumer's `except yaml.YAMLError`, keep working with no per-site change. The alternative shapes were worse:

- **Widening nine guards to `ValueError`** would have broadened `try` blocks that hold more than the parse — `document.py:372` wraps `parse_note` *and* a second read, `git_query.py:378` wraps two section calls — so a `ValueError` from anything else in them would start being swallowed as "malformed frontmatter".
- **A shared exception tuple** still lets a tenth call site type `except yaml.YAMLError` and be wrong. With the wrapper, a thirteenth parse site cannot be.

Adding `ValueError` as a second base was considered and rejected: it buys nothing today (the one guard phrased that way, `embeddings.py:1118`, already lists both) and would let a future `except ValueError` catch frontmatter failures by accident.

## Verified end to end, not only in unit tests

On a scratch vault, a note whose body is `{\n  not json,\n}\n# body\n`:

- `write` succeeds and the file lands on disk; `process_dirty_paths` logs `skipping bad.md (malformed frontmatter)`; `read` returns `None` and logs `could not parse file`. Before, the same file's read raised.
- With `OKF_WRITE` on, writing that body is still rejected — the provenance stamp cannot be applied to a block it cannot parse. That was already true; only the exception's type changed (`JSONDecodeError` → `MalformedFrontmatterError`). Filed as #1454 rather than widened into here: the refusal is right, its surface is not.

## Tests

Eight, each mutation-checked by narrowing the wrapper's `except ValueError` (all six defect tests fail; the two well-formed-JSON regression guards stay green):

- `audit_bundle` — the bad note is `unparseable_frontmatter` with its path, and the rest of the bundle is still reported.
- `DocumentManager.read` — `None` plus the documented degrade warning naming the path.
- `parse_note_categorized` — `parse_error`, not `internal_error`.
- `ConventionsResolver` — falls back to raw text, exactly as for malformed YAML.
- `OkfDetector` — reports undeclared instead of crashing detection outright.
- `parse_frontmatter` — raises `MalformedFrontmatterError`, which *is* a `yaml.YAMLError`, chaining `__cause__` to the original `JSONDecodeError`; a malformed YAML block still raises the YAML error itself.
- Well-formed JSON frontmatter still parses, in the wrapper and through the audit.

- Touch the nine guards. The wrapper is what makes them correct; changing them too would be the same fix twice.
- Touch `_server_prompts.py` / `git/conflict.py`'s `except Exception` guards: already tolerant. They use the wrapper for consistency, and their behaviour is unchanged.
- Touch `strip_frontmatter_block` / `split_frontmatter_block`: they call `handler.split`, not `loads`, and already catch the `ValueError` it raises.
- Enforce a syntax dialect. A block that parses in any handler the library ships is as valid here as it was before.
- Add `types-PyYAML`. The `# type: ignore[misc]` on the class carries its reason: PyYAML ships no stubs, so `yaml.YAMLError` is `Any` and strict mypy rejects subclassing it; the base is the point of the class, and typing every `yaml` call site in the package is a wider change than this fix.

No bump. No stored row's derivation from a note's bytes changes: a file that was skipped as `internal_error` and is now skipped as `parse_error` produces the same tombstone either way, and every file that parsed before parses identically.

## Local review

- [x] Ran a local code-review pass on the cumulative diff before `gh pr create`. It caught one real design risk in my own first shape — sites that used to catch the JSON error *as a* `ValueError` would stop catching it — so I traced every `except ValueError` in the package and confirmed none wraps a frontmatter parse (`_transfer_sink.py:273` is an attachment read, the two in `_server_prompts.py` guard signature building).
- [x] No `!`. A new name in `scanner` is additive, the package root's import surface is unchanged, and `MalformedFrontmatterError` is a subclass of the error consumers already catch.

Gates, run locally on `dd852e62`: `pytest` full suite (4650 passed, 2 skipped), `ruff check`/`format --check`, `mypy src/ tests/`, `scripts/structural_gate.sh` (100%), Vale, plus `tests/test_config_import_floor.py` for the four modules that gained a `scanner` import.

## Review round 1 (Repowise health gate)

The gate failed on `tests/test_okf.py` (-0.5, "introduced duplicated assertion block"), and it was right: my two new audit tests asserted a near-identical pair, `conformant_notes == 1` plus an `unparseable_frontmatter` count, so neither name matched what it actually checked. Each now asserts its own claim — the crash test checks that *both* files were examined and the readable one was not swept up (`total_notes == 2`, `missing_type.count == 0`), the regression test checks only that a readable JSON block counts as conformant. Re-mutation-checked after the reshape: both still fail when the wrapper's `except ValueError` is narrowed.

## Docs impact

- [ ] `README.md` (does not describe parse-failure handling)
- [x] `docs/` site pages: `docs/releases/4.2.md` — one paragraph covering the audit, `read` and the skip category (watermark not moved)
- [x] `docs/design/`: the `scanner.py` section now states the single-entry-point rule, why the error subclasses `yaml.YAMLError`, and that syntax conformance is not enforced
- [x] Inline docstrings: `MalformedFrontmatterError`, `parse_frontmatter`

---

### DIFF
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -2913,6 +2913,21 @@ def parse_note(path: Path, source_dir: Path) -> ParsedNote: ...
 **Frontmatter parsing**: use `python-frontmatter` library. Schema-agnostic.
 Documents without frontmatter get an empty dict and proceed normally.
 
+**One parse entry point, one parse error (#1408).** Every call into
+`frontmatter.loads` in this package goes through `scanner.parse_frontmatter`.
+The library picks a handler by the block's opening delimiter and lets that
+handler's own error escape — YAML raises `yaml.YAMLError`, the JSON handler
+`json.JSONDecodeError`, a TOML handler (registered whenever `toml` is
+importable in the consumer's environment) `toml.TomlDecodeError` — and every
+guard in the package named only the first, so a `{`-delimited block that was
+not valid JSON aborted the whole OKF audit and escaped
+`DocumentManager.read`. `parse_frontmatter` re-raises the non-YAML failures
+as `MalformedFrontmatterError`, **a subclass of `yaml.YAMLError`**: the
+failure is the same one in any dialect, and the nine existing guards (and any
+downstream consumer's) keep working without a per-site change. A block that
+parses in any supported dialect is unaffected; syntax conformance is not a
+rule this package enforces.
+
 **Exclude patterns**: glob patterns (such as `[".obsidian/**", "_templates/**"]`)
 matched against relative paths from `source_dir` using `pathlib.Path.match()`.
 
--- a/docs/releases/4.2.md
+++ b/docs/releases/4.2.md
@@ -416,6 +416,19 @@ fields to each index it generates, which keeps the file in search. The
 audit reports those fields like any others. Leave them in place. Indexing
 is unchanged. The conformance ratio still counts notes only.
 
+**A note that opens with a brace no longer takes down the audit.** The
+frontmatter library picks its parser from the block's opening delimiter, so a
+file starting with `{` is read as JSON. When that JSON was malformed, the
+error it raised was not the one every caller guarded against, and it escaped
+([#1408](https://github.com/pvliesdonk/markdown-vault-mcp/issues/1408)):
+`okf_validate` returned no report at all for the whole vault, naming no path,
+and `read` on such a file raised instead of returning nothing. Both now treat
+it as what it is. The audit lists the file under `unparseable_frontmatter`
+and reports the rest of the bundle; `read` returns nothing and logs a
+warning; the indexer records the file as a `parse_error` skip rather than an
+`internal_error` one, so `get_index_status.skipped_files` names the real
+cause. A well-formed JSON block still parses as it always did.
+
 **A byte-order mark no longer hides a file's frontmatter.** Every read of
 vault markdown strips a leading UTF-8 BOM, a contract from
 [#673](https://github.com/pvliesdonk/markdown-vault-mcp/issues/673). Three
--- a/src/markdown_vault_mcp/_server_prompts.py
+++ b/src/markdown_vault_mcp/_server_prompts.py
@@ -19,10 +19,10 @@ from typing import TYPE_CHECKING, Any
 if TYPE_CHECKING:
     from collections.abc import Callable
 
-import frontmatter
 from fastmcp import FastMCP
 
 from ._icons import _TOOL_ICONS
+from .scanner import parse_frontmatter
 from .utils.text import read_text_utf8
 
 _BUILTIN_PROMPTS_DIR = importlib.resources.files("markdown_vault_mcp").joinpath(
@@ -173,7 +173,7 @@ def _load_user_prompt_defs(prompts_folder: str | None) -> dict[str, dict[str, An
     for md_file in sorted(folder.glob("*.md")):
         name = md_file.stem
         try:
-            post = frontmatter.loads(read_text_utf8(md_file))
+            post = parse_frontmatter(read_text_utf8(md_file))
         except Exception:
             logger.warning(
                 "Failed to parse user prompt file %r — skipping",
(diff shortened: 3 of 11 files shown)


pvliesdonk:
@claude review

PR #1455, opened 2026-09-12
fix(okf): an enforced-write refusal says which note and why

### LINKED ISSUE #1454: [Bug]: an OKF-enforced write of an unparseable frontmatter block fails with a raw parser error
### What happened?

With `MARKDOWN_VAULT_MCP_OKF_WRITE=true` on an active bundle, writing a note whose body opens with a frontmatter block the parser cannot read fails with the parser's own exception. The write is correctly refused — a provenance stamp cannot be applied to a block that does not parse — but the failure reaches the caller as a parse exception rather than a stated refusal, and its message names no path.

### Observed

`[verified: probed on d7d6bc0d, a declared bundle with okf_write=True]`

```python
vault.writer.write("bad.md", "{\n  not json,\n}\n# body\n")
# markdown_vault_mcp.scanner.MalformedFrontmatterError:
#   Expecting property name enclosed in double quotes: line 2 column 3 (char 4)
```

The stamp enricher (`_okf_write.py:188` → `okf.apply_okf_write_stamp`) parses the text it is about to stamp, and nothing between there and the caller turns a parse failure into a refusal. The same shape was raised as `json.JSONDecodeError` before #1408; that issue normalised the exception type across dialects but deliberately did not change this path.

With `OKF_WRITE` off the write succeeds, the file lands on disk, the indexer records a `parse_error` skip and `read` degrades to `None` — the tolerant behaviour #1408 restored. `[verified: same probe, okf_write=False]`

### Expected

A refusal a caller can act on: the path, and that the note's frontmatter could not be parsed, in the error type this layer already uses for a rejected write. Whether an enforced-write vault should refuse such a write at all, or stamp around an unparseable block, is the substance of the issue rather than a given.

### Reproduction

Declared bundle, `okf_write=True`, `write` any body whose first line is `{` followed by invalid JSON (a malformed YAML block behaves the same way).

### Suspected cause

`[verified: read _okf_write.py and okf.apply_okf_write_stamp]` The enricher runs inside the write path and parses unguarded; the write kernel has no arm for a parse failure originating in an enricher.

### Open questions / scope

- Whether the refusal belongs in the enricher or in the write kernel, and which error type this layer should use, is a decision for the fix.
- `[unverified]` What an MCP client sees for this today: the library-level behaviour above was probed, the tool-level surface was not.
- Surfaced while implementing #1408, deliberately not fixed there: that issue is about read paths tolerating a malformed block, and this is a write path correctly refusing one with the wrong surface.

---

pvliesdonk:
## Closes / Refs

Closes #1454. Refs #1453 (which filed it, and whose review observation this carries), #1408 (the parse-error normalisation that made this path's failure legible), #1174.

## What & why

With `MARKDOWN_VAULT_MCP_OKF_WRITE` on, the enforced-write layer stamps `generated` provenance **into the note's own frontmatter**. A note whose block no parser can read leaves it nothing to stamp, so the write is refused. That was already the behaviour and it is the right one. The surface was not:

```
Error calling tool 'write': Expecting property name enclosed in double quotes: line 2 column 3 (char 4)
```

No note named, no cause, and server-side it logged as an unexpected tool failure. The caller cannot tell a rejected write from a server fault.

The refusal is now raised where the path is known — `DocumentManager._finalize_content_write`, around the enricher call — as the `ValueError` this layer already uses for every other rejected write. `[verified: probed through a live MCP client]` it reads:

```
Error calling tool 'write': Cannot write notes/bad.md: its frontmatter block is not
parseable, so the OKF enforced-write layer cannot stamp provenance on it —
Expecting property name enclosed in double quotes: line 2 column 3 (char 4)
```

**The alternative was writing the note unstamped, and it is worse.** The layer's whole claim is that every write it passes carries current provenance; a silent exception to that is the kind an operator finds months later, in a bundle that says it is enforced. Recorded in `docs/design/okf.md` §6 with that reasoning rather than as a bare fact.

**Unchanged:** with `OKF_WRITE` off there is no stamp to fail — the write lands and the indexer records a `parse_error` skip (#1408). Nothing is written in either the old or the new refusal; the enricher runs before `atomic_write`.

## Carried from #1453's review

`okf.py` imports both `frontmatter as fm` and `scanner.parse_frontmatter`, which the reviewer noted a future reader might stumble over. The import block now says why: `fm` serialises the write stamps (`fm.dumps` / `fm.Post`), every *read* goes through the single parse entry point. The review's other observation — that the `# type: ignore[misc]` rationale for subclassing `yaml.YAMLError` is accurate and well-documented — needed no change.

## Scope check I ran, and what it found

The issue left `[unverified]` what an MCP client sees. Probed, and it is above. I also probed the **log** side, where the issue's "cannot distinguish it from a server fault" claim came from: the traceback panel and `tool_call_failed error_type=ToolError` line are how this server logs *every* refusal — an existing `ValueError` ("Path traversal detected") logs the same shape. So the log side is uniform layer behaviour, not something this path does differently, and I have not touched it. The message is what changed.

## Tests

Six, each mutation-checked (reverting the guard fails all the defect tests; reverting just the `{operation}` interpolation fails only the edit one):

- `write` — refused with a `ValueError` naming the path and the reason.
- The refused write leaves no file on disk.
- Malformed **YAML** is refused identically: the dialect is not what the refusal is about.
- `edit` on an already-malformed note — reaches the stamp (it matches on raw text) and the message names that operation.
- Tool level — an MCP client gets a `ToolError` naming the note and the reason, and no file appears.
- A parseable note still stamps, including one with JSON frontmatter.

- Change *whether* the write is refused. Only how the refusal reads.
- Touch the log shape for refusals: uniform across this server, as probed above. A PR that changes it should change it for all of them.
- Touch `append`: the enricher is a no-op for it (`operation not in ("write", "edit")`), so appending to a malformed note still succeeds unstamped. That asymmetry predates this issue and is not what #1454 reported.
- Migrate or repair notes already on disk with unparseable blocks.

No bump. No stored row's derivation changes: the refused write never reaches disk, and a note written with the layer off is skipped exactly as before.

## Local review

- [x] No `!`. A refusal that already failed now fails with a clearer message and a different exception type on an error path; no env var, config, tool schema or importable name changes meaning.

Gates, run locally on `7cb94690`: `pytest` full suite (4656 passed, 2 skipped), `ruff check`/`format --check`, `mypy src/ tests/`, `scripts/structural_gate.sh` (100%), Vale.

## Review round 1 (Repowise health gate)

The gate flagged `tests/test_okf_write.py` (-0.4, "introduced low cohesion"). Part of that was mine and is fixed: I had dropped the tool-level refusal test into `TestWriteToolThreadsActor`, a class about *which actor string gets stamped*, because the fixture was convenient there. It now lives in its own `TestUnparseableFrontmatterRefusalTool` beside it, sharing the same `enforced_env` fixture; no test content changed.

The gate still flags the file, and the rest is not this diff. `test_okf_write.py` was already 786 lines and 13 classes covering everything from pure stamp builders to three `okf_verify` attribution modes, half on a library `Vault` fixture and half on an MCP one; any new class lowers its cohesion further. Splitting it is a move, not a fix, and doing it here would bury the change — so it is filed as **#1456** with the measurements, per `AGENTS.md`'s rule for decay noticed outside a change's scope. `Repowise / code health` is advisory: the repository's required checks are `Test (3.11/3.12/3.13)`, `Lint`, `Type Check`, `codecov/patch`, `CI Success` and `SPA source up-to-date`.

## Docs impact

- [ ] `README.md` (does not describe enforced-write failure modes)
- [x] `docs/` site pages: `docs/guides/okf.md` (what an operator sees and does about it), `docs/releases/4.2.md` (paragraph; watermark not moved)
- [x] `docs/design/`: `okf.md` §6 records the refusal, why unstamped-write was rejected, and where the error is raised
- [x] Inline docstrings: `_finalize_content_write` gains its `Raises:` entry; `okf.py`'s import block explains its two frontmatter entry points

---

### DIFF
--- a/docs/design/okf.md
+++ b/docs/design/okf.md
@@ -422,6 +422,17 @@ Tool tags following switches: #1412.
   `markdown-vault-mcp/<version>` as a tool actor. Existing `generated`
   values are overwritten (it describes the current bytes); `sources` are
   never touched.
+
+  A stamp is written *into* the note's frontmatter, so content whose block
+  no parser can read leaves nothing to stamp and the write is **refused**
+  (#1454). Writing it unstamped was the alternative and is worse: the layer's
+  whole claim is that every write it passes carries current provenance, and a
+  silent exception to that is the kind an operator finds months later. The
+  refusal is raised where the path is known (`_finalize_content_write`) as the
+  `ValueError` this layer uses for every other rejected write, so a caller is
+  told which note and why rather than being handed the parser's own exception.
+  With `OKF_WRITE` off there is no stamp to fail: the write lands, and the
+  indexer records the file as a `parse_error` skip (#1408).
 - **Verification invalidation** (`OKF_WRITE`): a content-changing `write`/`edit` to a note
   carrying `verified` clears the `verified` list — verification attests to
   bytes that no longer exist. Frontmatter-only edits that do not touch the
--- a/docs/guides/okf.md
+++ b/docs/guides/okf.md
@@ -104,6 +104,8 @@ With the layer on, and only while the vault is an active OKF bundle, every `writ
 
 The one-shot migration transforms above (`okf_convert_links`, `okf_generate_index`, `okf_seed_log`) are exempt: a mechanical rewrite does not re-stamp provenance or discard a human attestation.
 
+The stamp goes into the note's own frontmatter, so the layer refuses a write whose frontmatter cannot be parsed: there is nowhere to record the provenance it promises. The error names the note and the reason, and nothing is written. Fix the block and write again. With the layer off the write goes through, and the note is left out of the index until its frontmatter parses.
+
 ### Recording a human review
 
 Enabling the layer also exposes the [`okf_verify`](../tools/index.md#okf_verify) tool. Call it on a note you have reviewed and it appends a `{by: human:<subject>, at}` entry (`at` a UTC instant) to that note's `verified` list, which promotes the note's trust tier to `human-reviewed`. The verification write is exempt from the invalidation above, so attesting a note does not immediately clear the attestation you just added.
--- a/docs/releases/4.2.md
+++ b/docs/releases/4.2.md
@@ -429,6 +429,18 @@ warning; the indexer records the file as a `parse_error` skip rather than an
 `internal_error` one, so `get_index_status.skipped_files` names the real
 cause. A well-formed JSON block still parses as it always did.
 
+**An enforced-write vault says why it refused a note.** With
+`MARKDOWN_VAULT_MCP_OKF_WRITE` on, the server stamps provenance into each
+note it writes, so a note whose frontmatter block cannot be parsed has
+nowhere to carry the stamp and the write is refused. That much was already
+true; what reached the client was the raw complaint from the frontmatter
+library, naming no note and reading like a server fault
+([#1454](https://github.com/pvliesdonk/markdown-vault-mcp/issues/1454)). It
+now reads as the refusal it is, naming the note, the reason, and the parse
+error behind it, in the same form as every other rejected write. Nothing is
+written either way. With the layer off, the write lands and the note simply
+stays out of the index until its frontmatter parses.
+
 **A byte-order mark no longer hides a file's frontmatter.** Every read of
 vault markdown strips a leading UTF-8 BOM, a contract from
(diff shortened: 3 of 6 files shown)


pvliesdonk:
@claude review

PR #1457, opened 2026-09-12
fix(okf): a migration transform is mechanical through every entry point

### LINKED ISSUE #1401: [Bug]: okf_generate_index stamps provenance when called through the library facade but not through the tool
### What happened?

Calling `okf_generate_index` through the library facade (`vault.writer.okf_generate_index`) stamps `generated` provenance into the regenerated `index.md`, while the same operation through the MCP tool leaves the file unstamped. The two entry points to one mechanical rewrite do not agree.

### Observed

`[verified: read src/markdown_vault_mcp/_server_tools/writer.py and facets/writer.py at 390d4e8b; reproduced while writing tests/test_okf_index_triggers.py on the #1392 branch]` The MCP tool wraps the call in `okf_write_suppressed()`, so the enforced-write enricher is a no-op for it. The facade method does not, so a library caller gets `generated: {by: markdown-vault-mcp/<version>, at: …}` written into `index.md` on every regeneration — and, since `ReservedFrontmatterPolicy.build` keeps existing keys, it then survives every later regeneration. A test fixture that settled listings through the facade had to wrap the calls in `okf_write_suppressed()` itself to get the tool's behaviour.

The same shape applies to `okf_seed_log` and `okf_convert_links` `[unverified]`: the suppression lives in the tool layer for all three, not in the facade or the migration manager.

### Expected

A mechanical transform stamps the same way regardless of which entry point invoked it. `docs/design/okf.md` describes the one-shot migrations as suppressed writes without saying the suppression is a property of the tool layer only.

### Reproduction

```python
col = Vault(source_dir=root, read_only=False, okf_mode="on", okf_write=True)
col.index.build_index()
col.writer.okf_generate_index(folder="guides")
print((root / "guides/index.md").read_text())   # starts with a `generated:` block
```

### Suspected cause

`[verified: as above]` The `okf_write_suppressed()` context is entered in the tool handlers, not around the migration manager's writes.

### Open questions / scope

- Whether the suppression belongs on the facade, on the migration manager, or on each write the manager issues is a decision for the fix.
- Surfaced while fixing #1392; not addressed there.

Mechanical transforms must preserve supplied metadata consistently through library and MCP paths; consistency does not mean extending unwanted stamping to both. Stage 2 of #1425. The original observations above retain their recorded baseline; this update changes the target, not the reported evidence.

Keep in v4.2. Restore the already-documented suppression of knowledge stamping for mechanical transforms consistently through library and MCP entry points. This is independent of the new ordinary-write provenance policy (#1443). Verify both entry points and preservation of existing metadata; do not migrate or clean up historical fields in this fix.

---

pvliesdonk:
## Closes / Refs

Closes #1401. Refs #963 (the transforms), #964 (where the tool-layer suppression was introduced), #1174 (why generated reserved files keep existing keys, which is what made the stray stamp permanent), #1443 (the ordinary-write provenance policy this is independent of).

## What & why

`docs/design/okf.md` exempts the one-shot transforms from the enforced-write layer — *"a mechanical rewrite does not re-stamp provenance or discard a human attestation"* — and never qualifies that by caller. But `okf_write_suppressed()` was entered in the three MCP tool handlers, so the exemption was a property of **the tool layer**, not of the transform.

A library caller of `vault.writer.okf_generate_index` therefore got `generated: {by: markdown-vault-mcp/<version>, at: …}` written into the `index.md` its own migration produced. And because `ReservedFrontmatterPolicy.build` keeps existing keys (#1174), that stamp then survived every later regeneration: the bundle permanently claims the server authored its navigation.

**The suppression moves onto every write `OkfMigrationManager` issues.** Being mechanical is a property of the transform; the entry point is not part of the question. The issue left the placement open between the facade, the manager, and each write — the manager's writes are where it lands, because the facade is a thin delegate and a direct `OkfMigrationManager` caller deserves the same guarantee.

**The three tool-handler wrappers are deleted, not left in place.** A second copy of a rule drifts, and this issue is a picture of that drift. `okf_verify`'s own suppression is untouched: different purpose (it sets `verified` deliberately), different layer.

## The issue's `[unverified]` is now verified

It recorded the defect for `okf_generate_index` and suspected `okf_seed_log` and `okf_convert_links` had "the same shape". Both reproduced before the fix — the seeded `log.md` carried a `generated` block, and a link conversion re-stamped every note it rewrote — and both are covered by tests.

## Tests

Four, in `tests/test_okf_migrate.py` against a declared, `okf_write=True` vault driven through the **library facade** (the entry point that was wrong):

- `okf_generate_index` — the generated `index.md` carries no `generated`.
- `okf_seed_log` — the seeded `log.md` carries none either.
- `okf_convert_links` — a rewritten note keeps its own `generated` (a human's, from 2020) and its `verified` list: the transform must not overwrite provenance it did not create.
- A transform does not make the folder log itself.

Mutation-checked: replacing any one `with okf_write_suppressed():` fails exactly its own test.

That fourth test needs a note in review. I wrote it expecting suppression to be the reason — it passed before the fix, so I instrumented `ConventionMaintainer.maintain` and found it is **never called** for a transform: `WriterFacet` invokes the maintainer for ordinary `write`/`edit`/`append` only, and the transforms delegate straight to the manager. So the property holds by routing, not by suppression. The test stays, with a comment saying that, because nothing else pinned it and the suppression this PR moves would otherwise look like the reason.

- Clean up `generated` blocks already written into reserved files by a library-path migration. The triage says not to migrate historical fields here; a regeneration overwrites the body but keeps the key (#1174), so removing them is its own decision.
- Change the ordinary-write provenance policy: #1443.
- Change what the transforms write, their frontmatter preservation, or the reserved-frontmatter policy.
- Touch `okf_verify`'s suppression, or the convention maintainer's own.

No bump. Stored rows derive from a note's bytes exactly as before; this only stops a key being added to bytes the server writes.

## Local review

- [x] Ran a local code-review pass on the cumulative diff before `gh pr create`. It caught two things in my own diff: I had truncated an unrelated `# pragma: no cover` comment to fit the new indentation (restored as `deleted mid-scan`, same meaning), and the module docstring still described the transforms without the suppression that now lives in them (extended).
- [x] No `!`. Library callers of `vault.writer.okf_*` see one fewer frontmatter key on files the transform writes — the documented behaviour they should already have had. No env var, config, tool schema or importable name changes meaning.

Gates, run locally on `dc7dcf80`: `pytest` full suite (4660 passed, 2 skipped), `ruff check`/`format --check`, `mypy src/ tests/`, `scripts/structural_gate.sh` (100%), Vale.

## Docs impact

- [ ] `README.md` (does not describe the transforms' stamping behaviour)
- [x] `docs/` site pages: `docs/releases/4.2.md` (paragraph, including that existing stray stamps are left alone; watermark not moved). `docs/guides/okf.md` already states the exemption without qualifying it by caller, which is exactly what the code now does.
- [x] `docs/design/`: `design.md`'s phase-4 entry and `okf.md`'s maintenance bullet both say where the suppression lives and why
- [x] Inline docstrings: `managers/okf_migrate.py`'s module docstring

---

### DIFF
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -1496,7 +1496,11 @@ the spec itself says, dated and sourced, in
   bundle root seeds whole-vault history (#974). The
   tools are tagged ``{"okf", "write"}`` — hidden in read-only mode and
   under ``OKF_MODE=off`` — and gate on read-only only, not a future
-  ``OKF_WRITE`` flag (they are migrations, not enforcement). Bundle export
+  ``OKF_WRITE`` flag (they are migrations, not enforcement). Every write
+  a transform issues is entered under ``okf_write_suppressed`` **in the
+  manager**, not in the tool handler (#1401): being mechanical is a
+  property of the transform, so a library caller of ``vault.writer.okf_*``
+  gets the same unstamped rewrite the tool does. Bundle export
   ships as an overloaded download ref rather than a bespoke tool: the domain
   ``VaultTransferSink`` recognises an ``okf-bundle`` / ``okf-bundle:<folder>``
   ref on pvl-core's ``create_download_link`` and generates the archive at
--- a/docs/design/okf.md
+++ b/docs/design/okf.md
@@ -486,7 +486,11 @@ Tool tags following switches: #1412.
   `write` / `edit` on an OKF-active vault, skips a write whose target is
   itself a reserved file (`index.md` / `log.md`) so it never recurses, and is
   skipped for suppressed writes (`okf_verify`, the one-shot migrations) so an
-  attestation or mechanical rewrite does not churn the reserved files. The
+  attestation or mechanical rewrite does not churn the reserved files. A
+  transform suppresses each write it issues from inside `OkfMigrationManager`,
+  so the exemption is the transform's own and does not depend on which entry
+  point called it (#1401); it had been entered in the MCP tool handlers alone,
+  which left a library caller's bundle stamped by its own migration. The
   affected folder is the one directly containing the written note; the
   `index.md` refresh reuses the migration `generate_index`, draining the
   single-writer index first so a just-created note is listed. A brand-new
--- a/docs/releases/4.2.md
+++ b/docs/releases/4.2.md
@@ -429,6 +429,20 @@ warning; the indexer records the file as a `parse_error` skip rather than an
 `internal_error` one, so `get_index_status.skipped_files` names the real
 cause. A well-formed JSON block still parses as it always did.
 
+**A migration transform stays mechanical for library callers too.** With
+`MARKDOWN_VAULT_MCP_OKF_WRITE` on, the one-shot transforms
+(`okf_convert_links`, `okf_generate_index`, `okf_seed_log`) are exempt from
+provenance stamping: a mechanical rewrite is not authorship. The exemption
+was entered in the MCP tool handlers only, so the same call through the
+Python API stamped `generated` into the files it rewrote. Because generated
+reserved files keep their existing keys, that stamp then survived every later
+regeneration
+([#1401](https://github.com/pvliesdonk/markdown-vault-mcp/issues/1401)). Each
+transform now suppresses the stamp on every write it issues, so both entry
+points agree. Existing frontmatter, including a note's own `generated` and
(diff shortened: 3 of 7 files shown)


pvliesdonk:
@claude review

PR #1479, opened 2026-09-14
refactor(git)!: remove deprecated claim constructor arguments

### LINKED ISSUE #1236: [Feature]: remove the deprecated commit_name_claim/commit_email_claim GitWriteStrategy kwargs at the next major

`GitWriteStrategy.__init__` still accepts `commit_name_claim` and `commit_email_claim`, marked `.. deprecated::` in its docstring since #1226. They no longer do what their names say: OIDC claims are resolved into a `Principal` at the MCP tool edge (#1160) rather than by the strategy, and the kwargs now feed only a startup warning in `_check_identity` when git's own `user.email` is unset. [verified: read `src/markdown_vault_mcp/git/strategy.py` constructor and `_check_identity` at 8e20b25]

A parameter that is named for a job it no longer does is a trap for the next reader of that constructor.

## Desired outcome

The constructor's parameter list reflects what the strategy actually uses. The startup identity warning either keeps working from a source that is not a vestigial kwarg, or is dropped deliberately.

## Implementation ideas

Non-binding. Deferred to a major because these kwargs shipped in stable v4.1.0 and `GitWriteStrategy` is importable, so removing them is a public-library break under `AGENTS.md`'s two-surface test — the same reason #1225 defers the `to_vault_kwargs` removal. [verified: `git merge-base --is-ancestor` places the #1226 deprecation after the v4.1.0 tag]

Worth checking at the same time whether `_check_identity`'s warning is still earning its place, and whether it can read the claim keys from the identity layer that now owns them rather than from constructor arguments.

Refs #1226, #1231.

pvliesdonk:
## Closes / Refs

Closes #1236.

## What & why

`GitWriteStrategy` no longer accepts the deprecated `commit_name_claim` and `commit_email_claim` constructor arguments. Author claims continue to be resolved at the request boundary and carried into writes as a `Principal`; configuration assembly stops forwarding claim keys to the strategy.

The old startup `git config user.email` probe and warning are deliberately removed. The strategy supplies a configured/default committer to writes and rebases, so an empty checkout identity is supported. The identity layer still reports configured claims that an authenticated token cannot supply.

- Change service-token author classification: #1463 remains separate correctness work.
- Prepare the package's combined release narrative: that remains tracked on #1225. This PR supplies the Git API migration instructions for inclusion there.

## Design

Removing parameters from the middle of a positional signature could silently reinterpret old claim values as `git_lfs` and `repo_path`. Those two retained options are now keyword-only, and tests reject the old positional forms as well as both removed keywords (including `None`). The earlier positional parameters retain their ordering.

Static committer identity, per-write Principal propagation, operator claim configuration, and the request-side unusable-claim diagnostic retain their existing behavior. No identity-layer dependency is added to Git to preserve the obsolete startup suppression logic. README/API/design/reference descriptions are updated, including related identity docstrings.

## Validation

- `uv run pytest -x -q --cov=src/markdown_vault_mcp --cov-report=term-missing --cov-report=xml` — **4,804 passed, 3 skipped**, **97.41%** total coverage.
- `uv run diff-cover coverage.xml --compare-branch=origin/main --fail-under=80` — passes; the production diff has no newly executable lines (deletions, signature changes, and documentation).
- Focused Git/identity tests — **258 passed**, including real default-committer writes without checkout/global identity, rejection of old keyword/positional forms, and dispatcher Principal propagation.
- Ruff check/fix → format → format check, `uv run mypy src/ tests/`, and `uv run pre-commit run --all-files` — passed.
- `bash scripts/structural_gate.sh` — **100%**, zero new violations.
- `uv run mkdocs build --strict` — passed.
- Stable manifest versions remain unchanged at **4.2.0**.

## Local review

- [x] Ran `$code-review` over the cumulative diff before publication.
- [x] The `!` removes a shipped public-library surface: stable `v4.2.0` accepts the claim arguments and positional trailing options.

### Self-review 80d560ea..6bae917e

Charters: written rules, diff, adjacent commitments, recorded intent, normative applicability. Coverage: all ten changed files, with independent diff, full surrounding-file, and history passes. Verified unchanged-content reuse from earlier full reads was limited to unchanged files/regions. No formal schema or RFC contract changes; optional past-PR charter skipped.

No outstanding important or blocker findings. The class sweep found three descriptions of the removed startup warning in the design and identity docstrings; all were corrected in `6bae917e` and re-reviewed. History checks confirm the announced removal and preserve the intent of prior identity fixes.

## Docs impact

- [x] `README.md`: Git migration link inside a DOMAIN section.
- [x] `docs/` site pages: removed keywords, keyword-only trailing options, author identity, and warning removal.
- [x] `docs/design/`: final identity ownership and warning decision; reference's internal behavior/test pointers updated.
- [x] Inline docstrings: current constructor and request-side diagnostic.
- [x] Enduring decisions reflected in the design specification.

---


pvliesdonk:
@codex review

PR #1482, opened 2026-09-14
fix(index): refresh queued writes before index-dependent mutations

### LINKED ISSUE #1464: [Bug]: okf_convert_links skips a wikilink whose target was written moments earlier
### What happened?

Expected `okf_convert_links` to convert a wikilink whose target note was just written successfully; got the link counted as skipped and left as a wikilink.

### Observed

On the 4.2.0-rc.3 obsidian-vault deployment (commit `6c900898`, 2026-09-13):

1. `write` of `_rc3-livetest/wiki-target.md` → success.
2. Immediately `write` of `_rc3-livetest/wiki-source.md` containing `See [[wiki-target]].` → success.
3. `okf_convert_links` with `folder="_rc3-livetest"` returned:

```json
{"files_changed": 0, "links_converted": 0, "links_skipped": 1, "notes_scanned": 2}
```

with the wikilink left untouched. The identical sequence with ~8 seconds between the writes and the convert returned `{"files_changed": 1, "links_converted": 1, "links_skipped": 0}` and rewrote the link as `[wiki-target](/_rc3-livetest/wiki-target.md)`. On two other rc.3 deployments (work-vault, corpus) the immediate sequence converted on the first try, so the window varies by deployment.

The `write` tool's docs say "The search index is updated immediately; do not call 'reindex' afterward", while `okf_convert_links` converts "only links whose target is indexed". Under this race both claims cannot hold at once for the same note.

### Expected

A link whose target's `write` returned success converts without an arbitrary settle delay.

### Reproduction

As above, against the deployed instance. Version checked: 4.2.0rc3 (`6c900898`).

### Suspected cause

I have not verified the cause. [verified: read `managers/document.py` `_finalize_content_write` at tag `v4.2.0-rc.3`] the write path calls `_mark_paths_dirty([path])` rather than writing index rows inline. [unverified]: the dirty-path flush is asynchronous, and the convert tool's "is the target indexed" resolution read the index before the flush landed — a settle race whose window depends on vault size and flush timing.

### Open questions / scope

- [unverified]: whether other index-reading tools (`get_backlinks`, `get_broken_links`, the rename link-updater) can observe the same stale window after a write.
- Not about: wikilink resolution rules — the link resolved and converted correctly once the index settled.

---

### LINKED ISSUE #1483: [Decay]: Index build completion ownership is split across entry points
### What

Index builds expose separate job-completion and readiness-completion signals, with finalization repeated across synchronous, asynchronous, and legacy background entry points.

### Where

Observed at `cca011588f9044a6aaafda99e38d76fc9f0b681c` in `src/markdown_vault_mcp/indexing/coordinator.py`: `build_index`, `build_index_async`, `_on_build_index_done`, and `start_background_build_index`; readiness publication lives in `indexing/readiness.py`.

[verified: source inspection] The synchronous caller publishes the persistent completion marker and readiness after the writer Future resolves. The asynchronous path publishes them in a Future callback, while the legacy background wrapper also manages completion/error state around the synchronous path.

[verified: held-finalization regression tests in PR #1482] Holding `set_build_completed` after a successful writer build allowed link conversion, index generation, and convention maintenance to report `never_built` through the synchronous and legacy entry points. Six cases failed before the correction. The correction waits for readiness finalization as well as the queued refresh; the regression suite covers all three entry points.

Review evidence: https://github.com/pvliesdonk/markdown-vault-mcp/pull/1482#discussion_r4003367426. The immediate behavior fix belongs to #1464 / #1482.

### Why it compounds

[verified: source inspection and the regression above] FIFO writer completion alone does not establish the readiness required by dependent operations. Each new caller must know about both completion boundaries, while failure, cancellation, and readiness publication remain distributed across the build entry points. Existing lifecycle fixes (#585, #590, #591, #598) and this review correction touch different parts of that coordination.

This records structural debt beyond the immediate correction. It is distinct from #1361, which concerns build-provenance arguments.

### Suggested direction

A useful question for a later design pass is which existing completion signal can represent the full build outcome consistently for callers. Any resulting lifecycle contract would need to preserve current cancellation/failure behavior and be reflected in `docs/design/design.md` and the library documentation. No particular abstraction or implementation is selected here.

---

pvliesdonk:
## Closes / Refs

Closes #1464. Closes #1483.

## What & why

A note written immediately before link conversion could be absent from the background index, causing a valid target to be skipped. Conversion, index generation, rename with `update_links=True`, and folder move now refresh prior writes before using index data. Failed refresh work remains retryable, while healthy sibling notes receive vector updates without repeated provider requests for unchanged content.

Build completion now has one owner. Previously, build-job completion, marker publication, and readiness could disagree; a failed build could also let rename or folder move proceed against an incomplete backlink graph. Every dependent mutation now checks the captured build outcome before changing files.

No part of #1464 or #1483 is deferred. Remaining package work stays separate: conflict-loop diagnostics (#1382) and combined migration guidance (#1225).

## Design

`BuildLifecycle` owns scheduling, execution, persistent completion markers, and readiness publication. Synchronous builds wait on the same complete-build Future returned by asynchronous builds. The legacy one-shot entry point uses the same writer, without another thread. Async marker failures now fail the Future instead of returning a successful scan result with failed readiness.

Each attempt retains its own outcome. The latest scheduled attempt alone updates query readiness; an older completion cannot make a newer rebuild appear ready. Warm-reuse checks preserve an existing queryable verdict and clear it only when a cold build is needed. Build capture and refresh submission share a scheduling lock. The mutation barrier requires the captured attempt to succeed and the queued refresh to finish within one 60-second deadline. A `TimeoutError` raised by refresh work propagates as the job error; only expiration of the wait budget gets the generic refresh-timeout message. No scheduled build preserves disk-only rename/move support; failed, cancelled, and interrupted attempts reject dependent mutations. Existing cancellation and error classification remain intact. Concurrent edits or builds started after the barrier remain outside its prior-work guarantee.

The writer notifies cancelled callers outside its submission lock, rejects all submissions after a crash, and joins an already-closing worker before downstream teardown. Normal shutdown still permits internal flush follow-ups.

Deferred embedding work compares prepared metadata with stored vectors, preserves matching rows, and retries failed saves without another provider request. Vector-load failures retain pending work. Convention maintenance preserves the primary note/log and existing listing when its secondary refresh fails. Link derivation and stored formats are unchanged; no index-semantics bump is needed.

The lifecycle state table and ownership contract are in `docs/design/design.md`; `docs/design/reference/python-futures.md` records the relevant Python Future behavior and its test pins.

## Local review

- [x] Ran `$code-review` on the cumulative diff before this push.
- [x] No `!` marker: public signatures, operator options and stored formats remain compatible; failures now propagate through the intended build outcome.

Self-review `a649a6da..08361a39`: rules, cumulative diff, full adjacent commitments, recorded intent, and prior PR findings covered across all 33 changed files. Prior full-file reads were reused only where unchanged; the complete lifecycle refactor and final review delta were reviewed afresh. The new reference passes the repository conformance checker. The stale legacy-thread docstrings found during review were corrected. The two latest findings now have race-focused regression tests; no unresolved important findings remain.

Validation: **4,949 passed, 3 skipped** in the full suite; **100% local patch coverage** and **97.63% total coverage including branches**. The new lifecycle module has **100% line and branch coverage**. Ruff check/format, mypy (`src/` and `tests/`), all-file pre-commit, strict docs build, reference checks, and the structural diff gate (**100%, zero violations**) pass. The PR adds 123 regression cases across freshness, build outcomes/finalization, embedding reuse, and writer cancellation. Warm-restart tests verify that scanning is skipped and that reuse checks preserve queryability; the environment-timeout test checks the configured deadline without relying on traceback-rendering speed.

## Docs impact

- [x] `README.md`: disk completion versus background indexing.
- [x] `docs/` site pages: library build completion, freshness, and convention maintenance.
- [x] `docs/design/`: lifecycle ownership, state policy, FIFO refresh, retries, and shutdown; module map updated.
- [x] External reference: Python Future completion/cancellation, checked against official docs and pinned CPython source.
- [x] Inline docstrings and enduring decisions updated in the same change.

---


pvliesdonk:
@codex review

pvliesdonk:
@codex review

pvliesdonk:
@codex review

pvliesdonk:
@codex review

pvliesdonk:
@codex review

pvliesdonk:
@codex review

PR #1363, opened 2026-09-06
fix(git): rebase with the configured commit identity, not the checkout's

FoundationOperations:
## Closes / Refs

Closes #1362

## What & why

The pull pipeline's `git rebase` (and `rebase --continue`) ran with no committer identity, so in the published container — no `~/.gitconfig`, served clone with no `user.*` — the first divergence stopped at commit 1 with `Committer identity unknown`, zero unmerged paths, and the resolver's "loop exceeded" line pointed at a merge conflict that never happened; pushes were non-fast-forward from then on while `write`/`edit` kept returning success. `git_env()` now takes the configured identity and exports it as `GIT_AUTHOR_*`/`GIT_COMMITTER_*`, `_git_env()` always passes the strategy's commit name/email, the rebase fallback logs git's own stderr, and the resolver says plainly when it stopped with no unmerged paths.

Reproduced and verified on a live deployment: a three-writer race wedged the door on `main`'s code path; the same race after giving the served clone a repo-level identity self-healed (`ff-only not possible, rebased local commits onto upstream`). The new test fails on `main` with `reason='conflict_resolution_failed'` and passes with this change.

## What this PR deliberately does NOT do

- (deferral, #1381): it does not change `_check_identity`, which still reads only `git config user.email` and so still warns where `GIT_COMMITTER_EMAIL`/`EMAIL` would satisfy git — noted as a departure in `docs/design/reference/git-staging-and-commits.md`.
- (deferral, #1382): it does not split the "conflict resolution loop exceeded" message into its two cases (cap reached vs. exited on the first pass); the maintainer commit's DEBUG line before it and the `cause=` on the transition line make the two distinguishable in the log.
- (observation, not touched): `tests/test_git_force_methods.py::TestPullDivergenceClassification::test_an_available_fast_forward_can_still_fail_on_a_dirty_tree` fails on unpatched `main` on the contributor's machine (git 2.55.0, macOS) — the rebase succeeds where the test expects it to refuse. It passes on CI's Linux runners; pre-existing and independent of this change.

## Maintainer commit (1527e0c0)

Pushed by the maintainer so the PR merges with contributor credit. The branch was rebased onto `main` (after #1379, and again after #1380 to clear a `log.md` conflict; both sides kept) before the commit was added; the two contributor commits are unchanged in content. The commit resolves the findings of the local review of the cumulative diff:

### Self-review 957565b3..1527e0c0

Charters: rules, diff, comments, history, conformance (none: no normative content), past-PRs (#1345's review threads).
[...]

PR #1488, opened 2026-09-14
fix(git): distinguish conflict resolver exit diagnostics

### LINKED ISSUE #1382: [Bug]: "conflict resolution loop exceeded" is logged for two different exits of the resolver
## What happened?

`resolve_rebase_conflicts` (`src/markdown_vault_mcp/git/conflict.py`) ends with one line for every path out of its loop:

```
Git pull: conflict resolution loop exceeded 50 iterations
```

It is emitted both when the loop genuinely reached its 50-iteration cap and when it left on its first pass because `git diff --name-only --diff-filter=U` listed no unmerged paths. The two situations read identically in the log.

Raised as an open question on #1362, where the second exit was the whole story: the rebase had stopped for a missing committer identity with nothing unmerged, and the line sent the operator hunting for a merge conflict that never happened. #1363 adds a DEBUG line naming the no-unmerged-paths exit and carries the rebase's stderr onto the sync-health transition line, so the two are now distinguishable at DEBUG; the loop-cap line itself still says the same thing for both.

## Observed

- `[verified: reading the function]` the `for` loop's `break` on an empty `conflicting` list and its natural exhaustion after 50 iterations both fall through to the same `logger.debug("... loop exceeded %d iterations", max_iterations)`.
- A third path reaches the same line: `git rebase --continue` returning non-zero on every iteration until the cap, which is the case the message actually describes.

## Expected

A log line that says which of the exits happened: cap reached, or nothing left unmerged after a rebase that stopped for another reason.

## Reproduction

1. Diverge a managed clone from its remote and make `git rebase` stop before its first replayed commit without a conflict — a `pre-rebase` hook that exits non-zero does it (`tests/test_git_health.py::TestPullFailureIsDiagnosable` on #1363 sets one up).
2. Run a pull at DEBUG level and read the resolver's last line.

## Suspected cause

`[verified: reading the function]` one exit line serves every way out of the loop. I have not verified whether any operator runbook or alert keys on the exact wording.

## Open questions / scope

- `[unverified]`: whether the line should stay at DEBUG when it records the cap (a repeating per-cycle detail, #1287) — the level question is separate from the wording one.

---

pvliesdonk:
## Closes / Refs

Closes #1382

## What & why

The conflict resolver logged its iteration-cap event after both natural loop exhaustion and an early exit with no unmerged paths. It now returns directly from the no-unmerged-paths exit and gives each exit a distinct structured DEBUG event.

- Change per-cycle DEBUG diagnostics into operator-facing transition logs; #1287 defines the transition-only ERROR policy.
- Change rebase identity diagnosis or stderr propagation; #1363 owns that behavior.

## Design

The no-unmerged-paths branch returns the saved conflict records immediately. Only 50 failed conflict-resolution iterations can reach `git_pull_conflict_resolution_exhausted`.

## Local review

      existed at the **last stable release** (see the breaking-change policy
      in `AGENTS.md`) — this PR carries no `!` commit.

Self-review range: `0a3087f1..a62c45b1`. The rules, cumulative diff, full-file context, and history charters were clean. The change has no normative schema or wire-contract content. Past-PR review did not apply because this branch had no PR.

## Validation

- `uv run pytest -q --cov=src/markdown_vault_mcp --cov-report=term-missing --cov-report=xml:/tmp/1382-coverage.xml` — 4,949 passed, 3 skipped; 97.65% total coverage
- `uv run diff-cover /tmp/1382-coverage.xml --compare-branch=origin/main --fail-under=80` — 100% patch coverage
- Required Ruff sequence, mypy, pre-commit, and `scripts/structural_gate.sh` — passed

## Docs impact

- [x] `README.md` — checked; no user-facing behavior changed.
- [x] `docs/` site pages — checked; no user-facing behavior changed.
- [x] `docs/design/` — documents both resolver exit events.
- [x] Inline docstrings — checked; no public API changed.
- [x] Ported lasting decisions the change explains; the event distinction is recorded in `docs/design/design.md`.

---


pvliesdonk:
@codex review

PR #1514, opened 2026-09-17
fix(links): percent-encode spaces in destinations the server writes

### LINKED ISSUE #1494: [Bug]: generated and rewritten Markdown links contain unescaped spaces and fail in Obsidian
## What happened?

Server-generated and rewritten Markdown links can contain unescaped spaces in their destinations: the vault indexer accepts them, but they are invalid CommonMark links and fail in Obsidian.

## Observed

An operator reported an agent-authored link to a note under a directory containing spaces that failed in Obsidian. Investigation found that the server itself produces the same form. The examples below use synthetic paths rather than private vault content.

[verified: direct calls at `63680757`, 2026-09-14] All three paths produce a plain destination with raw spaces:

| Producer | Input | Output |
| --- | --- | --- |
| OKF index generator | Note at `/Project Notes/target.md` | `- [Target](/Project Notes/target.md)` |
| OKF wikilink converter | `[[Target]]` resolved to `Project Notes/target.md` | `[Target](/Project Notes/target.md)` |
| Rename destination rewriter | Plain `/target.md` link; new target `Project Notes/target.md` | `/Project Notes/target.md` as the replacement destination |

[verified: direct calls to `scanner.extract_links` at the same commit] The raw-space, angle-bracketed, and percent-encoded forms all resolve to `Project Notes/target.md`. This means successful resolution through our own indexer does not establish that generated output is valid Markdown.

[verified: CommonMark 0.31.2 §6.3 and Obsidian Help, checked 2026-09-14] Plain Markdown destinations cannot contain spaces; angle-bracketed destinations can. Obsidian documents `%20` encoding and angle brackets. The repository's existing Obsidian reference also records a desktop 1.13.7 observation that `[w](Sub/My Note.md)` is not a link. Obsidian was not rerun during this investigation.

- [CommonMark link destinations](https://spec.commonmark.org/0.31.2/#link-destination)
- [Obsidian formatting guidance](https://obsidian.md/help/syntax)
- `docs/design/reference/obsidian-markdown.md`, “Markdown links Obsidian writes”
- `docs/design/reference/commonmark-gfm.md`, “Inline links”

## Expected

Links the server generates or rewrites remain valid Markdown when a note or directory name contains spaces. They retain the intended target and path convention, including root-absolute paths. Valid examples are `[Target](/Project%20Notes/target.md)` and `[Target](</Project Notes/target.md>)`.

The scanner may remain tolerant of existing malformed content. Tolerant ingestion does not determine what the server should emit. Verification of generated output establishes Markdown syntax validity as well as target resolution; the tolerant indexer alone cannot catch this defect.

## Reproduction

The following function calls were run at `63680757`:

```python
from markdown_vault_mcp.okf import (
    build_index_markdown,
    convert_wikilinks_to_markdown,
)
from markdown_vault_mcp.types import OutlinkInfo
from markdown_vault_mcp.utils.links import compute_new_raw_target

build_index_markdown(
    "Index", [("Target", "/Project Notes/target.md", None)]
)
# '# Index\n\n- [Target](/Project Notes/target.md)\n'

convert_wikilinks_to_markdown(
    "[[Target]]",
    [OutlinkInfo(
        target_path="Project Notes/target.md",
        link_text="Target",
        link_type="wikilink",
        raw_target="Target",
        exists=True,
    )],
)[0]
# '[Target](/Project Notes/target.md)'

compute_new_raw_target(
    "markdown", "/target.md", None, "Project Notes/target.md",
    source_path="source.md", old_path="target.md",
)
# '/Project Notes/target.md'
```

## Suspected cause

[verified: read `okf.py`, `managers/okf_migrate.py`, and `utils/links.py` at `63680757`] The OKF converter and index builder interpolate paths into plain Markdown destinations without escaping spaces. `generate_index()` supplies raw note paths to the builder. The rename rewriter preserves the original destination form and introduces no space escaping when the original was plain and unencoded.

[verified: read `docs/design/design.md`, “Link Extraction”, and the decision on #1353] Accepting spaces during ingestion and preserving unencoded output were explicitly retained in the earlier destination-parser work. This is a remaining output-compatibility problem, not a claim that #1353 regressed.

## Open questions / scope

- Scope: server-generated OKF index links, wikilink conversion, and rename/move destination rewriting for paths containing spaces. General Markdown validation, rejecting arbitrary writes, and automatically repairing existing prose are outside this report.
- Documentation impact: the design's output-preservation policy needs to reflect the corrected behavior. Relevant README/operator guidance, model-facing authoring instructions, and examples should distinguish a raw file path from a valid Markdown destination.
- The existing dated references cover the external behavior; implementation should keep their test pointers and documented departures accurate.
- Related: #1332 covers reading percent-encoded destinations; #1353 covers other destination spellings; #1358 records the Obsidian observations.
- Release scope: the owner explicitly requested milestone 12, `010 reliable-vault`.

---

pvliesdonk:
## Closes / Refs

Closes #1494

## What & why

A vault path is a name; a plain markdown destination is not. All three producers the issue names — the OKF index builder, the OKF wikilink converter, and the rename/move link rewriter — interpolated a path into `[t](…)` verbatim, so a note or folder whose name carries a space produced a destination that ends at that space. Invalid under CommonMark 0.31.2 §6.3, and not a link in Obsidian either (`docs/design/reference/obsidian-markdown.md` records the 1.13.7 observation that `[w](Sub/My Note.md)` is no link).

The defect survived because this project's own scanner accepted the output: the scanner is deliberately tolerant and resolves the raw-space, pointy and percent-encoded spellings alike, so resolution through the index never established that the emitted markdown parses. Ingestion tolerance is unchanged; what the server *emits* is now a separate, explicit decision.

`utils/links.py::encode_plain_destination` is the write-side counterpart to `decode_link_target`. It touches only what §6.3 forbids — the space and the ASCII control characters — leaving `(`, `)`, `[`, `]`, `#` and non-ASCII literal, the same restraint Obsidian's own link writer shows. The rename path applies it as a second bounded repair beside the existing `#` escape, and only in the plain form: `<…>` already holds a space, and a percent-encoded original is re-encoded by `quote` as before.

The server's model-facing instruction now says `%20 for spaces` too, since the reported link was agent-authored rather than server-generated.

Before → after, for a note under `Project Notes/`:

```
- [Target](/Project Notes/target.md)     →  - [Target](/Project%20Notes/target.md)
[Target](/Project Notes/target.md)       →  [Target](/Project%20Notes/target.md)
/Project Notes/target.md                 →  /Project%20Notes/target.md
```

- (existing prose is not repaired): out of scope per the issue; a vault's already-written literal-space links stay as they are and stay readable, because the scanner is unchanged.
- (other link metacharacters in generated output): #1513 — a generated index line is still not a link when the path carries a `#` or an unbalanced `(`, or the title carries a `]`. Verified and filed separately rather than widened into here.
- (the rename rewriter's unbalanced-paren case): unchanged, and `docs/design/design.md` still records it as knowingly unrepaired.
- (a fragment the author wrote): re-attached as found on rename. The rewrite repairs the name it introduces, not prose it was not asked to touch. The OKF converter does encode fragments, because there it *generates* them from a wikilink, which legitimately holds spaces — the asymmetry is commented at both sites.
- (a worked example in the server instructions): the hint is `%20 for spaces`, not a spelled-out path. The reviewed instructions budget (#1253) was at 98% of its 1536 UTF-16 ceiling before this PR; the long form cost 54 units and broke it, the terse form costs 16. A comment at the site records why, so the next editor does not expand it back. Raising the budget is a decision for its owner, not a side effect of this fix.
- (`INDEX_SEMANTICS_VERSION` bump): not applicable. Nothing about how a note's stored rows are derived from its bytes changed; only what the write paths emit.

## Design

Small bug fix; the decisions are recorded in `docs/design/design.md` (the rewrite-spelling section gains its second bounded exception, and the OKF migration-transforms section gains the encoder), and pinned from `docs/design/reference/commonmark-gfm.md` and `docs/design/reference/obsidian-markdown.md`.

One test-design note worth stating, because the issue makes the point: resolution through our own index does not prove the output is markdown. `tests/test_links_generated_destinations.py` therefore asserts the emitted spelling as well, and carries a parametrised property test that no producer emits a character §6.3 forbids. Mutation-checked — reverting `encode_plain_destination` to the identity fails 14 of its 24 tests, including all three of the issue's table rows.

## Local review

- [x] Ran a local code-review pass on the cumulative diff before opening this PR. It surfaced one finding: the `links.py` module docstring named only the read side after a write-side function was added. Fixed in `dc23bf4`, along with the module-map entry. No blockers or importants survived refutation.
- [x] No commit carries `!`. Nothing an operator sets changes; `encode_plain_destination` is additive and `tests/public_import_surface.txt` is untouched. The behaviour change is confined to output that was invalid markdown before.

## Docs impact

- [ ] `README.md` — no user-facing env var, tool, or option changed; the README documents neither OKF transform.
- [x] `docs/` site pages — `docs/guides/okf.md` and `docs/tools/index.md` now state the encoding for both OKF transforms.
- [x] `docs/design/` — `design.md` (both sections), `okf.md`, `module-map.md`, and pins added to the two reference pages.
- [x] Inline docstrings — `encode_plain_destination`, `build_index_markdown`, `compute_new_raw_target`, and the `links.py` module docstring.
- [x] Ported enduring decisions: the "no encoding where the author used none" rule now has two bounded exceptions rather than one, stated in `design.md` with the reason each repair is forced.

## CI round on `dc23bf4`

The test matrix went red on all three of 3.12/3.13/3.14 with the same two assertions, and they were this PR's:

```
test_client_surface_budget.py::test_maximal_client_surface_stays_within_reviewed_budgets
  client-surface category budgets exceeded: {'instructions': (1560, 1536)}
test_server.py::TestServerIdentity::test_maximal_generated_instructions_fit_target
  assert 1558 <= 1536
```

Reproduced locally, then fixed in `a033a99` by shortening the instruction hint as described above. Not a flake and not version-specific; the earlier local full-suite run had imported `_instructions.py` before that edit landed, which is why it went out unnoticed.

## Verification on `a033a99`

- `uv run pytest -q`: 4951 passed, 6 skipped, 3 failed. All three failures are environmental in the sandbox this ran in and unrelated to the diff: `test_git_env_carries_the_commit_identity` fails because the sandbox exports `GIT_ASKPASS`, and the two `test_release_flow_contract.py` failures because that checkout has no tags. Both budget tests now pass.
- `uv run ruff check --fix .` → `uv run ruff format .` → `uv run ruff format --check .`: clean.
- `uv run mypy src/ tests/`: no issues in 246 source files.
- `bash scripts/structural_gate.sh`: 0 violations, 100%.
- `uv run pre-commit run --files <changed>`: all hooks pass, Vale included.
- `uv run python scripts/check_references.py`: exit 0, pins resolve.
- Coverage of the changed modules over the affected tests: `utils/links.py` 98%, `okf.py` 98%; the only misses are pre-existing branches.


pvliesdonk:
@claude review

pvliesdonk:
All three addressed in `40e5f2e`. Both documentation findings were correct, and they had the same cause — new prose spliced into the middle of an existing sentence rather than between two.

**1. `design.md` stray `The`** — confirmed. The original read `…(#974). The\n  tools are tagged …`; the insertion anchored on `  tools are tagged`, so the `The` that used to open that sentence was orphaned and a second one was added at the end of the new paragraph. Dropped the orphan.

**2. `obsidian-markdown.md` pin placement** — confirmed, and slightly worse than reported. The insertion did not only misplace the pin: it split `…`(` `)` `[` `]` `!` `#` literal.` from `Obsidian's own reader then cannot resolve it: …`, which is that same observation's continuation, and it left the closing `[observed: Obsidian 1.13.7 …]` marker appearing to cover a claim that is not an Obsidian observation at all. Rather than only moving the pin, the original claim is now whole again and the addition follows it, with the pin at the end of the bullet where the file's other pins sit.

**3. `convert_wikilinks_to_markdown` docstring** — taken, not deferred. It now states the encoding the way `build_index_markdown`'s does; the site had only an inline comment, and the PR body did claim docstrings were updated.

Gates on `40e5f2e`: `ruff check` / `ruff format --check` clean, `mypy src/` clean, `scripts/check_references.py` exit 0 (the moved pin still resolves), `scripts/structural_gate.sh` 0 violations, 207 tests across `test_okf.py`, `test_okf_migrate.py` and `test_links_generated_destinations.py` pass. No code behaviour changed by this commit.


---

PR #1518, opened 2026-09-17
fix(links): escape what would re-point a generated or rewritten link

### LINKED ISSUE #1513: [Bug]: a generated OKF index line is not a link when a note's path or title carries a markdown metacharacter
## What happened?

`okf_generate_index` writes each entry as `- [{title}]({path})`. When the path or the title carries a character that is structural in a markdown inline link, the line either stops being a link or points somewhere else than the note.

Noticed while closing #1494, which fixed the same producer for spaces only. Spaces are now percent-encoded; `#`, an unbalanced `(`, and a `]` in the title are not touched.

## Observed

[verified: direct calls at `dc23bf4`, 2026-09-17] `build_index_markdown` output re-read through `scanner.extract_links`:

| Input | Generated line | Links found |
| --- | --- | --- |
| Title `Hash`, path `/notes/a#b.md` | `- [Hash](/notes/a#b.md)` | one link to `notes/a`, fragment `b.md` |
| Title `Paren`, path `/notes/a(b.md` | `- [Paren](/notes/a(b.md)` | none |
| Title `Bra]cket`, path `/notes/plain.md` | `- [Bra]cket](/notes/plain.md)` | none |

The first case is the one that fails quietly: the line still renders as a link, but it names a note that does not exist and carries the rest of the file name as a fragment. The other two render as literal text.

The rename rewriter already backslash-escapes a `#` the *rename* introduces into a destination (`utils/links.py::_escape_meaning_changers`, #1353); the index builder has no equivalent, and neither escapes the link text.

## Expected

An entry the server generates for a note is a link to that note, whatever the note is called and whatever its title is, in the same way #1494 established for spaces.

## Reproduction

```python
from markdown_vault_mcp.okf import build_index_markdown

for title, path in [
    ("Hash", "/notes/a#b.md"),
    ("Paren", "/notes/a(b.md"),
    ("Bra]cket", "/notes/plain.md"),
]:
    line = build_index_markdown("I", [(title, path, None)]).splitlines()[-1]
    print(repr(line), [(link.target_path, link.fragment) for link in extract_links(line, "index.md")])

# '- [Hash](/notes/a#b.md)' [('notes/a', 'b.md')]
# '- [Paren](/notes/a(b.md)' []
# '- [Bra]cket](/notes/plain.md)' []
```

## Suspected cause

[verified: read `okf.py::build_index_markdown` and `utils/links.py` at `dc23bf4`] The builder interpolates `title` and `path` into the `[…](…)` template directly. `encode_plain_destination`, added in #1494, encodes only what CommonMark §6.3 forbids in a plain destination — the space and the ASCII control characters — which is deliberate and does not cover a `#` (legal in a destination, but read as the fragment marker), an unbalanced paren (§6.3 requires parentheses escaped or balanced), or anything in the link-text slot.

## Open questions / scope

- Scope: what `okf_generate_index` writes. #1494 covers spaces across all three producers and is closed.
- `[unverified]`: the rename rewriter's plain form may share the unbalanced-paren half of this — `docs/design/design.md` records that case as knowingly unrepaired, but I did not run it.
- `[unverified]`: `convert_wikilinks_to_markdown` builds the same template from a resolved target path and an alias; a wikilink target cannot carry a literal `#`, so the destination half probably cannot arise there, but the display-text half may. I have not verified either.
- Repairing existing prose is out of scope, as it was for #1494.
- Whether the repair is a backslash escape (what the rename path uses for `#`), a percent-escape, or the `<…>` destination form is a decision for the fix, not a claim here.

### LINKED ISSUE #1516: [Bug]: a rename into a name carrying an unbalanced ) rewrites the link onto a different note
## What happened?

`compute_new_raw_target` writes a new plain destination without escaping parentheses. When the new name carries an unbalanced `)`, the rewritten destination ends at that `)` and the link resolves to a shorter path — a different note, or a broken one — rather than to the renamed file.

`docs/design/design.md` records the parenthesis case as knowingly unrepaired, on the stated ground that such a name "is written literally in plain form and comes back *broken*, the pre-existing class". That holds for `(`; it does not hold for `)`.

## Observed

[verified: direct calls at `c49851e`, 2026-09-17] Renaming a note that a plain root-relative link `/old.md` points at:

| New name | Replacement destination | What `scanner.extract_links` then resolves |
| --- | --- | --- |
| `notes/a(b.md` | `/notes/a(b.md` | no link |
| `notes/a)b.md` | `/notes/a)b.md` | `notes/a` — **a link to a different note** |
| `notes/a(b).md` | `/notes/a(b).md` | `notes/a(b).md`, correct |
| `notes/a#b.md` | `/notes/a\#b.md` | `notes/a#b.md`, correct |

The `#` row is the escape #1353 added, for the stated reason that an unescaped `#` would leave the link "read as a fragment or an anchor, the backlink silently gone". The `)` row lands in that same class: the link still parses, still renders, and points somewhere else. Only the `(` row is the visible-breakage class the design doc describes.

## Expected

A rename leaves the link pointing at the renamed file, whatever the new name contains — or leaves it visibly broken. It does not leave a link that resolves to an unrelated note.

## Reproduction

```python
from markdown_vault_mcp.utils.links import compute_new_raw_target

raw = compute_new_raw_target(
    "markdown", "/old.md", None, "notes/a)b.md",
    source_path="s.md", old_path="old.md",
)
print(repr(raw))
# '/notes/a)b.md'
print([link.target_path for link in extract_links(f"[t]({raw})", "s.md")])
# ['notes/a']
```

## Suspected cause

[verified: read `utils/links.py::_escape_meaning_changers` at `c49851e`] The helper escapes `#` always, and `<`/`>` in the pointy form, and nothing else. CommonMark 0.31.2 §6.3 admits parentheses in a plain destination only escaped or balanced, so an unbalanced one ends the destination; the parser then reads what precedes it as the whole target. The pointy and percent-encoded forms are unaffected — `<…>` admits parentheses, and `quote` encodes them.

## Open questions / scope

- Scope: the plain-form branch of `compute_new_raw_target`, reached by `rename` and `move_folder`.
- This is the same character class as #1513, which covers what `okf_generate_index` writes. The two share `_escape_meaning_changers` and may be worth closing together.
- Whether the repair should be a backslash escape (what the `#` case uses) or something else is a decision for the fix, not a claim here.
- `[unverified]`: whether an existing vault holds such a link today. The rename path is the only producer I ran.

pvliesdonk:
## Closes / Refs

Closes #1513, closes #1516

## What & why

#1494 fixed the **invalidity** half of generated links: a space ends a plain destination, so the line stops being a link. This is the other half — a name that stays a link and *means something else*.

| Producer | Name | Was emitted | Resolved to |
| --- | --- | --- | --- |
| index builder | `/notes/a#b.md` | `- [Hash](/notes/a#b.md)` | `notes/a`, fragment `b.md` |
| index builder | `/notes/a(b.md` | `- [P](/notes/a(b.md)` | nothing |
| index builder | `/notes/a)b.md` | `- [P](/notes/a)b.md)` | **`notes/a`** |
| index builder | title `Bra]cket` | `- [Bra]cket](/notes/x.md)` | nothing |
| rename rewriter | new name `notes/a)b.md` | `/notes/a)b.md` | **`notes/a`** |

The `)` rows are the reason this is worth doing carefully. #1513 reported that case as "no link"; it is not. The destination ends at the `)`, the line still renders as a link, and it points at a different note. That is the same silent-repointing class `#` has been escaped for since #1353 — which is why `docs/design/design.md` was wrong to file parentheses under visible breakage and leave them unrepaired, and why #1516 exists.

The repairs now live behind `build_plain_destination`, which the two OKF producers call and which applies exactly what the rename path applies, so a generated destination and a rewritten one cannot disagree about what a name needs.

**Parentheses are tested on the assembled destination, not per-part.** CommonMark §6.3 balances them as one run, so `a(b.md#c)d` parses: the fragment closes what the path opened. My first cut checked the path alone, which both missed an unbalanced parenthesis living only in the fragment *and* would have broken that cross-part case by escaping the `(` on its own. Caught in self-review before the first push; pinned by `TestParenthesesBalanceAcrossTheWholeDestination`.

Escaping stays minimal. A balanced `a(b).md` is legal in the plain form and stays literal, and `_MAX_PLAIN_PAREN_DEPTH` mirrors the parser's three-level limit rather than guessing, with a parametrised test pinning the agreement so a change on either side fails loudly instead of silently emitting a link that no longer resolves.

- (make a bracketed title indexable here): **#1517**, filed and verified. `escape_link_text` makes the line valid CommonMark, which is what Obsidian, GitHub and the docs site read, but `_RE_INLINE_LINK_OPEN` is not escape-aware, so `[Bra\]cket](/x.md)` still yields no row. Such an entry renders correctly and contributes no edge to the link graph. `test_a_title_bracket_is_escaped_but_still_not_indexed_here` states that rather than hiding it, and says so in its comment: when it starts failing, #1517 has landed.
- (fix the scanner's link-text parsing): that is #1517's job. It changes which rows a note's bytes produce, so it needs an `INDEX_SEMANTICS_VERSION` bump and a far wider blast radius than this.
- (repair existing prose): unchanged from #1494. The scanner stays tolerant; only what the server emits changed.
- (`INDEX_SEMANTICS_VERSION` bump): not applicable here. `scanner.py` and `fts_index.py` are untouched; the same bytes still produce the same rows.

## One noted behaviour change

Because both plain-form repairs are properties of the destination as a whole, a rename now also repairs a **fragment the author left unparsable** — `[x](old.md#My Heading)` is rewritten to `new.md#My%20Heading` rather than re-emitting a line that is not a link. A fragment that already parses is still left exactly as found, and a fragment is never reformatted. Both halves are pinned by `TestTheRepairsReachTheAuthorsFragment`. The comment at the site says which it is; flagging it here because it is a widening of #1494's repair that neither issue asked for, and it is cheap to revert if you would rather the rewriter left a broken fragment broken.

## Design

Recorded in `docs/design/design.md`. The rewrite-spelling section is restructured around the two classes — re-pointing (escaped) and invalidity (encoded) — rather than around which character it is, since that framing is what got parentheses filed wrongly the first time. The OKF section gains the `build_plain_destination` / `escape_link_text` path and the #1517 caveat.

## Local review

- [x] Ran a local code-review pass on the cumulative diff. It surfaced two findings in my own work, both fixed before the first push: the per-part parenthesis check described above, and a comment claiming the fragment is "re-attached as found" that the assembled repair had made untrue.
- [x] No commit carries `!`. Nothing an operator sets changes. `build_plain_destination`, `escape_link_text` and `escape_unparsable_parentheses` are additive; `tests/public_import_surface.txt` is untouched. The behaviour change is confined to output that was already invalid or already pointing at the wrong note.

## Docs impact

- [ ] `README.md` — no user-facing env var, tool, or option changed.
- [x] `docs/` site pages — `docs/guides/okf.md` and `docs/tools/index.md` state the escaping for both OKF transforms.
- [x] `docs/design/` — `design.md` (both sections, one rewritten), `module-map.md`.
- [x] Inline docstrings — the three new helpers, `_escape_meaning_changers`, `build_index_markdown`, `convert_wikilinks_to_markdown`.
- [x] Ported enduring decisions: the design doc now carries *why* parentheses were reclassified, not just that they were.

## Verification on `4b2dcf1`

- `uv run pytest -q`: 4982 passed, 6 skipped, 3 failed. The three are environmental in the sandbox this ran in and unrelated to the diff: `test_git_env_carries_the_commit_identity` fails because the sandbox exports `GIT_ASKPASS`, and the two `test_release_flow_contract.py` failures because that checkout has no tags.
- Mutation-checked in two passes: reverting the escaping helpers fails 13 of the new tests; reverting the generator wiring fails 7.
- `uv run ruff check --fix .` → `uv run ruff format .` → `uv run ruff format --check .`: clean.
- `uv run mypy src/ tests/`: no issues in 246 source files.
- `bash scripts/structural_gate.sh`: 426 lines, 0 violations, 100%.
- `uv run pre-commit run --files <changed>`: all hooks pass. Vale flagged two em-dashes in the guide on the first run; rewritten.
- `uv run python scripts/check_references.py`: exit 0.
- Patch coverage over the affected tests: `utils/links.py` 99%, `okf.py` 98%; the only misses are pre-existing branches.

---
_Generated by [Claude Code](https://claude.ai/code/session_01W85921xkQnx1wR3yG4wdpF)_


pvliesdonk:
@claude review

pvliesdonk:
Confirmed and fixed in `a75a8b4`. The finding is correct, and the consequence is one step worse than the hand-trace concluded.

I ran the traced case:

```python
compute_new_raw_target("markdown", r"old.md#a\(b", r"a\(b", "new.md", "src.md", "old.md")
# before: 'new.md#a\\(b'
extract_links("[x](new.md#a\\\\(b)", "src.md")
# []
```

So it is not only that the destination is corrupted: re-parsed, it yields **no link at all**. A rename touching a link whose author had escaped a parenthesis turned a working link into a non-link — exactly the class this PR exists to close, reintroduced by the fix. Good catch, and the citation of §6.3's "escaped **or** balanced" is the right reading.

Both halves are now escape-aware through `_is_escaped`, the helper the `#` marker has used since #1353, rather than a second implementation:

| Written | Parses? | Emitted | Names |
| --- | --- | --- | --- |
| `a\(b` | yes | `a\(b` (untouched) | `a(b` |
| `a\\(b` | no | `a\\\(b` | `a\(b` |
| `a\\\(b` | yes | `a\\\(b` (untouched) | `a\(b` |
| `a\(b(c` | no | `a\(b\(c` | `a(b(c` |
| `a(b` | no | `a\(b` | `a(b` |

The even-run row is the one worth stating: two backslashes are an escaped backslash, so the parenthesis after them is bare and does count. Tests are in a new `TestAnAlreadyEscapedParenthesisIsLeftAlone`, covering the traced rename plus the odd/even/three-backslash runs and a mixed case, and asserting the repaired destination still names the same file. Mutation-checked: reverting either half to the escape-blind version fails 5 of them.

On the note about executing Python in the sandbox: the repro above is what I ran here, so that gap is closed for this finding.

Gates on `a75a8b4`: `uv run pytest -q` 4993 passed (the same 3 sandbox-environmental failures as before: `GIT_ASKPASS` exported, no tags in this checkout), ruff check/format clean, `mypy src/ tests/` clean, `scripts/structural_gate.sh` 0 violations.


---

PR #1522, opened 2026-09-17
fix(links): read a reference label with escapes honoured

### LINKED ISSUE #1519: [Bug]: a reference link whose label carries an escaped ] is not indexed
## What happened?

`[Bra\]cket][ref]` is a valid CommonMark reference link, and the scanner records no row for it. Without the escape the same link indexes normally.

#1517 fixed this for *inline* links by replacing the inline opener's negated character class with an escape-aware scan. `_RE_REF_USAGE` and `_RE_REF_DEF` still carry the plain class, so the same spelling fails in the reference family.

## Observed

[verified: direct calls at `b7de5a3` plus the #1517 branch, 2026-09-17]

```
'See [Bra\]cket][ref] here.\n\n[ref]: notes/x.md\n'   -> []
'See [Bracket][ref] here.\n\n[ref]: notes/x.md\n'     -> [('Bracket', 'notes/x.md')]
```

The control differs only in the escape, so the label's `\]` is what loses the row: the note's outlink and the target's backlink both go missing, the same outcome #1517 described for inline links.

Not observed here: the definition side. `[a\]b]: x.md` uses `_RE_REF_DEF`, which carries the same class, so it is the obvious next case; I ran only the usage side.

## Expected

A reference link whose label carries an escaped bracket is indexed like any other, with its target resolved, matching what #1517 established for inline links.

## Reproduction

```python

print(extract_links("See [Bra\\]cket][ref] here.\n\n[ref]: notes/x.md\n", "s.md"))
# []
```

## Suspected cause

[verified: read `scanner.py` at the #1517 branch] `_RE_REF_USAGE` is `\[([^\]]*)\]\[([^\]]*)\]` and `_RE_REF_DEF` is `^\s*\[([^\]]+)\]:\s*(.+)$`. Neither class can see a backslash, so the first `]` closes the label whether or not it is escaped — the defect #1517 fixed for the inline opener.

## Open questions / scope

- Scope: `_RE_REF_USAGE` and `_RE_REF_DEF`, and whether a usage and its definition have to spell the label the same way to match.
- #1517 deliberately left these: the reference usage is a different match shape (two adjacent bracket spans, not one followed by `(`), so the inline scan does not drop in, and folding it in would have added an independent semantic delta to that bump.
- Any fix changes which rows a note's bytes produce, so per `AGENTS.md` it needs an `INDEX_SEMANTICS_VERSION` bump unless the current value already covers the next release.
- `[unverified]`: whether a naive escape-aware class is affordable here. It is not for the inline opener — #1517 measured seconds on one pathological paragraph and used a scan instead — and `_RE_REF_USAGE` measured worse in the same shape, but I did not measure a fix.
- Not in scope: bracket balancing in link text, a departure recorded in `docs/design/design.md`.

pvliesdonk:
## Closes / Refs

Closes #1519. Refs #1517 (the inline half, merged as #1520), #1343 (improved again, measured below), #1521 (filed while reading the rewrite side; not touched here).

## What & why

#1517 taught the *inline* link text CommonMark's escape rule and deliberately left the reference family. So `[Bra\]cket][ref]` and `[r\]ef]: x.md` — both valid CommonMark — still produced **no row at all**: no outlink on the note, no backlink on the target, the same hole one bracket to the left.

Both sides now read a label through `_find_bracket_span`, the primitive `_find_inline_link_open` was refactored onto. It returns the next `[…]` with escapes honoured; each caller adds its own shape test:

| Caller | Shape test |
| --- | --- |
| `_find_inline_link_open` | a `(` after the span |
| `_find_reference_usage` | a second span immediately adjacent |
| `_iter_reference_definitions` | a `:`, then the same multi-line tail the old pattern used |

Verified against `markdown-it-py` as an oracle (the reader this repo's CommonMark reference already cites):

| Input | Before | After | markdown-it |
| --- | --- | --- | --- |
| `[Bra\]cket][ref]` | no link | `notes/x.md` | `notes/x.md` |
| `[Bracket][r\]ef]` + `[r\]ef]: …` | no link | `notes/x.md` | `notes/x.md` |
| `[a\]b][r\]ef]` + `[r\]ef]: …` | no link | `notes/x.md` | `notes/x.md` |
| `[Bra\]cket][]` (collapsed) | no link | `notes/x.md` | `notes/x.md` |
| `[Bracket][ref]` | `notes/x.md` | `notes/x.md` | `notes/x.md` |

**The definition side is not separable from the usage side.** A label ends at its first *unescaped* `]`, so a label holding one can be written no other way than `\]` — a usage that reads the escape and a definition that does not can never meet.

## Why a scan and not a wider character class

#1519 left this `[unverified]`: whether an escape-aware class is affordable here, given #1517 found it was not for the inline opener. Measured, and it is not:

| Input | old class | escape-aware class | this scan |
| --- | --- | --- | --- |
| usage, 19 KB of `[a\]b` | 0.17 ms | **1884 ms** | **2.36 ms** |
| definitions, 23 KB of `[a\]b\n` | 0.36 ms | **2479 ms** | **2.44 ms** |

Both class figures quadruple per doubling of the input where the scan doubles.

The definition scan also removes an *existing* quadratic that has nothing to do with escapes, because the old pattern's `[^\]]+` rescanned to the end of the body from every line-opening `[`:

| Input | before | after |
| --- | --- | --- |
| `extract_links` on 4000 × `[a\n` | 395 ms | **9 ms** |
| `extract_links` on `[` × 8000 (#1343's shape) | 1811 ms | 1477 ms |

The price, stated plainly: **+6%** across this repository's own 2.8 MiB of markdown (247 ms → 262 ms, the same 381 links), and roughly **2x** the old class on a bracket-dense body that yields no links at all (4.4 ms → 10.4 ms on 5000 × `[a\]b`). Linear either way.

## `INDEX_SEMANTICS_VERSION` stays 10

Checked rather than assumed: `git tag --contains f9f63687` is empty against all 86 fetched tags, so the value #1520 set has not shipped and already covers the next release. Per `AGENTS.md` ("once per release", #1365) the note is **extended**, not bumped to 11 — a deployed vault rebuilding to 10 gets both changes at once.

Its safety claim is that **nothing else moved**, pinned as a differential property rather than by example. `TestTheScansAgreeWithThePatternsTheyReplaced` asserts that each scan and the pattern it replaced produce the same result on *every* string up to length 9 (usages, over `[`/`]`/filler) and length 6 (definitions, over `[`/`]`/`:`/newline/space/filler) — exhaustive, not sampled, and comparing the whole match *sequence* so the scans' resume points are pinned against `finditer`'s non-overlapping ones.

- (the **shortcut** form `[label]`): still not extracted, a standing gap rather than anything new. It is why `[a\][ref]` loses a row: its label runs past the escaped `]`, so it is not a full reference link, and the link a reader still finds there is the leftover shortcut. markdown-it renders that input as `<p>See [a]<a href="notes/x.md">ref</a> here.</p>`, which is the evidence — the link's text is `ref`, not the label. Tested and stated in both the design doc and the CommonMark reference; no issue filed, because it is the same standing gap those documents already record.
- (#1521): a rename does not rewrite an inline link whose text carries an escaped `]`, and reports it as updated anyway. Found while reading `apply_link_replacement` for this change, verified end to end, filed rather than folded in — it is the write side, and the same "is an escape-aware class affordable here" question needs its own measurement.
- (#1343, **not** closed here): the unmatched-bracket run is *improved* again (1811 ms → 1477 ms on `[`×8000) but its dominant term remains `_RE_WIKILINK`, untouched here. That issue was closed by accident when #1520 merged — its body carried a bullet whose text began with a closing keyword and the number, which GitHub acted on — and is reopened with the explanation. This bullet is deliberately worded so the same thing cannot happen twice.
- (bracket balancing): `[Note [draft]](x.md)` is still not a link here. Needs counting, not escape-awareness; unchanged departure.

## Design

`docs/design/design.md` gains the change, the measured reason for a scan, the not-separable argument, and the shortcut gap. `docs/design/reference/commonmark-gfm.md` pins the label rules this now implements and re-points its "depended on by" entry at the two scans; `obsidian-markdown.md`'s footnote pointer likewise no longer names deleted regexes.

## Local review

- [x] Ran a local code-review pass on the cumulative diff. Three findings in my own work, all fixed in `7c2c26a0` before this PR: a claim that the two halves were "one release apart" when they are in the same unreleased range; a CommonMark-reference line calling the 999-character label cap and the non-whitespace floor departures, when a check showed markdown-it does not enforce them either, so they are not; and an explanation of `[a\][ref]` asserted rather than observed, now carrying markdown-it's actual rendered output. A fourth, found after opening: this body's own #1343 bullet had reproduced the accidental-closing-keyword mistake it describes.
- [x] No commit carries `!`. Nothing an operator sets changes and `tests/public_import_surface.txt` is untouched; the index rebuild is automatic and already covered by version 10.

## Docs impact

- [ ] `docs/` site pages — link extraction is internal; no tool, env var or option changed. (`docs/guides/okf.md`'s statement that generated link text is bracket-escaped was already accurate and is now also true of the graph.)
- [x] `docs/design/` — `design.md`, `module-map.md`, and both reference pages.
- [x] Inline docstrings — `_find_bracket_span`, `_find_reference_usage`, `_iter_reference_definitions`, the two new patterns, and the `INDEX_SEMANTICS_VERSION` note.

## Verification on `7c2c26a0`

- `uv run pytest -q`: 5072 passed, 4 skipped, 1 failed. The failure is environmental in this sandbox and unrelated: `test_release_flow_contract.py` needs a `knope/port/*` branch this checkout lacks. (`test_git_env_carries_the_commit_identity` was deselected for the same class of reason — the sandbox exports `GIT_ASKPASS`.)
- Patch coverage: `diff-cover` reports **100%** over the 44 changed executable lines.
- `uv run ruff check --fix .` → `uv run ruff format .` → `--check`: clean. `uv run mypy src/ tests/`: clean, 248 files.
- `bash scripts/structural_gate.sh`: 417 lines, 0 violations, 100%.
- `uv run pre-commit run --files <changed>`: all hooks pass. `uv run python scripts/check_references.py`: exit 0.

### One note on this branch's history

The branch carries three extra commits (`1e244647`, `2e15714a`, `222c6211`) and an `ours` merge. They are #1520's: its squash merge left them in `main` by content but not by ancestry, and restarting the branch from `main` would normally need a force-push, which this environment blocks. Recording the old tip as merged achieves the same thing without rewriting anything: the merge changes no file, and the PR's merge base is `main`'s head, so the diff above is #1519 alone.


pvliesdonk:
@claude review

PR #1524, opened 2026-09-17
perf(links): match wikilinks with a linear scan

### LINKED ISSUE #1343: [Decay]: link regexes backtrack quadratically on an unmatched-bracket run
### What

The three link regexes in `scanner.py` backtrack quadratically on a long run
of unmatched `[` characters. A single such file stalls indexing for the whole
vault, because `scan_directory` / `build_index` runs single-threaded over one
write queue.

Measured on this machine, matching over a body of literal `[` with no closing
bracket anywhere:

```
n=16000   _RE_INLINE_LINK  0.54s
          _RE_REF_USAGE    0.53s
          _RE_WIKILINK     3.50s
n=40000   whole extract   ~29s
```

Doubling `n` roughly quadruples the time. `_RE_WIKILINK` is the dominant
term by ~6x.

### Where

`src/markdown_vault_mcp/scanner.py`: `_RE_INLINE_LINK`, `_RE_REF_USAGE`,
`_RE_WIKILINK`.

#1339 reduced the exposure for the first two by matching one paragraph at a
time rather than over the whole document, which shrinks the backtracking
window to a paragraph — on realistic multi-paragraph markdown that measured
~90x faster than the whole-document form. It does nothing for a document with
no blank lines in it, which is the shape above, and `_RE_WIKILINK` still runs
per paragraph with the same quadratic character.

### Why it compounds

The input is not hypothetical: a botched bracketed-citation export, or pasted
log/JSON that was never fenced, produces exactly this shape. Vaults of
converted PDFs are where #1334 was found in the first place.

It compounds because the cost is paid on every indexing pass over that file,
and because the single-owner writer thread means one file's stall is the
whole vault's stall — a cold build or a reindex simply does not finish. The
larger the vault, the more likely one such file exists, and nothing surfaces
which file is responsible.

### Suggested direction

`[unverified]` A guard on the scanned unit rather than a cleverer regex looks
more promising than tuning the patterns: bounding what is handed to the
matcher (or refusing to link-scan a body whose bracket density is
pathological, and recording a skip the way #1129's tombstones do) keeps the
cost bounded regardless of pattern. An atomic group around the text span was
measured to recover roughly a third of the cost in one variant, which is a
mitigation rather than a fix. I have not verified which approach holds up.

Surfaced by an automated review on
[#1339](https://github.com/pvliesdonk/markdown-vault-mcp/pull/1339), which
measured the pre-existing baseline while reviewing a correctness change to
the same regexes.

---

pvliesdonk:
## Closes / Refs

Closes #1343. Refs #1522 (carries one non-blocking note from its review), #1384 (the departure this preserves).

## What & why

`\[\[([^\]|\n]+)(?:\|([^\]]+))?\]\]` rescans forward from every `[[`, so a run of `[` that never closes costs O(n²). The issue measured `extract_links` at about 29 s on 40000 of them. The shape is not hypothetical — a bracketed-citation export gone wrong, or pasted log or JSON that was never fenced — and one such file stalls the **whole vault**, because indexing runs single-threaded over one write queue and nothing surfaces which file is responsible.

#1339's region walk bounds the other patterns to one paragraph, which is why the two escape-aware scans (#1517, #1519) are cheap on ordinary notes. It does nothing for a document with no blank lines in it, and that is exactly this shape.

`_find_wikilink` replaces the pattern with a scan:

| Input | before | after |
| --- | --- | --- |
| `extract_links` on `[` × 16000 | 6026 ms | **16 ms** |
| `extract_links` on `[` × 40000 | ~29 s (issue), 36 s for the pattern alone | **40 ms** |
| `_find_wikilink` on `[[a\|` × 32000 | 515 ms† | **0.03 ms** |
| `_find_wikilink` on `[[a\|b` × 32000 + `]` | ~510 ms† | **0.04 ms** |

† Against the *first* version of this PR, not against the old pattern — see below.

## Why it is linear, not just faster

The retries all ask the same question. A target ends at the first `]`, `|` or line ending after the `[[`, and **that stop is the same character whichever `[[` of a run opened it**; everything the match then needs is read from the stop onwards. So a failure at the stop is a failure for every opener before it, and the search resumes past the stop rather than at the next `[`. The stop is never itself a `[`, so resuming there skips no opener that could have matched.

**That argument binds every forward search, and the first version of this PR applied it to only one of them.** A pipe also ends a target, and the alias search that follows it resumed at the pipe rather than past its own reading, so every opener re-ran it and `[[a|` × n was quadratic — the very shape this PR exists to remove, moved one character right. Caught by the `@claude` review on this PR (`35202756` fixes it), which found it by reading the argument rather than by running the benchmark, because the benchmark only ever used bare `[`.

A second quadratic route came out of investigating the first and was not in the report: `[[a|b` × n closed by a single trailing `]`, where the alias search *succeeds* and the `]]` test fails, so the resume point matters even on a hit. A fix addressing only the unterminated case would have left it.

Both alias paths now resume past the `]` the alias search read, licensed by the same shared-stop reasoning: that `]` is the first after the pipe, so it is the same character for every opener behind it and they all fail with it. Where no `]` remains at all, the scan returns outright, as it already did for a missing target stop. All three shapes now grow **2.0x per doubling**.

## Two alternatives, measured and rejected

**Possessive quantifiers** (`[^\]|\n]++`, Python 3.11+). Provably equivalent here — giving a character back can never help, because the given-back character is by construction not the `|` or `]` the pattern needs next — and they do remove the inner backtracking, 9x on this input. They leave the shape quadratic, because the retry from every `[[` is the real cost:

| n | old | possessive | this scan |
| --- | --- | --- | --- |
| 4000 | 366 ms | 40 ms | 0.025 ms |
| 8000 | 1471 ms | 165 ms | 0.039 ms |
| 16000 | 5868 ms | 642 ms | 0.076 ms |

**A guard on the scanned unit**, the issue's own suggested direction: refusing to link-scan a bracket-dense body, recording a skip the way #1129's tombstones do. Not pursued — it drops real rows to bound a cost the scan simply removes.

## No `INDEX_SEMANTICS_VERSION` note

The scan changes no row for any input, which is a claim only a property can carry, so it is pinned the same way the escape-aware scans were: `TestTheScanAgreesWithThePatternItReplaced` compares the whole match sequence against the superseded pattern on *every* string up to length 8 over the five characters that pattern can distinguish, and development additionally checked 400k random strings over a wider alphabet. Both were re-run after the resume point moved in `35202756`, since that is exactly the kind of change that could have shifted a match. The departures the pattern carried are carried too, and tested: a single `]` still ends the target where Obsidian needs `]]` (#1384), an embed is still indexed as a link, and the leftmost `[[` still wins, so `[[[a]]` links `` [a ``.

**What that property could not do is catch the defect above**, and the distinction is worth recording: it is exhaustive on *correctness* and silent on *complexity*. The linearity tests are what should have caught it, and they measured only the shape I had in mind. All three shapes are pinned now — by a growth ratio, and by a wall clock chosen to be one the defect actually crosses (the quadratic version cost about 0.8 s at n=40000, which the original 5 s bound would have waved through; that case now runs at 100000 against 1 s). Verified in both directions: reverting the fix fails all four new cases, applying it passes them.

- (#1384): `[[Sq]uare]]` is a link for Obsidian and still is not one here. The scan preserves the old stop-at-first-`]` behaviour deliberately; changing it would alter rows and need its own semantics bump.
- (#1521): the rename-side escaped-bracket defect is untouched — different file, different question.
- (the reference and inline scans): already linear as of #1519; measured again here at 5.2 ms on 16000 `[`, against `_RE_WIKILINK`'s 5849 ms, which is why this PR is only about the wikilink.
- No guard, tombstone, or skip record is introduced. If a pathological file still costs too much after this, that is a new observation, not this fix.

## Design

`docs/design/design.md` gains the change with the linearity argument, both rejected alternatives, and the "binds every forward search" correction; `module-map.md` and `docs/design/reference/obsidian-markdown.md` re-point their `_RE_WIKILINK` references at `_find_wikilink`.

## Local review

- [x] Ran a local code-review pass on the cumulative diff. One finding, in my own test rather than the code: I had asserted `link_text == "four"` for `[[four#Heading]]` where it is `four#Heading`, because the text is read before the `#` split. The code was right and pre-existing; the expectation was wrong, and is now spelled out with a comment. The pass did **not** find the alias-branch complexity gap — the review did.
- [x] Also carries `4d0136fa`, the one actionable non-blocking note from #1522's review: a comment saying why `_iter_reference_definitions` returns rather than continues. Carried here rather than pushed to #1522, whose review and CI were green and for which the finding was explicitly not blocking.
- [x] No commit carries `!`. Nothing an operator sets changes, `tests/public_import_surface.txt` is untouched, and no stored row changes.

## Docs impact

- [ ] `docs/` site pages — link extraction is internal; no tool, env var or option changed.
- [x] `docs/design/` — `design.md`, `module-map.md`, `reference/obsidian-markdown.md`.
- [x] Inline docstrings — `_find_wikilink` (including the corrected invariant), the three patterns it replaced one with, and the note in `_iter_reference_definitions`.

## A note on the subject line

`perf:` cuts no release and never reaches `CHANGELOG.md` (`AGENTS.md`, "Conventional commits"). That is the right classification here: nothing a user can observe changes except how long a pathological file takes. If you would rather it ship on its own, the honest alternative is `fix:` — it does fix a performance defect with a filed bug behind it — but I did not assume that. Retitling is enough to change it; the `PR Title` job reads the live title, so it needs no push.

## Verification on `3c168824`

- Rebuilt on the current `main`, which carries Renovate's fastmcp 4.0.5 (#1523) and openai 3.15.0 (#1525); the environment was re-synced with `uv sync --all-extras --all-groups` and everything below re-run after each.
- `uv run pytest -q`: 5104 passed, 20 skipped, 1 failed. The failure is environmental in this sandbox and unrelated: `test_release_flow_contract.py` needs a `knope/port/*` branch this checkout lacks. (`test_git_env_carries_the_commit_identity` is deselected for the same class of reason — the sandbox exports `GIT_ASKPASS`.)
- Patch coverage: `diff-cover` reports **100%** over the 34 changed executable lines; Codecov agrees.
- `uv run ruff check .` / `ruff format --check .`: clean. `uv run mypy src/ tests/`: clean, 249 files.
- `bash scripts/structural_gate.sh`: 335 lines, 0 violations, 100%.
- `uv run python scripts/check_references.py`: exit 0.


pvliesdonk:
@claude review

pvliesdonk:
Confirmed and fixed in `35202756`. The finding was correct and the claim it challenged was the PR's core one, so thank you for reading the argument rather than the benchmark.

**Reproduced first.** On `4d0136fa`, `[[a|` × n against `_find_wikilink`:

| n | before fix |
| --- | --- |
| 4000 | 10.7 ms |
| 8000 | 33.7 ms (3.1x) |
| 16000 | 130.5 ms (3.9x) |
| 32000 | 514.9 ms (3.9x) |

**A second shape, not in the report.** `[[a|b` × n closed by a single trailing `]` is quadratic by a different route: there the alias search *succeeds* and the `]]` test fails, so the resume point matters even on a hit. 175 ms at n=16000, same 4x signature. A fix that only handled the unterminated case would have left this one.

**The fix.** Both alias paths now resume past what the alias search read, and the same shared-stop argument licenses it: the first `]` after the pipe is the same character for every opener behind it, so they all fail with it and resuming past it skips no possible match. Where no `]` remains at all the scan returns outright, as it already did for a missing target stop. That is the suggested direction, arrived at independently for the second shape as well.

| shape | before | after |
| --- | --- | --- |
| `[[a|` × 32000 | 515 ms | **0.03 ms** |
| `[[a|b` × 32000 + `]` | ~510 ms | **0.04 ms** |
| `[` × 32000 | 0.06 ms | 0.06 ms |

All three now grow 2.0x per doubling.

**Equivalence re-verified**, since the resume point moved: identical to the superseded pattern on every string up to length 8 over the five characters it distinguishes, and on 400k random strings over a wider alphabet.

**Tests.** You were right that `TestTheScanIsLinear` only exercised bare `[`. The ratio test is now parametrised over all three shapes. One further thing your report did not reach: the *wall-clock* test would not have caught this either even with the right input, because the quadratic version costs about 0.8 s at n=40000 and the bound was 5 s. That case now runs at 100000 against 1 s, where the defect costs seconds and the fix costs under a millisecond. Reverting the fix fails all four new cases; applying it passes them — checked in both directions rather than assumed.

**Docs corrected.** The docstring and `docs/design/design.md` both claimed "each character is looked at a bounded number of times" without qualification. Both now state that the argument binds every forward search, name this as the trap the first version fell into, and record the pipe shape as the realistic one — an unfenced markdown table pasted into a note is a wall of `|`.

Verification on `35202756`: 5104 passed (only the sandbox's known `knope/port/*` case failing), 100% patch coverage over the 34 changed lines, structural gate 335 lines / 0 violations, ruff + mypy clean.

---

pvliesdonk:
`Test (Python 3.13)` failed on `3c168824`, and it was my own new test, not the code. Fixed in `03a80462`.

```
FAILED tests/test_links_wikilink_scan.py::TestTheScanIsLinear::
  test_doubling_the_input_does_not_quadruple_the_work[unterminated-alias]
AssertionError: (3.0313999985764895e-05, 0.00010893099999975675)
assert 0.00010893099999975675 < (3.0313999985764895e-05 * 3)
```

30 µs at n=20000 against 109 µs at n=40000 — a 3.6x reading on a commit whose growth is provably 2x.

**The instrument was wrong, not the bound.** Telling linear (2x per doubling) from quadratic (4x) needs roughly ±40% accuracy, and the linear scan over 20000 openers takes about 50 µs. That is too small a quantity to measure that closely on a shared runner. Measured locally over seven repetitions, the same ratio spans 1.82–2.26 at that size; CI's spread is evidently wider. Loosening the bound does not rescue it either — anything approaching 4x stops detecting the defect at all. So the test reported on the runner rather than on the code, which is worse than not having it.

`codecov/patch` erroring with `diff-cover step did not run` is downstream of the same failure: the test job exits before the coverage step.

**What replaces it.** The wall-clock tests carry the coverage alone now, and they can: the defect's signature is four orders of magnitude (0.03 ms against 515 ms at n=32000), so a bound placed between them is crossed by the defect and nowhere near the truth on a slow day. Re-verified in both directions after the change — reverting the alias fix still fails both alias cases, at n=100000 against a 1 s bound.

**The honest cost**, recorded in the test file rather than left implicit: a future regression milder than the wall-clock bound would now slip through, where a ratio test that actually worked would have caught it. I could not make one work at a size where the linear path is microseconds, and a flaky gate is worse than a coarse one. If you would rather have the ratio test back with a much larger `n` (the base cost reaches ~1.1 ms at n=400000, where the local spread tightens to 1.97–2.01), say so and I will add it at that size instead.

Verification on `03a80462`: 5101 passed locally (only the sandbox's known `knope/port/*` case), structural gate 313 lines / 0 violations, ruff and mypy clean. Also picked up `main` (`3c168824`) after the openai 3.15.0 bump in #1525 left the branch behind.

---

pvliesdonk:
@claude review

PR #1529, opened 2026-09-18
fix(scanner): close a link at the nearest unmatched bracket

### LINKED ISSUE #1526: [Bug]: an image whose alt text opens a bracket hides a real link from the index
## What happened?

`[a[b](x.md)` is literal text `[a` followed by a link to `x.md`, and the scanner stores no row for it. The `!` lookbehind reads the whole construct as an image.

## Observed

[verified: `extract_links` against `markdown-it-py` 3.0.0 `commonmark`, at `5cc2284`, 2026-09-18]

| Input | `extract_links` | markdown-it renders |
| --- | --- | --- |
| `[a[b](x.md)` | no rows | `[a<a href="x.md">b</a>` |
| `[[](x.md)` | no rows | `[<a href="x.md"></a>` |
| `[alt](x.md)` | no rows | ``'&lt;img src="x.md" alt="alt" /&gt;'`` |
| `[[](x.md)` | one row, `raw_target` `x.md` | `[<a href="x.md"></a>` |

The first two rows are links a reader follows and the graph does not hold: no outlink on the note, no backlink on the target. The third is a genuine image and correctly skipped; the fourth shows the same shape without the `!` is indexed.

`docs/design/reference/commonmark-gfm.md` claimed "Right: images by `!` lookbehind" without qualification. That line is corrected in the PR for #1521; this issue is the behaviour it now describes.

## Expected

A link that a CommonMark reader finds is indexed, or the departure is a deliberate one with a recorded reason. `[` followed by `[` is literal text plus a link, not an image.

## Reproduction

```python

print(extract_links("[a[b](x.md)", "s.md"))   # []
print(extract_links("[[](x.md)", "s.md"))      # one markdown link to x.md
```

## Suspected cause

[verified: read `_extract_inline_links` at `5cc2284`] The image test is `region[open_index - 1] == "!"`, where `open_index` is the **first** `[` of the span. For `[a[b](x.md)` that first `[` is the one right after the `!`, so the construct is dropped whole. CommonMark instead closes the link at the innermost unmatched `[`, leaving `[a` as text.

## Open questions / scope

- Scope: the image test in `_extract_inline_links`. Any change alters which rows a note's bytes produce, so per `AGENTS.md` it needs an `INDEX_SEMANTICS_VERSION` bump unless the current value already covers the next release.
- `[unverified]`: whether the correct rule is reachable without bracket balancing. CommonMark's is "the innermost unmatched `[` opens the link", which is the balancing question already recorded as a departure (`[a [b] c](x)`), so the two may have to be settled together. I have not tried.
- Surfaced while fixing #1521: the rewrite side used to replace these by accident, because a failed `(?<!!)` made the regex engine retry from the inner `[`. That retry is quadratic (26 s on 16000 `[`), so the rewrite now matches the index instead. Whichever way this issue is decided, both sides move together.
- Not in scope: bracket balancing in ordinary link text, and the `!` handling for wikilink embeds (`[[pic.png]]`), which is decided separately by the target's extension (#1333).

pvliesdonk:
## Closes / Refs

Closes #1526
Refs #1528

## What & why

Take a line holding an image and then a real link: `[alt](a.png)` with a leading
exclamation mark, a space, `and`, then `[real](b.md)`. It indexed no link for
`b.md` at all. (The literal is spelled out because this PR body is written
through an API that strips the image marker, even inside a code fence;
`tests/test_links_commonmark_openers.py` has it as written.)

The inline-link opener took the **outermost** `[` in a region, so the image's
alt text swallowed the opener and every real link after it on the same line was
lost. No outlink on the note, no backlink on the target. CommonMark 0.31.2
Ex. 575 is explicit that an image description may contain a link.

The opener is replaced by `iter_inline_links` in `utils/links.py`: one
left-to-right scan holding a delimiter stack, which is what CommonMark 0.31.2
Section 6.3 actually specifies. A `]` closes the **nearest unmatched** `[`, an
image opener yields nothing, and "links may not contain links" (Ex. 518)
deactivates every opener still on the stack the moment one closes. The
destination grammar (`_parse_destination` and its cluster) moves out of
`scanner.py` into `utils/links.py` alongside it, so the rewrite path in
`managers/document.py` and the index path in `scanner.py` now read the same
inline links **by construction** rather than by two patterns kept in step by
hand. That is the invariant #1521 had to assert; it is now structural.

### Scope flag

**#1526 is not separable from the bracket-balancing departure. They are the same
defect.** Both come from taking the outermost `[` where CommonMark takes the
nearest unmatched one. Fixing it correctly also changes stored `link_text` for
nested links (`[a[b](x)` becomes `b`, not `a[b]`) and adds rows for
`[a [b] c](x)`. Those are behaviour changes beyond the issue's literal text, and
they are the price of a correct fix rather than a patch over one symptom.

Three existing test modules pinned the old departures as intended behaviour and
are flipped here, each with its reasoning rewritten rather than the assertion
merely inverted.

### Evidence

Validated against `markdown-it-py` (its `commonmark` preset) as an out-of-band
oracle, comparing which inline links each side finds:

- **0 disagreements** over 194,801 generated inputs reaching divergent shapes.
  The superseded opener has **200** on the same corpus.
- Linearity confirmed at 2.0 to 2.1x per doubling on bare `[`, escaped, nested
  and real-link shapes, at roughly 2x the old opener's constant while also
  parsing destinations. It replaces a pattern, not a scan, so this is not a
  regression against #1343's work.

`INDEX_SEMANTICS_VERSION` stays at **10**: `git tag --contains` on the commit
that set it is empty across 86 tags, so the value has never shipped and covers
this change too (per the once-per-release rule in AGENTS.md, #1365). Its note
gains a third entry. The note says plainly that this change does **not** claim
"nothing else moved", because it does move things.

### A defect this branch introduced and fixed (05f0fd80)

Worth recording rather than burying in a commit. The first draft had
`_replace_inline_destinations` re-derive each destination's position by
searching the link for the literal string, taking the last occurrence before
the closing `)`. That is the wrong occurrence whenever the text after the
destination repeats it, which a title routinely does: `[a](old.md "old.md")`
had its **title** rewritten and its destination left naming the moved file, and
a title merely containing the destination (`"the old.md note"`) was mangled the
same way. A broken link, reported as a successful rewrite, because the length
delta came out exactly right.

Caught in review. The scan on `main` anchored at a fixed offset and never had
the ambiguity, so this was a regression of this branch's own making.
`iter_inline_links` now yields an `InlineLink` named tuple carrying
`target_start`, and the splice uses it; no content search remains in the
rewrite path.

The property that should have caught it is strengthened rather than left
standing. `test_links_rewrite_agrees_with_index.py` compared only
`len(after) - len(content)`, which cannot tell a destination from a same-length
substring elsewhere in the link; and its generated alphabets hold no quote, so
**no input they produce has a title at all** — the shape was unreachable, not
merely unsampled. It now re-indexes the result and compares destinations, and
gains an exhaustive pass over the title spellings Section 6.3 allows.
`test_links_commonmark_openers.py` pins the offset invariant directly. Verified
both ways: with the old splice restored, three fixtures and the new property
fail; with the fix, all pass.

- **The reference family still opens at the outermost bracket**: #1528. Its match
  shape is two adjacent bracket spans rather than one followed by `(`, so the
  inline iterator does not drop in. Folding it in would put a second semantic
  delta into one `INDEX_SEMANTICS_VERSION` bump, and the two want separate
  before/after evidence.
- **The shortcut reference form `[ref]` is still not extracted.** A standing gap,
  unchanged here, restated in the tests so the reference cases are read as
  instances of it rather than as something this PR introduced.

## Design

`docs/design/design.md` gains a section, "**A `]` closes the nearest unmatched
`[` (#1526)**", with a before / now / CommonMark table over the twelve shapes
that distinguish the two rules. Nine of the twelve differ from markdown-it under
the old rule.

Two claims in that section were verified empirically before shipping rather than
reasoned to: that reference links still take the outermost `[` (`a[b` where
CommonMark reads `b`), and that `[a]()` stores no row.

`docs/design/reference/commonmark-gfm.md` gains three claim blocks (Ex. 518,
Ex. 575, and the nearest-unmatched rule) with pins onto the tests that hold
them.

## Local review

- [x] No commit here carries `!`. The MCP tool surface is unchanged; stored index
      rows change, but they are rebuilt from vault bytes on the version bump with
      no operator action, and no importable name changes signature.

## Docs impact

- [x] `docs/design/` (the new section, `module-map.md`, and the reference page)
- [x] Inline docstrings (`iter_inline_links`, `InlineLink`, and
      `_extract_inline_links` rewritten to point at where the grammar now lives)
- [x] Enduring decisions ported: the delimiter-stack rule and why the reference
      family is out of scope both live in `docs/design/design.md`, not only in
      this PR body.

## Gates

Local, on `05f0fd80`:

- `uv run pytest`: **5154 passed, 0 failed.** Two deselected, both environmental
  in this clone and unrelated to the diff: the `knope/port/*` release-flow case
  and the `GIT_ASKPASS` one the sandbox exports.
- `mypy src/ tests/`: clean, 251 files.
- `scripts/structural_gate.sh`: 662 lines, 0 violations, 100%.
- `diff-cover`: 88 lines, 0 missing, **100%** patch coverage.
- `scripts/check_references.py`: exit 0.

CI on `05f0fd80` is green across the board, and Codecov reports all modified
lines covered with project coverage up 0.03 points. The `scanner.py` file-level
delta reads `-0.23%` because the file shrank, not because new code is
uncovered; its patch coverage is 100%.


pvliesdonk:
@claude review

pvliesdonk:
Finding 1 is real, I reproduced it, and it is worse than the narrowest reading. Both nits fixed too. Pushed as 05f0fd80.

### 1. The splice offset (High) — confirmed and fixed

Reproduced before touching anything. The reviewer could not execute Python; the trace was right anyway:

| input | old output |
|---|---|
| `[a](old.md "old.md")` | `[a](old.md "new.md")` |
| `[a](old.md "the old.md note")` | `[a](old.md "the new.md note")` |

So it is not only a title that *equals* the destination: any title *containing* it is mangled, and in both cases the destination stays stale. A broken link, reported as a successful rewrite, because the length delta comes out exactly right.

Also confirmed the regression is mine and not pre-existing. `origin/main` anchors with `tail.match(content, after_open)` at a fixed offset; this branch replaced that with `content.rindex(old_raw, 0, close)`. The diagnosis in the review is precisely correct.

Fixed the way the review suggests: `_parse_destination` now returns the destination's start offset, `iter_inline_links` yields an `InlineLink` named tuple carrying `target_start`, and the splice uses it. No content search remains in the rewrite path. It is a named tuple rather than a fifth positional element because `target_start` and `open_index` are easy to confuse, and confusing them corrupts a file quietly.

### The property that let it through

This is the part worth more than the fix. The review is right that `len(after) - len(content) == stored_count` cannot distinguish "replaced the destination" from "replaced a same-length substring elsewhere". Two changes:

1. `_rewritten` now re-indexes the result and compares the destinations themselves, not just how many bytes moved.
2. The generated alphabets (`[]()!\` and `[]()!\ax.`) contain no quote, so **no input they generate has a title at all** — the shape was unreachable, not merely unsampled. Added `test_titles_agree_too`, exhaustive over the title spellings §6.3 allows, each built to repeat the destination.

Verified in both directions rather than assumed. With the old splice restored:

```
FAILED TestATitleThatRepeatsTheDestination::test_a_title_naming_the_same_file_is_left_alone
FAILED TestATitleThatRepeatsTheDestination::test_a_title_merely_containing_it_is_left_alone_too
FAILED TestATitleThatRepeatsTheDestination::test_the_rename_end_to_end_keeps_the_title
FAILED TestTheRewriteSeesWhatTheIndexSees::test_titles_agree_too
4 failed, 25 passed
```

With the fix, all pass. `test_links_commonmark_openers.py` also gains the offset invariant directly (`region[target_start : target_start + len(raw_target)] == raw_target`), exhaustively over short strings, so the anchor cannot drift again without a test noticing.

### 2. Stale `find_inline_link_open` references (Low) — fixed

All three: the `_replace_inline_destinations` docstring, `docs/design/module-map.md:23`, and the pins entry at `docs/design/reference/commonmark-gfm.md:380` (which named it as current implementation evidence, so it mattered more than the other two).

The `docs/design/design.md` mentions stay as written: they are past-tense narrative of what #1517 and #1519 did at the time, in those issues' own sections, which is how that document records history.

### Gates

5154 passed, 0 failed (two deselected: the `knope/port/*` and `GIT_ASKPASS` cases, both environmental in this clone and unrelated). ruff, mypy, pre-commit clean. Structural gate 662 lines, 0 violations. `diff-cover` 88 lines, 0 missing.

---

pvliesdonk:
@claude review

PR #1530, opened 2026-09-18
fix(scanner): open a reference link at the nearest unmatched bracket

### LINKED ISSUE #1528: [Bug]: a reference link opens at the outermost bracket, not the nearest unmatched one
## What happened?

`See [a[b][r] here.` with `[r]: x.md` stores a reference link whose text is `a[b`. CommonMark reads the text as `b`: a `]` closes the **nearest unmatched** `[`, so `[a` stays literal.

## Observed

[verified: `extract_links` against `markdown-it-py` 3.0.0 `commonmark`, on the #1526 branch, 2026-09-18]

```
'See [a[b][r] here.\n\n[r]: x.md\n'
  ours : [('a[b', 'x.md', 'reference')]
  cmark: <p>See [a<a href="x.md">b</a> here.</p>
```

The target is the same; the stored `link_text` is not. Same class as the inline half of #1526, one family to the right.

## Expected

A reference link's text is the text a CommonMark reader sees, as the inline family's now is.

## Reproduction

```python

print([(l.link_text, l.raw_target) for l in
       extract_links("See [a[b][r] here.\n\n[r]: x.md\n", "s.md")])
# [('a[b', 'x.md')]  -- CommonMark says the text is 'b'
```

## Suspected cause

[verified: read `_find_reference_usage` and `find_bracket_span` on the #1526 branch] The reference scans still use `find_bracket_span`, which takes the first `[` since the last unescaped `]`. #1526 replaced that rule for the inline family with `iter_inline_links`, but deliberately left the reference family alone: its match shape is two adjacent bracket spans rather than one followed by `(`, so the inline iterator does not drop in, and folding it in would have put a second semantic delta in one `INDEX_SEMANTICS_VERSION` bump.

## Open questions / scope

- Scope: `_find_reference_usage` and `_iter_reference_definitions`, and whether the nearest-unmatched rule can be expressed for a two-span shape without duplicating the inline stack.
- Any change alters which rows a note's bytes produce, so per `AGENTS.md` it needs an `INDEX_SEMANTICS_VERSION` bump unless the current value already covers the next release.
- `[unverified]`: whether the shortcut form (`[label]`, still not extracted — see #1519) has to be settled at the same time. The delimiter stack CommonMark uses resolves inline, full, collapsed and shortcut references in one pass, so separating them may cost more than doing them together. I have not tried.
- `[unverified]`: whether any stored `target_path` changes, or only `link_text`. Only `link_text` differed in the case above, but I checked one shape, not a generated corpus.
- Not in scope: the inline family (#1526), wikilinks, and empty destinations (`[a]()`, which stores no row on either side).

pvliesdonk:
## Closes / Refs

Closes #1528
Refs #1531

## What & why

#1526 gave the *inline* family CommonMark's opener rule and left the reference
family on the old one, because its shape is two adjacent bracket spans rather
than one followed by `(` and the inline iterator did not drop in. This carries
the rule across. The same three shapes fall out of it:

| Input | stored before | stored now | a CommonMark reader |
| --- | --- | --- | --- |
| `[a[b][r]` | text `a[b` | text `b` | text `b` |
| `[a [b] c][r]` | **no row** | text `a [b] c` | text `a [b] c` |
| an image reference (`[alt][r]` with a leading bang) | text `alt` | **no row** | an image, no link |

The first is #1528 as filed. The second and third were not filed at all: the
reference scan had **no image test whatsoever**, where the inline scan always
had one. (The bang is spelled out because this body goes through an API that
strips image markup, even inside a code fence.)

What is shared is the **walk**, not the scan. `iter_bracket_links` holds the
delimiter stack, the image rule and the "links may not contain links"
deactivation; each family passes only a test for what must *follow* a closed
span. The rule now lives in one function and cannot drift between the families,
which is #1521's lesson one abstraction lower.

### On removing rows

The third shape is the only one that removes anything, and it removes only an
image reference naming a **note**. One naming an actual image already stored
nothing, dropped by the attachment filter, which is why this went unnoticed.
I checked before deciding: nothing in the tests or design docs pinned the old
behaviour, and a definition shared with a real link keeps its row, so rename
coverage is not lost.

### Evidence

Containment against `markdown-it-py`, asserted over the whole generated corpus
with **no exclusions**: every row stored is a row a reader stores, with the same
text, target and order. That is the exact shape of the defect, since inventing a
row, mistexting one or reordering them all break it.

- Over **97,655** arrangements: **5,682** disagreements before, **5,574** after.
- Comparing the two sets of disagreeing inputs directly, **108 fixed and 0
  introduced** — not just a smaller count, a strict subset.

**The shortcut form (`[label]`) is still not extracted**, and it is now the whole
of the remaining disagreement. Filed as #1531 with the evidence, rather than
guessed at here. Out of scope on evidence, not preference:

- A reader falls back from the full form to the shortcut one when the label
  lookup *fails*, so the scan would have to consult the definition table it
  cannot see. Bolting a shortcut shape test onto the walk leaves **774** inputs
  still disagreeing.
- Supporting it also means excluding definition lines from usage scanning, which
  the two-span shape never had to do — I found that by trying it and watching
  every `[r]: x.md` line become a link.
- The gap is **not purely subtractive**, which is the part worth recording: a
  shortcut link deactivates the openers enclosing it like any other link, so not
  making one leaves an opener active that a reader has retired. In `[[a]b][r]` a
  reader makes two shortcut links where this scanner makes one full reference
  with the text `[a]b`.

## A correction to the #1526 narrative

#1526's design section said the reference family was deferred to keep a second
semantic delta out of one `INDEX_SEMANTICS_VERSION` bump. **That reasoning was
wrong** and this PR corrects it in place: the bump is once per *release*, so a
second delta in the same unreleased range rides the same version and costs
nothing, which is exactly what this PR then does. The real reason for deferring
was the match shape, and wanting this family's before/after measured on its own.

`INDEX_SEMANTICS_VERSION` accordingly stays **10** and its note gains a fourth
entry: no tag contains the commit that set it (checked: 0 of 86), so this
reaches a deployed vault in the same rebuild as #1517, #1519 and #1526.

## Test-quality fixes found while building the evidence

All mine to own rather than incidental cleanups.

Two came out of writing the property:

1. **My first property was scoped by our own output.** It skipped inputs where
   our row count differed from the reader's, which means it could never have
   failed by dropping a row. Replaced with containment plus a pinned per-length
   shortfall count, so losing rows now fails a test.
2. **The markdown-it oracle did not descend into image descriptions**, so it was
   blind to Ex. 575 and would have turned a real disagreement into a pass. This
   is latent in #1526's already-merged module too — I measured it there at
   **zero** inputs of that corpus, so that change's evidence stands, but both
   oracles now descend and the case is pinned by name.

Three more came out of Codecov, and are worth reading as one pattern rather than
three chores. Moving two callers onto a shared walk **narrowed how the code
beneath them is entered**, and three branches silently stopped being exercised:
the anchor-target and empty-`<>`-definition paths in `scanner.py`, and
`find_bracket_span`'s "a `]` before any `[`" path, which both remaining callers
now enter past. None were dead — each is reachable and correct, which I checked
by calling them rather than assuming. The patch-coverage gate reads 100%
throughout and would never have shown this; only the project delta did.
`utils/links.py` now has no missed line and no partial branch.

## Design

`docs/design/design.md` gains a section for #1528 beside #1526's, with the
before / now / reader table, the reason the walk is what is shared, and the
shortcut analysis. `module-map.md` and the `commonmark-gfm.md` claim blocks for
Ex. 512, Ex. 518 and Ex. 575 are updated with pins onto the new tests.

## Local review

- [x] Ran a local review pass on the cumulative diff before opening.
- [x] No commit carries `!`. No operator surface or importable signature
      changes; stored rows change but are rebuilt from vault bytes on the
      existing version, with no operator action.

## Docs impact

- [x] `docs/design/` (new section, module map, reference claim blocks)
- [x] Inline docstrings (`iter_bracket_links`, `_iter_reference_usages`,
      `_follow_reference_label`, `_follow_inline_destination`)
- [x] Enduring decisions ported: the shared-walk rationale and the shortcut
      analysis live in `docs/design/design.md`, not only here.

## Gates

On the current head (`eeb36dc6`):

- `uv run pytest`: **5208 passed, 0 failed.** Two deselected, both environmental
  in this clone and unrelated: the `knope/port/*` release-flow case and the
  `GIT_ASKPASS` one the sandbox exports.
- `mypy src/ tests/`: clean, 252 files.
- `scripts/structural_gate.sh`: 600 lines, 0 violations, 100%.
- `diff-cover`: 55 lines, 0 missing, **100%** patch coverage.
- `scripts/check_references.py`: exit 0; `tests/test_reference_docs.py` passes.


pvliesdonk:
@claude review

pvliesdonk:
Nit confirmed and fixed in 1702b9ea.

It is exactly as described: inserting the new method at the top of `TestAgreementWithACommonMarkReader` pushed the class docstring to the second statement, where Python does not bind it. Verified rather than taken on trust, before and after:

```
__doc__ before: None
__doc__ now:    'Every bracket arrangement is read the way a reader reads it.'
```

Cosmetic, but it is a real loss and the fix is moving one line, so deferring it would have cost more to track than to do.

I also swept the whole tree with an AST check for the same shape — a string-literal statement that is not first in its block. Two other hits, both deliberate and left alone: `tests/test_task_backend.py:39` and `src/markdown_vault_mcp/exceptions.py:89` are PEP 258 attribute docstrings following an assignment, which document the preceding name rather than the enclosing block.

Two commits landed after the review started, so they were not in what you read:

- **3f9dd591** adds two tests for reference branches this change left unexercised. Codecov reported 100% patch coverage but one new missed line and one new partial in `scanner.py` — the anchor-target and empty-`<>`-definition paths. I checked they were reachable by calling them directly rather than assuming, so it was an unexercised branch and not dead code; which tests reached them shifted when the reference scan moved onto the shared walk.
- **1702b9ea** is this fix.

Local gates on the current head: 5204 passed / 0 failed (two environmental deselects), ruff and mypy clean, structural gate 572 lines with 0 violations, `diff-cover` 100% on 55 lines.

---

pvliesdonk:
@claude review

pvliesdonk:
Confirmed, and it was the most useful finding of either round. Fixed in 12f02de5.

## The figures didn't come from the cited test

Reproduced your numbers exactly before doing anything:

```
total=97655 compared=97131 disagree=462
per length: {1: 0, 2: 0, 3: 2, 4: 18, 5: 124, 6: 780, 7: 4650}   <- wide corpus
per length: {1: 0, 2: 0, 3: 0, 4: 1,  5: 9,   6: 62,  7: 390}    <- shipped corpus
```

The cause is exactly as you suspected — a different corpus. My measurement script defined **two single-character labels** (`[a]` and `[r]`); the shipped test defines **one two-character label** (`[ar]`). I then credited the numbers to the shipped file. The figures themselves reproduce, on both halves, with the corrected image-descending oracle:

| defined labels | disagreed before | after | fixed | newly broken |
| --- | --- | --- | --- | --- |
| one, `[ar]` (the shipped corpus) | 466 | 462 | **4** | 0 |
| two, `[a]` and `[r]` | 5682 | 5574 | **108** | 0 |

## The part that matters more than the prose

That gap isn't a labelling detail. With a two-character label, `[a[b][ar]` needs nine characters and the generator stops at seven — so **the shipped corpus exercises this change on four of its 97,655 inputs**. It carries the containment claim honestly, but it is nearly blind to the fix it exists to evidence. I'd have shipped a property that looked thorough and tested almost none of the behaviour in question. Your finding surfaced that; the prose error was the symptom.

So the fix is not only to relabel:

- **`test_the_wider_corpus_disagrees_only_where_recorded`** pins the wide corpus's per-length counts, so the figure quoted in `design.md` and the `INDEX_SEMANTICS_VERSION` note is now what CI computes.
- It is pinned **by count, not containment**, because containment genuinely fails there — single-letter labels fire the shortcut form constantly, and a shortcut link deactivates the openers around it, so a reader makes two shortcut links where this scanner makes one full reference. That is the interaction already documented in the module docstring; the narrow corpus carries containment, the wide one carries the count.
- Both the prose and the test now state that **only the *after* half can live in CI** — the before half needs the pre-change code — rather than implying CI reproduces the comparison.
- `design.md` gains the table above instead of a bare pair of numbers, with the nine-vs-seven-character explanation, so the next reader does not re-derive it.
- The 774-input shortcut estimate came from the same wide corpus and is now qualified as such, in both places it appears.

## Gates on 12f02de5

5215 passed / 0 failed (two environmental deselects), ruff and mypy clean, structural gate 646 lines with 0 violations, `diff-cover` 100% on 55 lines, `check_references.py` exit 0.

---

pvliesdonk:
@claude review

pvliesdonk:
Closing the one open item: the two gates you couldn't run. Both confirmed on `12f02de5`, and neither rests on my word alone.

```
scripts/check_references.py          exit=0
scripts/structural_gate.sh           646 lines, 0 violations, 100%
tests/test_reference_docs.py         2 passed
```

Worth adding that both are covered by CI independently of this comment, so they were never only self-reported:

- The **structural gate** runs as the `Structural Quality (diff)` job, green on this head. CI pins the compare branch to the PR's actual base where the local script derives it, so CI's is the stricter of the two.
- **`check_references.py`**'s guarantees are enforced in the test suite by `tests/test_reference_docs.py`, which the four Test jobs run — that's what catches a `[pins:]` entry naming a test that no longer exists, which is exactly the failure mode this PR could have introduced by renaming tests. It passed there too.

Thanks for reproducing the corrected figures rather than taking the commit message's word for it. The 462/5574 match is the part I most wanted independently checked, given that the previous round's finding was precisely that the documented numbers and the computed ones had drifted apart.

---

PR #1533, opened 2026-09-18
feat(scanner): resolve shortcut reference links

### LINKED ISSUE #1531: [Bug]: the shortcut reference form `[label]` is never extracted
## What happened?

`[label]` with a matching definition is a link for CommonMark (the *shortcut*
reference form, §6.3). The scanner has never extracted one: no outlink on the
note, no backlink on the target. After #1528 it is the **entire** remaining
disagreement between the reference family and a CommonMark reader.

## Observed

[verified: `extract_links` against `markdown-it-py` 3.0.0 `commonmark`, on the
#1528 branch, 2026-09-18]

```
'[ar]\n\n[ar]: x.md\n'
  ours : []
  cmark: <p><a href="x.md">ar</a></p>
```

Measured over 97,655 generated bracket arrangements with one defined label:
after #1528 the scanner disagrees with a reader on 5,574 inputs, and every one
of them is this form.

## Expected

A shortcut reference resolves like the full form, as `[label][label]` already
does.

## Suspected cause

[verified: read `_iter_reference_usages` and `_follow_reference_label` on the
#1528 branch] The shape test requires a second adjacent bracket span. A shortcut
has none, so the span is discarded.

## Why it is not a one-line shape test

Three findings from trying it, each measured rather than reasoned to:

1. **It needs the definition table.** A reader tries the full form first and
   falls back to the shortcut when the *label lookup fails* — `[a][zz]` with
   `zz` undefined is the shortcut link `[a]` followed by literal `[zz]`. The
   walk's `follow` callback is purely syntactic today and cannot see
   `ref_defs`, so the fallback cannot be expressed in it. Returning a shortcut
   whenever no second span follows leaves **774** inputs still disagreeing.

2. **Definition lines become usages.** `[r]: x.md` is itself a bracket span. The
   two-span shape never matched it, so nothing excluded it; a shortcut-aware
   scan matches it immediately and invents a link on every definition line.
   CommonMark strips definitions before inline parsing, and this scanner would
   have to do the same.

3. **The gap is not purely subtractive.** A shortcut link deactivates the
   openers enclosing it like any other link, so *not* making one leaves an
   opener active that a reader has retired. In `[[a]b][r]` a reader makes two
   shortcut links where the scanner makes one full reference with the text
   `[a]b`. So closing this gap will also change rows that have nothing to do
   with the shortcut form itself.

## Open questions / scope

- Scope: `_follow_reference_label` and `_iter_reference_usages` in `scanner.py`,
  plus whatever layering lets the shape test consult `ref_defs`. The collapsed
  form (`[label][]`) already works and should keep working.
- Any change alters which rows a note's bytes produce, so per `AGENTS.md` it
  needs an `INDEX_SEMANTICS_VERSION` bump *unless the current value has not yet
  shipped* — the bump is once per release, not once per change.
- `[unverified]`: whether the footnote departure (#1104) survives contact with
  the fallback ladder. `[^1]` is a bracket span with no second span, so a
  shortcut-aware scan will consider it, and the current skip happens after
  extraction rather than during the walk.
- `[unverified]`: whether adding rows for every `[label]` matching a definition
  is desirable for a real vault, or whether it mostly adds noise. This is a
  behaviour question, not only a correctness one, and worth answering before
  implementing.
- Not in scope: the inline family (#1526), the reference opener rule (#1528),
  wikilinks.

pvliesdonk:
## Closes / Refs

Closes #1531

## What & why

`[label]` standing alone is a link when the label is defined, and no version of
this scanner had ever stored one: a reference usage had to be **two** adjacent
bracket spans. After #1528 closed the opener question this was the *entire*
remaining disagreement with a CommonMark reader.

| corpus | disagreements before | after |
| --- | --- | --- |
| narrow, one label `[ar]` | 462 | **0** |
| wide, `[a]` and `[r]` | 5,574 | **0** |

So the family's property moves from **containment** to **equality**: over 97,655
bracket arrangements (97,131 after excluding those reaching a wikilink) the
scanner now reads references exactly as `markdown-it-py` does, under both
definition sets. The count that #1530 had to pin instead is gone because it is
zero.

## It is a ladder, not another shape

#1531 was filed to warn that the rungs are not independent: full
(`[text][label]`, only when *label* is defined), then collapsed (`[text][]`),
then shortcut (`[text]`). Two consequences, both found by measuring rather than
by reading the spec:

1. **The shape test consults the definition table.** Whether `[a][b]` is a link
   depends on what is defined, so the reference family's `follow` is a closure
   over `ref_defs` where the inline family's is a plain function. That is the
   layering change that kept this out of #1528.
2. **Failing to _resolve_ is not failing to _parse_.** A label that parses but
   is undefined *ends* the ladder — the reader spent the second span on it. A
   second span that is no valid label at all (`[[]`, unbalanced) was never a
   candidate, so the ladder falls through to the shortcut. That is the whole
   difference between `[ar][a]` (no link) and `[a][[]` (a link on `a`), and it
   needs `_parse_link_label`, which requires balance, rather than
   `find_bracket_span`, which does not. **I had this backwards twice**; the
   corpus said so both times, first at 32 disagreements and then at 40.

## Definitions had to stop being usages

A definition line is itself a label standing alone. `[r]: x.md` matched no
previous usage shape, so nothing excluded it; under the shortcut rung it matches
at once, and every definition in a vault would become a link to itself.

Definitions are therefore **blanked out of a region before the usage scan**,
which is what a reader does. Three attempts to test for one *in place* were each
wrong on some generated input:

- a bare `:` is not enough — `See [ar]: here` is a link;
- a line-initial `:` is not either — `[ar]:` with no destination defines nothing;
- and the destination may sit on the following line.

That is the argument for reusing the definition scan's own answer rather than
approximating it, and it is in the code as the reason.

## Three defects surfaced that are not this feature's

- **`iter_bracket_links` retired image openers along with link ones.** Ex. 575
  keeps an image live, so `[an [a][r] alt][r]` is one image and one link, not
  two links. Merged in #1526, invisible until a trailing label could resolve.
- **A definition's destination could be read from beyond a blank line**, and
  **its label could contain one**. CommonMark allows neither. `[:]:` followed by
  a blank line and an unrelated `[ar]: x.md` defined the label `:` with that
  whole line as its target; a stray `[` could swallow a real definition into a
  label, so `ar` was never defined at all.

The two definition defects share a cause worth stating: each wrote a wrong entry
into the table that **nothing ever resolved against**, because the two-span
usage shape could not reference it. The table has been wrong for as long as it
has existed without once producing a wrong row. The shortcut form is the first
thing to read it for a label the author wrote bare.

## And one regression of my own

Making the opener retirement *selective* by walking the stack to clear a flag is
quadratic on a run of unclosed `[`: **8.5 s on 20,000 `[ar][`**. Ranking link
openers separately keeps it selective and O(1). The run now costs about a tenth
of that and grows 2.0x per doubling.

It was caught by a cost test written for this change, which is the argument for
writing one: the shape is `[ar][`, a span whose second span never closes, and
nothing else in the suite generates it. Profiling also corrected my first guess
at the cause — I blamed an `rfind` that turned out not to be the bottleneck.

## Version

`INDEX_SEMANTICS_VERSION` stays **10**, note extended to a fifth entry: no tag
contains the commit that set it (checked, 0 of 86), so all five changes reach a
deployed vault as one rebuild.

This is the largest of them in rows *added* — every bare label matching a
definition gains an outlink and gives its target a backlink. A label with no
definition stores nothing, which is what stops a bracketed aside becoming a
link.

Label **normalisation** is still not CommonMark's. Labels are lower-cased rather
than case-folded (`[ẞ]`/`[ss]` miss), and internal whitespace is not collapsed,
so `[a  b]` and `[a b]` are distinct keys where a reader makes them one. Also
the *last* definition wins where the first should. All three are pre-existing,
none are reachable from the generated alphabets, and they are recorded in
`docs/design/reference/commonmark-gfm.md` rather than fixed here.

## Docs impact

- [x] `docs/design/` — a section for the ladder, the definition-removal argument
      and the two table defects; `module-map.md`; the `commonmark-gfm.md`
      accuracy block rewritten (it said "shortcut not extracted") with pins onto
      the two rung tests
- [x] Inline docstrings (`_make_reference_follow`, `_parse_link_label`,
      `_blank_reference_definitions`, `_iter_definition_matches`)
- [x] Enduring decisions ported to `docs/design/design.md`, not only here

## Gates

- `uv run pytest`: **5208 passed, 0 failed.** Two deselected, both environmental
  in this clone: the `knope/port/*` release-flow case and the `GIT_ASKPASS` one.
- `mypy src/ tests/`: clean, 253 files.
- `scripts/structural_gate.sh`: 606 lines, 0 violations, 100%.
- `diff-cover`: 63 lines, 0 missing, **100%** patch coverage.
- `scripts/check_references.py`: exit 0; `tests/test_reference_docs.py` passes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01W85921xkQnx1wR3yG4wdpF)_


pvliesdonk:
Codecov flagged one new miss and one new partial despite 100% patch coverage, same signal as on #1530. This time the answer was different, and worth writing down because the two look identical from the report.

On #1530 the flagged branches were **reachable and untested** — I checked by calling them, then added tests. Here both were genuinely **dead**, in `_extract_reference_links`:

- `raw_target is None` — the ladder only yields a usage whose label it *resolved*, so the consumer's lookup could not miss.
- the footnote skip — a footnote's definition is prose and `_collect_reference_definitions` keeps it out of the table (#1104), so its label resolves to nothing and the ladder declines it before the consumer sees it.

Deleting them would have left the actual problem, which is that there were **two lookups of the same label** in two places, each free to normalise differently from the other. That is the shape of #1521 one layer down. So in 14dae4a0 the ladder hands back the *target* it resolved rather than the label, and the second lookup stops existing; `_iter_reference_usages` sheds its unused `end` element with it.

I verified the footnote behaviour rather than reasoning about it: `_collect_reference_definitions` returns `{}` for a footnote body on this branch **and on `main`**, so nothing moved — the guarantee just lives where it is documented instead of being re-checked downstream. That mattered, because "the footnote guard is now dead" and "I broke footnote collection" produce the same coverage report.

Re-verified after the change:

- both corpora still agree with a CommonMark reader on every input (0 of 97,131 comparable, narrow and wide);
- the `[ar][` run is still linear, 2.0x per doubling;
- `scanner.py` is back to `main`'s missed-and-partial baseline — 6 and 7, the same counts, no new entries;
- 5208 passed, structural gate 641 lines / 0 violations, `diff-cover` 100% on 68 lines.

---

pvliesdonk:
CI went red on `4c37d609` with two timing failures. One is mine and is fixed in 38cec83d; the other is not, and I have filed it rather than quietly widen someone else's test.

**Neither is a Python 3.11 effect, and neither is a regression.** I measured both shapes on a local 3.11 before changing anything:

| test | CI | local 3.11 | bound |
| --- | --- | --- | --- |
| `test_a_long_run_does_not_stall[failed-second-spans]` (mine) | 6.45 s | **0.82 s** | 5.0 s |
| `test_the_alias_shapes_do_not_stall_either[alias-then-lone-bracket]` (#1524's) | 1.005 s | **0.29 s** | 1.0 s |

So the runner was roughly 8x slower than local for the first and 3.5x for the second. Both bounds were set against a fast machine and were policing the *constant* rather than the defect.

### Mine (fixed)

The constant in my shape is §6.3's own: a failed label parse reads up to 999 characters before giving up, once per span. Linear, but not cheap, so no bound near the linear time survives that much machine variance.

What the bound has to separate is **growth**. Either quadratic this test caught multiplies cost by 4 per doubling: at 20,000 repeats the linear version is 1.7 s locally and the quadratic one 8.5 s, which on an 8x-slower runner is about 13 s against 68 s. The bound is now 30 s, and the input grew rather than shrank — a smaller input is precisely what makes a quadratic hard to see. The numbers and the reasoning are in the test, so whoever next sees it go red knows whether to widen it or hunt a bug.

### Not mine

The wikilink one is #1524's test, unrelated to this diff, and missed by five milliseconds. It is a latent flake in `main`, not something this PR caused — though this PR's longer job plausibly helped tip it. Filed as #1534 with the measurements and a suggested direction (same treatment: widen the bound, keep the input large). I have not touched it here; widening another change's test inside this PR is exactly the scope creep the repo's rules warn against, and the push re-runs it anyway.

If it re-fires on the new head I will say so rather than let it sit.

---

pvliesdonk:
## Self-review found five regressions. All fixed; the body above is now wrong in one place.

CI went green on `38cec83d`, so before marking this ready I ran the repo's `code-review` skill over the cumulative diff, four charters in parallel. It did not come back clean, and the finding that matters is not any single defect.

**Retracting the headline.** The table above says the scanner "reads references exactly as `markdown-it-py` does" over 97,655 arrangements. That is literally true and it missed five defects, because the alphabets are `[]!ar` and `[]:\nar` — no `(`, no space, no `^`, no `[[`. Every defect below lives in one of those characters. An exhaustive sweep of the wrong alphabet is exhaustive about nothing else.

### What was broken

Measured against `main` and markdown-it. All five are new in this branch.

| input | `main` | this branch | a reader |
| --- | --- | --- | --- |
| `[a](x.md)` with `[a]` defined | `a → x.md` | **plus** `a → y.md` | `a → x.md` |
| `[TODO]: revisit [a][r] later` | `a → x.md` | **no row** | `a → x.md` |
| `[ ]: x.md`, then `[]` or `[ ]` | no row | **two rows** | no row |
| `[^a][r]` | no row | `^a → x.md` | CommonMark yes, GFM no |
| `[[a]]` with `[a]` defined | wikilink only | **plus** `a → y.md` | n/a (Obsidian) |

Two of them **remove or fabricate rows a deployed vault holds**, which is what made this worth stopping for rather than filing.

### One cause

Every earlier reference form needed a **second bracket span**. The shortcut rung does not — it is the first form decided by a span's *text alone*. So every guarantee that rested on "some other construct already owns this span, and a `(` or a `[[` is not a bracket span" stopped holding at once. It is not a fourth rung bolted on; it changes what the family is.

- **Inline precedence** is now rung zero (`inline_link_follows`). It asks whether the inline form *closed*, not whether it yielded a target: `[a]()` is a link to the empty string for a reader — a span spoken for with nothing to index — while `[a](unclosed` never closed and is a genuine shortcut. Those pull opposite ways, which is why a test for a literal `(` is wrong in both directions.
- **The greedy definition tail met a consumer.** It has always run to the line's end and over-matches; that was free while nothing used the span bounds, and blanking uses them. `_blank_end` cuts to the destination and its title instead.
- **A whitespace-only label** — `[^\]]+` admitted `[ ]` and keyed it `""`. A third latent table entry of exactly the kind this PR set out to close, and the one that would have made every unchecked task box in the vault a link.
- **Half of #1104's footnote guard.** The ladder subsumes a footnote in the *second* span, never in the text. The body above claimed both halves were subsumed; only one was.
- **Wikilinks**, blanked from the reference scan the way definitions are, from `_find_wikilink`'s own answer rather than by shape. `[[a]b][ar]` is what a shape test gets wrong: it opens `[[`, never closes `]]`, and the full reference stands.

### The corpus, and one that was passing on cancelling errors

`TestAgreementWhereTheOtherFormsMeet` adds `()` and a space and compares **both** families' rows at once. Verified by removing the precedence fix and watching it fail at length five — it is an instrument, not decoration.

Worth stating plainly: `[[a]]` used to produce a wikilink row **and** a spurious reference row, and the spurious row *matched* markdown-it, because a reader really does read `[[a]]` as a bracket around a shortcut. The corpus passed because two errors cancelled. Removing the row exposed the older, deliberate Obsidian departure, so all three corpora now skip inputs the wikilink family claims, and say why.

### Three markdown-it deviations, recorded

Widening the alphabet turned up three places the **oracle** departs from the spec. These matter because a deviation in the oracle reads as a defect here. Each is in `docs/design/reference/commonmark-gfm.md` with its evidence, and excluded from the sweep by name rather than silently:

1. no fallback when only whitespace follows a `(` to the end of the inline block (§6.3 says a failed inline attempt falls through, so `[a](` is a shortcut);
2. no fallback on a whitespace-only label (`[ ]` is no label under §4.7);
3. a label scan that counts nesting, where §4.7 forbids inner brackets outright — we match markdown-it here, and `[unverified]` whether cmark does.

### Also corrected

Nine places where this PR's own prose had drifted from its code, including an `INDEX_SEMANTICS_VERSION` note that said "version 10 still does not cover the shortcut form" three lines above "version 10 covers the shortcut form", and a test module still claiming a "nothing else moved" property this PR had deliberately narrowed. Two further errors were caught in my *replacement* text, both by checking the source rather than trusting what I had just written.

`INDEX_SEMANTICS_VERSION` stays **10**, note unchanged: every defect here was introduced in this unreleased range, so the net delta against the last release is still the shortcut form alone.

### On cost

Not previously disclosed, and it should have been. The definition-blanking pass makes link extraction **1.16x** slower on a realistic 62 KB note (29.9 → 34.6 ms) and 1.28x on a pathological run of unclosed brackets. Still linear. I looked at hoisting the blanking to run once over the body rather than per region and decided against it: `_collect_reference_definitions` runs on `_strip_inline_code(body)` while blanking runs on the per-region strip, and those differ deliberately. Reordering strip-versus-blank is the class of change that was wrong three times already here.

### #1534 is fixed here too

Please treat this PR as closing **#1534** as well (the keyword is not in the body above, so it will not auto-close). The three linearity bounds in `test_links_wikilink_scan.py` asserted on `extract_links` while being about `_extract_wikilinks`:

| case | `_extract_wikilinks` | `extract_links` | ratio |
| --- | --- | --- | --- |
| 40000 bare openers | 0.198 ms | 54.6 ms | 275x |
| unterminated-alias | 0.291 ms | 285.1 ms | 980x |
| alias-then-lone-bracket | 0.358 ms | 296.3 ms | 827x |

A wrong-instrument problem, not a loose-bound one, and it bites at both ends: a bound loose enough to survive the extraction on a slow runner **under coverage** — which is how CI runs these — says nothing about the scan, and one tight enough to say something goes red when unrelated code slows down. Both happened. It also failed locally on this branch under `--cov`, because the definition-blanking pass moved the shape 1.31x without touching the scan. I filed #1534 blaming runner variance; that was the trigger, not the cause.

Now timed on the named function: 0.3 ms against 2 s, and the three cases run in under **5 ms each with coverage on**, while every quadratic they guard still costs seconds.

### Gates on `16294885`

- `uv run pytest`: 5229 passed. Four failures, none of them this change: the two wikilink bounds (fixed by the commit that landed after that run) and the two environmental ones this clone always shows (`GIT_ASKPASS`, and `knope/port/v4.2.0` which does not exist here).
- `ruff check` / `ruff format --check`: clean. `mypy src/ tests/`: clean, 253 files.
- `scripts/structural_gate.sh`: 1151 lines, 0 violations, 100%.
- Coverage: `utils/links.py` 100%, `scanner.py` 99%.
- `scripts/check_references.py` exit 0; `tests/test_reference_docs.py` passes.

I have not marked this ready for review. The diff is now considerably larger than what you last looked at, and the five regressions above are the reason — your call whether you would rather have the review fixes split out.

---

pvliesdonk:
@claude review

pvliesdonk:
Both findings verified and fixed in `0e9467d4`. Thanks — the first one is real and I would not have found it, since my own corpus provably cannot reach it.

### 1. `inline_link_follows` pointy branch — confirmed, with two corrections

Reproduced exactly as described: `[a](<x.md> not a title)` with `[a]: y.md` stored **nothing**, where a reader reads `[a]` as a shortcut to `y.md`. The precedence test claimed the span on paren balance alone; `_parse_pointy_destination` then rejected it because the tail is neither blank nor a title.

Two things measurement changed about the report:

**It is not a regression.** `main` drops the same input:

| input | `main` | before this fix | reader |
| --- | --- | --- | --- |
| `[a](<x.md> not a title)` | nothing | nothing | `a → y.md` |

`main` has no shortcut rung to fall back to, so it never stored that row either. What the rung should have fixed, my precedence test blocked. Mine and wrong, but a gap rather than a break.

**Widening the corpus alphabet would not have caught it.** I tried adding `<` and `>`, measured, and reverted. The shortest input where the two questions disagree *observably* is `[ar](<>a)` — **eight** characters: a defined label, a pointy destination that closes and names nothing, and a tail that is not a title. At seven symbols that sweep is 5.7 million inputs, against 56k for the current one. The missing characters were never the barrier; the shape's length is. So the remedy is a named pin, not a wider alphabet — and I put that reasoning next to the corpus so the next reader does not reach for the alphabet again.

The fix removes the duplicate grammar rather than patching it. `_pointy_destination_span` holds it; `inline_link_follows` asks whether a destination is *written*, `_parse_pointy_destination` whether there is one worth *storing*, and the `<>` rejection is precisely that difference. Same shape as the `_blank_spans` extraction earlier in this branch, and for the same reason: two copies of one rule is how a rule drifts.

Verified after:

| input | now | reader |
| --- | --- | --- |
| `[ar](<x.md> not a title)` | `ar → x.md` (shortcut) | same |
| `[ar](<y.md> 'T')` | inline `<y.md>` | same |
| `[ar](<>)` | nothing | empty href, nothing to index |
| `[ar](<>a)` | `ar → x.md` | same |

**Noted, not fixed:** `[a](x.md not a title)` stores a row whose target is the literal `x.md not a title`. The plain branch of `_parse_destination` strips a trailing title when one is present and keeps whatever remains, so it never rejects a junk tail. `main` does this too and this PR does not touch that path, so it is out of scope here — flagging it rather than silently widening the diff.

### 2. Missing `log.md` entry — confirmed and added

Correct, and the convention is unambiguous: `commonmark-gfm.md` has three prior dated entries in exactly that shape. Added one for the three markdown-it departures and the rewritten accuracy block.

While there I also recorded that **#1517, #1519, #1526 and #1528 each amended that page without an entry**. Not backfilled — they are separate merged changes — but written down so the gap is visible rather than inferred from silence. You were right that `scripts/check_references.py` validates entry format without noticing a missing one.

### Gates on `0e9467d4`

1049 link/scanner tests pass; `mypy` clean on 253 files; `ruff check` and `format --check` clean; structural gate 1385 lines, 0 violations; `check_references.py` exit 0 and `test_reference_docs.py` passes.

---

pvliesdonk:
@claude review

PR #1539, opened 2026-09-19
fix(search): delete notes_fts rows by rowid instead of scanning the content table

pvliesdonk:
## Closes / Refs

Refs #1535

This PR fixes only the O(N²) delete itself (issue's suggestion #1) — the
issue also proposes connection-pragma tuning, a boot-reindex kill switch,
and incremental checkpointing, which are separate follow-up PRs against the
same issue (see "What this PR deliberately does NOT do" below). #1535
stays open until all four land.

## What & why

`notes_fts` is a content-carrying FTS5 table, so an ordinary-column filter
has no index and forces a full scan of its shadow content table.
`_delete_document` used `DELETE FROM notes_fts WHERE path = ?`, called on
every upsert and delete — O(N × table_size) for a batch of N changed notes,
which the reporter measured as hours of wall-clock and terabytes of I/O on
a 25k-note vault. This PR adds a `notes_fts_rowid_map` bridge table
(`document_id → fts_rowid`, populated on insert) and rewrites the delete to
go through it by `rowid` — an indexed lookup FTS5 can execute without
scanning — plus a one-time backfill migration for databases created before
this fix.

- Connection-pragma tuning (`cache_size`/`mmap_size`) from the issue's
  suggestion — separate follow-up PR against #1535.
- A boot-reindex kill switch (env var to disable/defer the unconditional
  reindex) from the issue's suggestion — separate follow-up PR against
  #1535 (needs the `config-contract` skill's process, deliberately not
  bundled with this perf fix).
- Incremental `state.json` checkpointing from the issue's suggestion —
  separate follow-up PR against #1535.
- An explicit `BEGIN`/rollback wrapper around `_migrate_fts_rowid_map` to
  match its sibling `_migrate_notes_fts_summary`'s style — analyzed and
  confirmed safe without it (the backfill INSERT and the meta-sentinel
  write already share one implicit DML transaction, so a crash before
  commit discards both together and the guard correctly retries on next
  open); pure stylistic consistency, not a functional gap, not worth the
  added surface area in this PR.
- Reclaiming `notes_fts` rows that become permanently unbridged after a
  downgrade-then-upgrade across this fix (bounded impact — invisible to
  readers, since search joins through `documents` — but unbounded index
  growth over repeated cycles): #1538.

## Design

<details><summary>Approved design (bridge table vs. rejected alternatives)</summary>

`notes_fts` holds one row per **chunk**, not one row per document, so the
naive `rowid = documents.id` approach (a reasonable first read of FTS5's
`'delete'` command restriction) doesn't work — a document's several rows
can't share one rowid.

Three designs were weighed (recorded as design.md decision 26 and
`docs/design/reference/sqlite-fts5.md`):

1. **Adopted:** a small ordinary `notes_fts_rowid_map` bridge table
   (`document_id`, `fts_rowid`, `WITHOUT ROWID`, composite primary key),
   populated in `_insert_sections` and consulted by `_delete_document`.
   Migration for existing databases is a single pure-SQL join
   (`notes_fts`/`documents` on `path`) — correct regardless of row order,
   no filesystem rescan.
2. **Rejected:** `rowid = documents.id` directly — ruled out by the
   one-row-per-chunk fact above.
3. **Rejected:** converting to an FTS5 external-content table keyed off
   `sections.id` — `notes_fts`'s column set (`path`/`title`/`folder`/`summary`)
   doesn't map onto `sections`'s columns without denormalizing further, and
   migrating the `summary` column would need Python recomputation per
   document rather than a pure-SQL join.
4. **Rejected:** a `fts_rowid` column added directly to `sections` (1:1
   with `notes_fts` rows, no new table) — backfilling it on a legacy
   database would require assuming a positional correspondence between
   `sections` and `notes_fts` rows that nothing in the schema guarantees,
   whereas the bridge table's path-join backfill is order-independent.

The new reference page `docs/design/reference/sqlite-fts5.md` records the
sourced SQLite FTS5 facts this design leans on (content-carrying default,
the `xBestIndex` mechanism, the `'delete'` command's external-content-only
restriction), sourced from sqlite.org and dated 2026-09-19.

</details>

## Local review

- [x] Ran a local code-review pass on the cumulative diff before `gh pr
      create` (see self-review block below) — on top of an unusually deep
      prior review: 4 per-task reviews, one final whole-branch review that
      independently re-derived the EXPLAIN QUERY PLAN output, the FK
      cascade, and the full insert→delete→migrate lifecycle by hand, a
      7-finding fix wave (2 real Important bugs caught: a concurrency issue
      in the migration's `meta` write, and a factually-backwards claim in
      the new reference doc), and a scoped re-review confirming every fix.
- [x] No commit in this PR carries `!` — this changes only internal
      delete-plumbing, not an operator or library-consumer surface, and not
      MCP tool shape either.

## Self-review f3a38c65..670feb59

Charters: rules, diff, comments, history. Conformance: not applicable (no
formal schema/wire-format change in this diff). Past-PRs: skipped (no
review-history lookup beyond what the multi-agent process above already
performed).
Coverage: full — re-walked all four applicable charters directly against
the diff on top of the review history above; found nothing new. Confirmed:
no stray warning markers (`NOTE:`/`WARNING:`/`INVARIANT:`/`DO NOT`/`HACK:`)
in `fts_index.py`; `git log --follow` on `fts_index.py` shows no prior
delete/rowid fix this PR silently reverts or re-breaks; no placeholder or
weak assertions (`assert True`, bare `pass`, `TODO`/`FIXME`) in the new
test code.

No findings.

## Docs impact

- [ ] `README.md` — not applicable, no user-facing surface changed (no new
      config, tool, env var, or CLI flag)
- [ ] `docs/` site pages — not applicable, same reason
- [x] `docs/design/` — Database Schema DDL block, decision 26, and a new
      sourced reference page (`docs/design/reference/sqlite-fts5.md`)
- [x] Inline docstrings — `_delete_document`, `_insert_sections`,
      `_migrate_fts_rowid_map` all updated
- [x] Ported enduring decisions — the bridge-table-vs-alternatives
      rationale lives in design.md decision 26 and the reference page, not
      only in this PR body

---


pvliesdonk:
@claude review

pvliesdonk:
@claude review but not the whole test process, because of got stuck previous round and you can read CI status anyway

PR #1540, opened 2026-09-19
fix(index): checkpoint tracker state during a reindex pass

pvliesdonk:
## Closes / Refs

Refs #1535

This is the second of the issue's four suggestions. PR #1539 fixed the O(N²)
delete (suggestion #1); this one adds incremental `state.json` checkpointing.
The boot-reindex kill switch and the connection-pragma tuning remain separate
follow-up PRs, so #1535 stays open.

## What & why

`state.json` — the change tracker's record of which files are indexed at which
content hash — was written only when a full reindex pass *completed*. A pass
interrupted partway (the motivating case in #1535: a short-lived MCP client
process exiting while the boot reindex runs) therefore restarted from the last
completed pass and redid work it had already committed to the FTS index.

This adds a wall-clock timer to the reindex upsert loop that snapshots tracker
state every `_CHECKPOINT_INTERVAL_S` (30s). The snapshot is taken from the FTS
index's current contents rather than from a running tally, which makes it
accurate by construction: per-document upserts each commit their own
transaction, so a note already re-indexed carries its new hash and one not yet
reached still carries its old one.

**Scope note, stated plainly:** #1535 was written when the delete bug made an
interrupted pass an effectively unbounded restart loop. PR #1539 removed that,
so a restart now costs *linear* re-work rather than a quadratic spiral. This
change therefore saves minutes, not hours, and is scoped to match — a timer, not
a redesign of the reindex pipeline.

- **Protect the scan-and-parse phase.** `_parse_changed_notes()` runs to
  completion for every changed path before the first upsert lands, so an
  interruption during parsing is still redone in full. Making it resumable means
  restructuring parse→upsert into a streaming pipeline, which touches the memory
  profile and the vector lazy-load ordering. Partially self-mitigating: after a
  checkpointed restart the changed set is smaller, so the re-parse is smaller too.
- **Save the vector sidecar per checkpoint.** `_embed_note_inline` holds new
  vectors in memory until `vectors.save()` at pass end, so a checkpoint→crash→
  restart reads those notes as unchanged and the reindex does not re-embed them.
  Safe because `build_embeddings` (without `force`) reconciles missing vectors
  (#665), and `domain.py` queues `build_embeddings_async()` behind the boot
  reindex on every start — so this self-heals on the next boot with no operator
  action. A per-checkpoint `vectors.save()` would re-serialise the whole vector
  index every 30s, which is the wrong trade.
- **Prune the deletion side of that same window.** `changes.deleted` triggers
  `vectors.delete_by_path` in memory *before* the first checkpoint, while
  `vectors.save()` runs at pass end — so every interrupted pass that had
  deletions deterministically leaves orphan vectors in the sidecar. Safe for the
  same reason as above (boot `build_embeddings` removes vectors for documents no
  longer in FTS), and listed here for completeness rather than because it needs
  code.
- **Address the remaining two suggestions in #1535** — the boot-reindex kill
  switch (needs `config-contract` work for a new env var) and the
  `cache_size`/`mmap_size` pragmas. Separate PRs.

**One constraint to carry into the pragma PR:** the durability argument here
holds for process death, which is the failure #1535 names, but not for power
loss — the FTS connection runs WAL with `PRAGMA synchronous = NORMAL` and
`_save_state` does not `fsync` before its atomic rename, so a rename can reach
disk while the last WAL frames do not. That is documented in both doc locations.
If the pragma PR ever sets `synchronous = OFF`, this checkpoint scheme becomes
actively unsafe rather than merely power-loss-exposed.

## Design

<details><summary>Approved design (why a separate method, and why snapshot from the index)</summary>

**The carry hazard, which shaped the API.** `ChangeTracker.detect_changes` parks
unchanged-but-skipped files in `_skipped_carry` / `_skip_reasons_carry`, and
`update_state` merges them into the state it writes and then clears them — by
explicit contract, "consumed by exactly one `update_state` call". A mid-pass
checkpoint routed through `update_state` would eat the carries, and the closing
write would then produce a `state.json` missing every unchanged skipped file,
which the next scan would re-report as newly added.

So the merge logic was extracted into a private `_merge_state`, and
`checkpoint_state` was added as a second writer that does not clear the carries.
**Rejected:** a `final: bool` flag on `update_state` — the carry consumption is a
correctness contract, and a flag invites a caller to get it wrong; two methods
make the two intents distinct.

**Why snapshot from the index rather than track deltas.** `documents.content_hash`
is already ground truth for what landed, because each `upsert_note` commits its
own transaction. A checkpoint is therefore just the closing write run early —
correct by construction rather than by bookkeeping. **Rejected:** an in-memory
tally of applied hashes — `list_notes()` per checkpoint measured 119 ms on a
25,000-document index (2.42 MB `state.json`), which is ~0.4% overhead at a 30s
interval and not worth the extra state to avoid.

**Why 30s.** The cost argument alone does not distinguish 60s from 15–30s, so the
latency side decides it: the first snapshot lands one interval after the upsert
loop starts, so the interval bounds how much committed work an interrupted pass
redoes, and #1535's own evidence is that these client sessions can be short.
Recorded as decision 27 in `docs/design/design.md` with both halves of the
argument, so the value is a decision rather than an unargued default.

</details>

## Local review

- [x] Ran a local code-review pass on the cumulative diff before `gh pr create`
      (self-review block below), on top of: 3 per-task reviews, a final
      whole-branch review that traced the durability argument link by link and
      benchmarked the checkpoint cost, a 5-finding fix wave, and a scoped
      re-review that independently re-derived the new test's fake-clock trace.
- [x] No commit carries `!` — this changes internal reindex lifecycle only; no
      operator surface (no env var, config field, CLI flag, or on-disk format
      change — `state.json`'s `_STATE_VERSION` and schema are untouched) and no
      public library interface (`IndexManager` is not exported from the package
      root).

## Self-review 4ec29a55..90179bb8

Charters: rules, diff, comments, history. Conformance: not applicable — the one
format-like artifact in scope, `state.json`, is unchanged (`_STATE_VERSION`,
schema, and `_save_state` all outside the diff). Past-PRs: skipped.
Coverage: full. Confirmed the closure's captured `newly_skipped` /
`newly_skip_reasons` are assigned once before it is defined and never rebound, so
the by-reference capture is safe; `update_state`'s carry-consumption comment is
preserved verbatim and the design respects rather than contradicts it;
`_state_notes_from_index` has exactly one implementation and two callers.

No findings.

The most useful thing the review chain caught, recorded here because it nearly
shipped: the checkpoint timer's `last_checkpoint` reset could be **deleted with
every test still passing** — the original three tests used intervals `0.0` (where
the reset's effect is invisible) and `3600.0` (where it never executes). Without
it, every note after the first elapsed interval would checkpoint: roughly 50
minutes of added work per pass at 25k notes, the exact cost class this PR exists
to remove. `test_checkpoint_withholds_between_intervals` now pins it with a fake
clock, verified to fail `3 != 2` when the reset is removed.

## Docs impact

- [ ] `README.md` — not applicable; no user-facing surface changed
- [ ] `docs/` site pages — not applicable, same reason
- [x] `docs/design/` — the `tracker.py` change-detection section and decision 27
- [x] Inline docstrings — `checkpoint_state`, `_merge_state`,
      `_state_notes_from_index`, `_upsert_parsed_notes`, and the `_checkpoint`
      closure
- [x] Ported enduring decisions — the carry-consumption contract, the
      snapshot-from-index rationale, the rejected alternatives, and the
      power-loss qualification all live in `docs/design/design.md`, not only here

---


pvliesdonk:
@claude review

PR #1543, opened 2026-09-19
feat(index): add a switch to skip the boot reindex

### LINKED ISSUE #1535: [Bug]: _delete_document full-scans the content-carrying FTS5 table on every note change — O(N^2) reindex; boot reindex is unconditional with no checkpoint or kill switch
## Environment

- OS: Windows 11 Pro (10.0.26200)
- `markdown-vault-mcp` 4.2.0rc2, installed via the project's own `pyproject.toml` / `uv.lock`,
  running as `markdown-vault-mcp.exe serve --transport stdio`
- Vault: 25,380 markdown notes, ~230 MB of source text, no git remote configured
  (`GIT_REPO_URL`/`GIT_TOKEN` unset — pull thread never starts)
- `embeddings=disabled` (`EMBEDDINGS_PATH` unset, `numpy` not installed)
- File watcher not installed (`watchdog` not in the venv) — the server logs
  `file_watcher: watchdog not installed; external file changes will not trigger automatic
  reindex`
- Index: SQLite FTS5, `notes_fts` is a **content-carrying** virtual table (not external-content),
  holding 224,855 rows / ~275–300 MB in `notes_fts_content`
- Registered as an MCP server launched fresh on every client session start (Claude Code); no
  long-lived daemon, no persistent warm cache between runs beyond what `state.json` records

## Defect

`src/markdown_vault_mcp/fts_index.py`, method `_delete_document` (method spans roughly lines
1024–1038; the offending statement is **line 1036**):

```python
def _delete_document(self, cur: sqlite3.Cursor, path: str) -> int:
    """Delete a document row (cascade deletes sections and tags).
    Also removes all FTS rows for the document's path.
    """
    cur.execute("DELETE FROM notes_fts WHERE path = ?", (path,))   # line 1036
    cur.execute("DELETE FROM documents WHERE path = ?", (path,))
    return cur.rowcount
```

`notes_fts` is a content-carrying FTS5 virtual table. FTS5 places no ordinary index on a
non-full-text column, so `DELETE ... WHERE path = ?` has exactly one query plan, confirmed live:

```
EXPLAIN QUERY PLAN SELECT rowid FROM notes_fts WHERE path = 'x'
 -> SCAN notes_fts VIRTUAL TABLE INDEX 0:
```

That is a full scan of `notes_fts_content` — every call. `_delete_document` is invoked by
`upsert_note` for every changed note, so **every note that is added, modified, or reindexed pays
one full-table scan**, making a batch of N changed notes cost O(N × table_size) — quadratic in
corpus size when the whole vault is dirty (e.g. after a mass frontmatter edit, or on any boot
reindex over an unindexed/partially-indexed vault).

Compounding it, `src/markdown_vault_mcp/_fts_connection.py` sets only
`PRAGMA busy_timeout = 5000` and `PRAGMA synchronous = NORMAL` on each connection (lines 177–178)
— **no `PRAGMA cache_size` and no `PRAGMA mmap_size`.** SQLite's default page cache (~2 MB) cannot
hold any meaningful fraction of a 275 MB+ content table, so consecutive scans cannot even benefit
from OS/SQLite caching within the pass; each one re-reads from the start.

Two further defects turn a slow operation into a runaway:

1. **The boot reindex is unconditional and has no kill switch.** `src/markdown_vault_mcp/domain.py`
   `Service.start()`, lines 184–199, submits `vault.index.reindex_async()` on every server start,
   after the warm-start `build_index_async()`. There is no `MARKDOWN_VAULT_MCP_*` environment
   variable anywhere in `config.py` / `config_sections/` that can disable this. (Config knobs that
   look adjacent — `MARKDOWN_VAULT_MCP_FILE_WATCHER` — do nothing here: the watcher path is
   already inert on a fresh venv with no `watchdog` installed, and is unrelated to the boot
   reindex job.)
2. **Progress is committed only on full completion.** `state.json` is written only when a pass
   completes (`managers/index.py:675`, referenced by line number in the source at investigation
   time). If a client disconnects (e.g. the host process exits, as every short-lived MCP client
   session does) before the pass finishes, nothing is checkpointed, so the **next** server start
   repeats the identical full pass from zero. There is no per-note or time-based checkpoint.

## The O(N²) reasoning, with real numbers

Observed server log on a real boot: `reindex: 10480 added, 3271 modified, 0 deleted, 11629
unchanged, 0 skipped` — 13,751 changed notes out of 25,380.

- One `notes_fts` scan (table in OS cache, idle machine): **1.1–1.3 s**
- 13,751 changed notes × ~275 MB content table ≈ **~3.6–4 TB read, ~3.8 hours**, and this is what
  the *boot reindex alone* costs on every single server start, unconditionally, whenever any
  fraction of the vault has changed since the last **completed** pass.
- A live incident (2026-09-14 21:59–23:31 EDT) measured **1,526 GB read at up to 554 MB/s over
  85 minutes** — about 5,090 individual `notes_fts` scans, ≈37% of one full pass — while the
  process held only 93 MB RSS (a loop, not real work) and pegged 16 CPU cores.
- Reproduced by hand afterward, foreground, real MCP `initialize`, sampled every 15 s:
  **40.60 GB read in 180 s** — 225 MB/s average, 423 MB/s peak, 86% of one core sustained. The
  first ~45 s is a SHA-256 hash scan of all files (CPU-heavy, read-cheap); from t=60s the FTS
  scans begin and the read rate jumps roughly two orders of magnitude.
- The cost is invisible on a small vault: on ~900 notes the same defect costs a few seconds and
  looks like normal indexing latency. It becomes fatal specifically as corpus size grows — this
  is the textbook signature of an O(N²) algorithm being mistaken for O(N) at small scale.

Four other causes were checked and ruled out by reading the code and the live config, not by
guessing: not a watcher storm (`watchdog` not installed; the debounce would collapse many events
into one restartable timer regardless), not an embeddings build (`embeddings=disabled`,
`numpy` not installed), not git polling (`GIT_REPO_URL`/`GIT_TOKEN` unset, pull interval resolves
to 0), and not a corrupt-index cold rebuild (`meta.index_semantics_version` matched the source
constant, `build_completed_at` set, warm-start gate passed in O(1)).

## Minimal fix suggestion

Either of these removes the O(N²) behavior without changing external behavior:

1. **Delete by rowid through a path → rowid map.** Maintain the `documents` table's own `id` (or
   a `rowid`) as the FTS5 `rowid`, and delete via
   `DELETE FROM notes_fts WHERE rowid = (SELECT id FROM documents WHERE path = ?)` — an indexed
   lookup on `documents.path` (add an index there if one doesn't exist) followed by an O(1) FTS5
   rowid delete, instead of a full content-table scan.
2. **Or: make `notes_fts` an external-content table**, keyed by `content=` pointing at
   `documents` with `documents.id` as the shared `rowid`. External-content FTS5 tables delete via
   rowid triggers/statements in O(log n), not a linear scan of the shadow content table, because
   the searchable text is projected from the external table rather than carried inside the FTS5
   shadow tables.

Either change should also add `PRAGMA cache_size` and `PRAGMA mmap_size` tuned to the FTS5 shadow
tables' expected size, so that even a legitimate large scan (e.g. a genuine cold rebuild) reads
from page cache on the second and subsequent passes rather than re-reading from disk every time.

Separately (not the root cause, but what turns a slow pass into a repeating one): commit
`state.json` incrementally — e.g. after each N notes or every T seconds — rather than only on
full pass completion, so a boot reindex interrupted by a short-lived client (the common case for
an MCP server spawned per session) can resume instead of restarting from zero. And expose an
environment variable to disable or defer the unconditional boot reindex in `domain.py`
`Service.start()` (line 198, `vault.index.reindex_async()`), so operators of large vaults can
choose to reindex out-of-band instead of paying the cost on every session start.

## Repro steps (abbreviated)

1. Point `MARKDOWN_VAULT_MCP_SOURCE_DIR` at a vault of ~20,000+ markdown notes with no
   `MARKDOWN_VAULT_MCP_INDEX_PATH`/`state.json` warm cache (i.e. a fresh index or one where the
   prior pass never completed).
2. Touch (modify mtime or content of) a large fraction of the notes at once — e.g. a scripted
   frontmatter backfill — so the next boot's reindex has thousands of changed paths.
3. Start the server (`markdown-vault-mcp.exe serve --transport stdio`) and either let a client
   connect and disconnect quickly, or just watch process I/O.
4. Observe sustained high-rate disk reads (hundreds of MB/s) for minutes to hours with very low
   RSS (under 100 MB) — the signature of a scan loop, not indexing work — and confirm with
   `EXPLAIN QUERY PLAN SELECT rowid FROM notes_fts WHERE path = ?` that the plan is a virtual
   table scan.
5. Kill the process before the pass completes (or just let a short MCP session end) and start it
   again: observe the identical pass restart from zero, because `state.json` was never written.

Happy to test a patch against the same vault. Thanks for the project.

pvliesdonk:
## Closes / Refs

Refs #1535

Third of the issue's four suggestions. PR #1539 fixed the O(N²) delete,
PR #1540 added incremental checkpointing; this adds the boot-reindex
opt-out. The connection-pragma tuning is the last one and will close the
issue.

## What & why

`Service.start()` submits a reindex job on every server start, reconciling
changes made while no server was running (#665). This server is spawned
fresh per client session, so on a 25,000-note vault that is a full
filesystem scan per session. `MARKDOWN_VAULT_MCP_BOOT_REINDEX=false` skips
that submission.

The gate covers **only** `reindex_async()`. `build_index_async()` and
`build_embeddings_async()` stay unconditional, and not for tidiness: the
initial build is a correctness requirement (a fresh install must still get
an index — now pinned by a cold-vault test), and the embeddings submission
is what lets PR #1540's inline-embed vector windows self-heal at boot.

Default is `true`, so nothing changes for anyone who does not set it.

- **Defer, as opposed to disable.** The issue asked for "disable or defer".
  Deferring made sense when an interrupted reindex could restart forever;
  #1539 and #1540 bounded that cost and made a pass resumable, so a delay
  now only relocates bounded work while adding a numeric knob and its
  validation rule. Recorded as a rejected alternative in decision 28.
- **Reindex after the startup git pull.** See the section below — the
  docs now describe this honestly; the behaviour change is proposed in
  #1542 rather than made here, because it would change what the switch
  means.
- **Surface the resulting staleness anywhere queryable.** With the switch
  off, `get_index_status` reports nothing about whether reconciliation
  ran, and `index_stale` is a writer-idle signal that reads `false`
  regardless. Tracked in #1542, which also notes that the tool-description
  surface budget (#1458, 57,999/58,000) may constrain the fix.
- **Close #1535.** The `cache_size`/`mmap_size` pragma PR does that.

**Constraint inherited from PR #1540, repeated because the pragma PR is
next:** that PR must not set `synchronous = OFF`, or #1540's checkpoint
durability argument breaks.

## Design

<details><summary>The interaction the final review uncovered, and why the docs changed more than the code</summary>

The switch looked self-contained until the whole-branch review traced four
links that nobody — including this PR's own plan — had put together:

1. `sync_from_remote_before_index()` (`domain.py`) pulls at boot and
   deliberately does **not** reindex. Its docstring reasoned that
   `build_index()` would scan the updated working tree.
2. But `build_index_async()` short-circuits in O(1) on a warm FTS sentinel
   and scans **nothing**. So it was in fact the *boot reindex* — not the
   build — that indexed whatever the startup pull fetched. That docstring
   was already leaning on the boot reindex implicitly.
3. `should_start_file_watcher()` returns `False` whenever git pull or a
   deliverable webhook is active — so in exactly those deployments the
   watcher is stood down **by the code, not by the operator**.
4. Net: with this switch off on a git-synced deployment, content the
   startup pull just fetched stays unindexed until some later pull moves
   HEAD. Bounded and self-healing, but silent.

The design doc's first draft warned about "an operator who disables both
this and the file watcher" — the wrong combination, since the operator
never disables the watcher; enabling git sync disables it for them. That
is corrected here, along with the `sync_from_remote_before_index`
docstring and a published guide page (below).

The reviewer proposed a tighter behaviour: `sync_once()` already returns
whether HEAD advanced, so the boot reindex could be submitted when the
startup pull actually moved HEAD, regardless of the switch — a pull that
moved HEAD being positive evidence of change rather than the speculative
scan #1535 objects to. That is deliberately **not** in this PR: it changes
what the switch means and deserves its own review, and the operator who
reported #1535 runs exactly this configuration and should weigh in. It is
the substantive half of #1542.

</details>

## Local review

- [x] Ran a local code-review pass on the cumulative diff before
      `gh pr create`, on top of 3 per-task reviews, a final whole-branch
      review, a 6-finding fix wave, and a scoped re-review.
- [x] No commit carries `!`. The default preserves existing behaviour on
      every entry point — `Service.start()` is the only `reindex_async()`
      submission in `src/`, and `markdown-vault-mcp reindex` builds its own
      Vault and is unaffected by the flag.

## Self-review da3c58e9..54f7e042

Charters: rules, diff, comments, history. Conformance: not applicable.
Coverage: full, on top of the review chain above.

Two findings from that chain are worth surfacing here rather than burying,
because both were documentation defects of the class this project calls out
explicitly — prose describing changed behaviour, not found by grepping the
changed identifier:

1. `docs/guides/git-integration.md` promised a webhook delivery arriving
   during the initial build is "handled, not dropped", because "the boot
   reconciliation pass that follows the build picks the pulled changes up".
   With the switch off there is no such pass, so the promise inverted on a
   **published, Vale-linted** page. Now qualified.
2. The `sync_from_remote_before_index` docstring's stated reasoning was
   already only true on a cold build, before this PR existed. Corrected
   rather than left to rot.

No code findings.

## Docs impact

- [ ] `README.md` — not applicable; the domain table is a curated
      landing-page subset and this is a large-vault operator switch
- [x] `docs/` site pages — `docs/configuration.md` regenerates from the
      field's metadata (the config surface went 119 → 120 variables, also
      propagating to the wizard spec, four env examples, and `server.json`
      — all additive, no manifest version touched), and
      `docs/guides/git-integration.md` is corrected
- [x] `docs/design/` — the `#665` boot-reconciliation narrative amended,
      plus decision 28 with both rejected alternatives
- [x] Inline docstrings — `sync_from_remote_before_index`, the gated block
      in `Service.start()`, and the `test_boot_reconciliation` module
      docstring
- [x] Ported enduring decisions — decision 28 carries the scope rationale
      and the rejected defer knob; the git-sync interaction is in the
      design doc and #1542, not only in this PR body

---


pvliesdonk:
@claude review

